PR #19, opened 2021-11-02
Update Fidesops overview /w video & consistency

NevilleS:
# Purpose
Adds the embedded intro video for Fidesops, as well as some formatting consistency changes to the "What is Fidesops" page (plus a few minor edits).

# Changes

- Add embedded Fidesops video
- Edit index.md language

# Ticket
n/a

PR #20, opened 2021-11-02
Documentation updates and video additions

cilliankieran:
# Purpose
Update to include video asset as well as update to correct Fidesops logo.


NevilleS:
Merged into #19

PR #22, opened 2021-11-02
Add Forking Clarification 

pattisdr:
# Purpose

Dev docs specify to clone the original, but most contributors will be cloning a fork.

# Changes

- Clarify at the top of the docs that you should be cloning your fork
- Add some more specific instructions about how to keep your fork in sync with the upstream branch
- Also, make the PR from my fork!

# Ticket
[unticketed]


NevilleS:
Good point!

PR #25, opened 2021-11-02
MySQL for M1 Chips

pattisdr:
# Purpose
Users with M1 chips when running integration tests locally get the error:  `docker: no matching manifest for linux/arm64/v8 in the manifest list entries` because the MySQL image doesn't support arm64 architecture.

# Changes
- For MySQL, add --platform linux/amd64 to run an Intel image under emulation

From [Docker Desktop for Apple silicon](https://docs.docker.com/desktop/mac/apple-silicon/)

> Not all images are available for ARM64 architecture. You can add --platform linux/amd64 to run an Intel image under emulation. In particular, the mysql image is not available for ARM64. You can work around this issue by using a mariadb image.
- Get rid of the extra docker-compose file created specifically for quickstarts since the primary `integration-test` docker-compose file can be used - this also opens up the quickstart to be able to be used with mysql too.

# To test
Run `make pytest-integration-access` locally - should work for both M1 and intel chips.

# Ticket
[No ticket]


PSalant726 (approved):
Confirming that these changes allow `make pytest-integration-access` to run locally, and that all tests pass ✅ 

I tested on a 2020 M1 MacBook Pro.

pattisdr:
appreciate the review @PSalant726! ⭐

pattisdr:
Reopened in #54

PR #27, opened 2021-11-03
Target Specific Tests in Docker 

pattisdr:
# Purpose
Documentation suggests you can do something like `make path=path/to/dir/ pytest`, but that was removed because "path" is overloaded.   This is a really useful feature to target specific tests so I'd like to add this back, except don't use "path".

# Changes
- Target specific tests with `make pytestpath=path/to/dir/ pytest`
- Update contributing documentation
- Remove loading entire `tests` directory in integration tests which was making it impossible to target specific tests.  It was previously added to reuse the test configuration and test fixtures, but this no longer seems necessary.  My theory is that some integration tests have been added that are outside the integration_test directory.

# Ticket
[unticketed]


eastandwestwind (approved):
Tested these commands out, looking good. Thanks!

PR #28, opened 2021-11-03
Update Postman Collection / Tutorial re: Client Scopes 

pattisdr:
# Purpose

When an Oauth Client is created in Fidesops, its scopes can be set in the same request; we no longer need a followup request to set those scopes separately.  This has been updated in oauth docs, but update other documentation to reflect.

# Changes
-  Update postman collection to create client/set scopes in a single request
- Update tutorial to create client/set scopes in a single method. I didn't realize we had this capability so I split into two requests originally.

# Ticket
[not ticketed]


eastandwestwind (approved):
Looks great, thanks for noticing this inconsistency!

PR #30, opened 2021-11-03
Add Connection and Storage Secrets Schemas to API Documentation [SOL-269]

pattisdr:
# Purpose
The schemas for the secrets to connect to a database or create a storage location are not yet included in our docs.  These schemas have been more difficult to specify, as there are multiple possible schemas for both of these endpoints, and we can't specify which schema is the relevant one upfront. There are also overlaps in the possible fields between schemas, making it difficult for fastapi to automatically select the correct one - we need to wait until we load the connection config or storage config in the view to be sure we have the correct schema. However, we still want the possible schemas documented.

# Changes

- Create a new `DocsOnlySchema` to use when we still want a schema to show up in the docs, but we don't want it validated yet - the validation will happen later once we've determined the correct schema to use, of the multiple possibilities. This schema validation just passes the incoming request body through as-is, h/t @seanpreston for the idea.
- Get rid of needing to access the raw fast api request in connection secrets since we're passing through the unvalidated secrets
-  Catch a possible 500 error when validating mongo secrets, discovered while testing

## Docs!

Screen Shot 2021-11-03 at 6 02 55 PM
Screen Shot 2021-11-03 at 6 02 31 PM

# Ticket
https://ethyca.atlassian.net/browse/SOL-269 (old, internal ticket)

### DIFF
--- a/src/fidesops/schemas/base_class.py
+++ b/src/fidesops/schemas/base_class.py
@@ -2,8 +2,6 @@ from typing import List, Any
 
 from pydantic import BaseModel
 
-from fidesops.common_exceptions import ValidationError
-
 
 class FidesopsSchema(BaseModel):
     """
--- a/src/fidesops/schemas/base_class.py
+++ b/src/fidesops/schemas/base_class.py
@@ -22,11 +22,11 @@ class FidesopsSchema(BaseModel):
 BaseSchema = FidesopsSchema
 
 
-class DocsOnlySchema(BaseModel):
+class NoValidationSchema(BaseModel):
     """A schema to be used for API documentation only, when validation is handled later in the request process,
     but we still want valid request schemas to show up in the docs."""
 
     @classmethod
-    def validate(cls: "DocsOnlySchema", value: Any) -> Any:
+    def validate(cls: "NoValidationSchema", value: Any) -> Any:
         """Returns value exactly as it was passed in, when validation is going to be handled later."""
         return value
--- a/src/fidesops/schemas/base_class.py
+++ b/src/fidesops/schemas/base_class.py
@@ -1,6 +1,8 @@
-from typing import List
+from typing import List, Any
 
 from pydantic import BaseModel
+
+from fidesops.common_exceptions import ValidationError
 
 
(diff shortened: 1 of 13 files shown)


src/fidesops/service/connectors/mongodb_connector.py
```diff
@@ -45,7 +45,10 @@ def client(self) -> MongoClient:
         """Returns a client for a MongoDB instance"""
         config = MongoDBSchema(**self.configuration.secrets or {})
         uri = config.url if config.url else self.build_uri()
-        return MongoClient(uri, serverSelectionTimeoutMS=5000)
+        try:
+            return MongoClient(uri, serverSelectionTimeoutMS=5000)
+        except ValueError:
+            raise ConnectionException("Value Error connecting to MongoDB.")
```
seanpreston, line 51: Are we suppressing the error message here because it could contain the secrets?

src/fidesops/schemas/base_class.py
```diff
@@ -20,3 +20,13 @@ class Config:
 
 
 BaseSchema = FidesopsSchema
+
+
+class DocsOnlySchema(BaseModel):
```
seanpreston, line 25: The function of the class isn't anything to do with the docs, would `NoValidationSchema` be a better name here? It's great that we say explicitly in the docstring that validation is switched off too.

src/fidesops/service/connectors/mongodb_connector.py
pattisdr: yes, they often do contain secrets, but this is probably going too far in the other direction and suppresses all useful information.  A good future improvement would provide more information without displaying secrets.

src/fidesops/schemas/base_class.py
pattisdr: Yes that makes sense to me

src/fidesops/service/connectors/mongodb_connector.py
pattisdr: ah - I should also clarify that this was put here to begin with because a ValueError causes a 500 error to be thrown

PR #31, opened 2021-11-05
Update Query Builder to use Masking Strategies on Policy Rules [#47]

### LINKED ISSUE #47: Update Query Builder to use Masking Strategies on Policy Rules
Currently we only support the null masking strategy - look into updating the query builder to use the masking strategies attached to the Policy Rules for various data categories.

pattisdr:
# Purpose

Currently, when the query builder masks data, it replaces it with null values.  Instead, update the query builder to use the masking strategies attached to Policy Rules for various data categories.  As a first iteration, we'll still only allow null masking policy rules to be created while we get some other pieces in place, but this is a first step to connecting to actual masking strategies.

Future TODO's will include getting masking strategies to work with various data types #37 and have the caller be responsible for generating the key/salt, etc.

# Changes

- Restrict Policy Rules so we can only create ones with a null masking strategy specified (because this is all that technically works right now)
- Update the `generate_update_stmt` methods for the `SQLQueryConfig`  and `MongoQueryConfig` to replace values with those generated by the masking strategy on the rule policies instead of None
- Dry up a mapping that points data categories to fields on a collection for use in multiple places - node.collection.categories_to_fields
- Add base `QueryConfig.build_rule_target_fields` that maps all the rules on the policy to corresponding fields on a collection (via data categories)
- Add base `QueryConfig.update_value_map` that pulls masking strategies off rules and takes in an input value, and returns field names mapped to their masked value.

# Ticket
https://ethyca.atlassian.net/browse/SOL-220 (old internal ticket)
https://github.com/ethyca/fidesops/issues/47 
Resolves #47

### DIFF
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -77,6 +77,8 @@ Field identities:
 """
 from __future__ import annotations
 
+from collections import defaultdict
+
 from typing import List, Optional, Tuple, Set, Dict, Literal
 from pydantic import BaseModel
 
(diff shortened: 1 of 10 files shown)


src/fidesops/graph/config.py
```diff
@@ -223,6 +225,26 @@ def identities(self) -> Dict[str, Tuple[str, ...]]:
         flds_w_ident = filter(lambda f: f.identity, self.fields)
         return {f.name: f.identity for f in flds_w_ident}
 
+    @property
+    def categories_to_fields(self) -> Dict[str, List]:
```
stevenbenjamin, line 229: Minor nit - I think it's be a bit clearer for the function name to be consistent with the fact that it's a property rather than a verb, e.g something like "fields_by_category"

pattisdr: i agree, that's a better name, thanks!

Issue #40, opened 2021-11-06
Query execution must respect the configured `ConnectionConfig.access` property


seanpreston:
Currently the query builder will still try to generate erasure statements for database tables that the configured connection has no write access to. If no write access is granted, Fidesops should not generate these statements or try to execute them.

PR #49, opened 2021-11-08
Require ConnectionConfig Write Access to Mask Data [#40]

pattisdr:
# Purpose

Currently the query builder will still try to generate erasure statements for database tables that the configured connection has no write access to. If no write access is granted, Fidesops should not generate these statements or try to execute them.

# Changes

- Add a check in `GraphTask. erasure_request` that verifies that the related `ConnectionConfig` for the particular TraversalNode has `write_access` or it skips building and executing an update statement. 
- Fixes some of our integration erasure unit tests that were using connection configs with read access

# Ticket
Resolves https://github.com/ethyca/fidesops/issues/40

### DIFF
--- a/src/fidesops/task/graph_task.py
+++ b/src/fidesops/task/graph_task.py
@@ -18,7 +18,7 @@ from fidesops.graph.config import (
 )
 from fidesops.graph.graph import Edge, DatasetGraph
 from fidesops.graph.traversal import TraversalNode, Row, Traversal
-from fidesops.models.connectionconfig import ConnectionConfig
+from fidesops.models.connectionconfig import ConnectionConfig, AccessLevel
 from fidesops.models.policy import ActionType, Policy
 from fidesops.models.privacy_request import PrivacyRequest, ExecutionLogStatus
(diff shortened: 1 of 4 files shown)


src/fidesops/task/graph_task.py
```diff
@@ -123,6 +123,11 @@ def generate_dry_run_query(self) -> str:
         """Type-specific query generated for this traversal_node."""
         return self.connector.dry_run_query(self.traversal_node)
 
+    def can_write_data(self) -> bool:
+        """Checks if the relevant ConnectionConfig has been granted "write" access to its data"""
+        connection_config = self.connector.configuration
```
eastandwestwind, line 128: does `self.connector.configuration` have a schema or is it arbitrary keys?

pattisdr: good catch, thanks

eastandwestwind (approved):
lgtm!

pattisdr:
whoa thanks for that quick turnaround @eastandwestwind ⚡ ⚡

PR #53, opened 2021-11-09
PUT -> PATCH Endpoints [#15]

### LINKED ISSUE #15: Rename PUT `/policy` component endpoints to be PATCH instead
The Policy API operations are kind of misleading and a generally not great practice for APIs:

Current state: 
* `PUT /policy`
* `PUT /policy/{policy_key}/rule`
* `PUT /policy/{policy_key}/rule/{rule_key}/target`
and the Rules will overwrite themselves if they are already present on the Policy, however, any elements that do not exist in the list sent to the API will not be deleted.

Expected state:
I'd expect a PATCH op for the policy endpoints where you can make a point-change to the rules and target, without updating the whole policy

pattisdr:
⚠️  This is a breaking change to our v1 API - we should weigh whether this change should be made in v1 at all and make sure this is the route we want to go.

# Purpose

Our Bulk PUT endpoints really behave like PATCH endpoints, in that we can target specific changes to certain policies, rules, targets, connections, storage configs, and datasets, without supplying every resource.  Resources that are not included in the request are not deleted.  

Update these PUT operations to be PATCH instead.

# Changes
- Bulk requests to create/update Policies, Rules, RuleTargets, StorageConfigs, ConnectionConfigs and Datasets are now PATCH operations instead of PUT operations.
- Update the tests to reflect this change.
- Update the Guides and the Quickstart to reflect this change.

## Notes: 
- Only Policies/Rules/RuleTargets were specified in #15, but it seems this applies to the other bulk create/update endpoints.
- Fidesdemo https://github.com/ethyca/fidesdemo will need to be updated to make PATCH requests too
- We do have a pretty nonstandard way to create/update these resources. While BULK operations are very useful for some people, I think we should add support for singular create and update as well, as these endpoints are more intuitive to use.

# Ticket
Resolves #15


seanpreston (approved):
Thanks for this @pattisdr

PR #55, opened 2021-11-09
Refactor: Removes circular dependencies- closes #34

### LINKED ISSUE #34: Address circular dependencies
Code should be restructures so that we don't have to import statements nested within functions. A list of known instances were found in this PR: 
https://github.com/ethyca/solon/pull/91/files#diff-56330cb465bb07615b370f03a68a84fee0fb3424ce529869938ff57ed9682ff9R150

eastandwestwind:
# Purpose
Remove a couple instances of local import shortcut to get around circular dependencies

# Changes
- Extracts out privacy request runner into separate service
- Allows caller of `PrivacyRequestRunner.run()` to handle success / error instead of using callbacks

# Issue
https://github.com/ethyca/fidesops/issues/34


src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -0,0 +1,147 @@
...
+                privacy_request=self.privacy_request,
+                policy=policy,
+                graph=dataset_graph,
+                connection_configs=connection_configs,
+                identity=identity_data,
+            )
+            if not access_result:
+                logging.info(
+                    f"No results returned for access request {self.privacy_request.id}"
+                )
```
stevenbenjamin, line 75: Should probably return here if there are no results?

eastandwestwind: we still want to save the processed at time when there are no results for the specific request: https://github.com/ethyca/fidesops/pull/55/files#diff-f0ba0355b20a63eaf049b428f1a3dbddac8567fac5df5f27d0c6a1294066d485R123

pattisdr: I think the current functionality is useful here in that it lets the user confirm there are no results -

pattisdr (approved):
I think this looks good @eastandwestwind, I appreciate how you tried to not change functionality in this refactor, doing one thing at a time.  I agree that it's fine to let the caller of PrivacyRequestRunner.run() handle any followup operations - actually it's nice not having onetrust context be needed for the PrivacyRequestRunner.

Issue #37, opened 2021-11-06
Add type awareness/coersion to the query builder


seanpreston:
In order to write more complex erasure queries, the query builder must be aware of the type of the column it's writing to. The prioritized data types are:
* varchar/string
* integer
* bool
* float
* object id
* timestamp

The total, aspirational list of types we want to be able to support are: 
```
   VARCHAR(new String[] {"string","char","varchar","enum","nchar","nvarchar"}),
    STRUCTURED(new String[] {"structured_property"}),
    TEXT(new String[] {"text","ntext","object"}),
    GEOPOINT(new String[] {"geo_point"}),
    JSON(new String[] {"json", "map"}),
    VARIANT(new String[] {"variant"}),
    KEY(new String[] {"key"}),
    BOOLEAN(new String[] {"boolean","bool","bit"}),
    INT(new String[] {"integer","serial","int4","int"}),
    LONG(new String[] {"long","bigint","bigserial","int8", "int64"}),
    REAL(new String[] {"float","float8","float4", "real"}),
    DOUBLE(new String[] {"double", "double precision", "float64"}),
    TIMESTAMP(new String[] {"datetime","timestamp", "datetime2", "smalldatetime"}),
    BIGINT(new String[] {"bigint"}),
    DATE(new String[] {"date"}),
    TIME(new String[] {"time"}),
    DECIMAL(new String[] {"decimal","numeric", "number", "bignumeric", "bigdecimal"}),
    SMALLINT(new String[] {"smallint"}),
    TINYINT(new String[] {"tinyint", "byteint"}),
    OBJECTID(new String[] {"objectid"}),
    ARRAY(new String[] {"array"}),
    GEOMETRY(new String[] {"geometry"}),
    UUID(new String[] {"uuid"}),
    OTHER(new String[] {"other"});
```

pattisdr:
This seems to be a prerequisite for https://github.com/ethyca/fidesops/pull/31, where I am working on using the masking strategies on the Policy Rules instead of the default null masking strategy. Before we open this up to all the masking strategies, we need the type awareness described here.

PR #60, opened 2021-11-11
add functionality to types

stevenbenjamin:
# Purpose
Add type functionality to traversals
# Changes

- 

# Ticket


iamkelllly:
@stevenbenjamin will this address #37 ?

src/fidesops/graph/config.py
```diff
@@ -204,6 +203,21 @@ class Config:
...
+
+        - If the data_type is None, then it has not been specified, so just return the input value.
+        - Return either a cast value or None"""
+
+        if self.data_type:
+            # if no data type is specified just return the input value.
+            # Skip conversions for query tokens, which are only for display output
+            if not isinstance(value, QueryToken):
+                converter: DataTypeConverter = self.data_type.value
+                return converter.to_value(value)
```
eastandwestwind, line 217: Could we have a dev-friendly log if the data type isn't supported?

stevenbenjamin: Yes, but I think that should get caught in the validations on the dataset configuration in the endpoints.

src/fidesops/service/connectors/query_config.py
```diff
@@ -99,11 +93,11 @@ def query_sources(self) -> Dict[str, List[CollectionAddress]]:
     def display_query_data(self) -> Dict[str, Any]:
         """Data to represent a display (dry-run) query. Since we don't know
         what data is available, just generate a query where the input identity
-        values are assumed to be present and singulur and all other values that
+        values are assumed to be present and singular and all other values that
```
eastandwestwind, line 96: 💯

src/fidesops/util/collection_util.py
```diff
@@ -34,3 +34,10 @@ def partition(_iterable: Iterable[T], extractor: Callable[[T], U]) -> Dict[U, Li
     for t in _iterable:
         append(out, extractor(t), t)
     return out
+
+
+def filter_nonempty_values(d: Dict[Any, Any]) -> Dict[Any, Any]:
+    """Return the input map with empty values removed"""
+    if d:
```
eastandwestwind, line 41: if we're checking for `d`, is the input param `d: Optional[Dict[Any, Any])`?

src/fidesops/util/collection_util.py
```diff
@@ -34,3 +34,10 @@ def partition(_iterable: Iterable[T], extractor: Callable[[T], U]) -> Dict[U, Li
...
+        return {e[0]: e[1] for e in d.items() if e[1]}
```
eastandwestwind, line 42: possible to use `item` instead of `e` just for clarity?

tests/task/traversal_data.py
```diff
@@ -69,7 +69,7 @@ def combined_mongo_posgresql_graph(
         connection_key=mongo_config.key,
     )
 
-    return (mongo_dataset, postgres_dataset)
+    return mongo_dataset, postgres_dataset
```
eastandwestwind, line 72: weird, I thought tuples had to have enclosing `()`?

src/fidesops/service/connectors/query_config.py
```diff
@@ -198,10 +192,17 @@ def generate_query(
...
+                for f in self.primary_key_fields
+                if f.name in row
+            }
+        )
+        pk_clauses = [f"{k} = :{k}" for k in non_empty_primary_keys]
 
-        for pk in self.primary_keys:
-            update_value_map[pk] = row[pk]
+        for k, v in non_empty_primary_keys.items():
+            update_value_map[k] = v
```
eastandwestwind, line 205: overall can we make the above more readable, ie instead of `f` and `k` using more descriptive var names?

src/fidesops/service/connectors/query_config.py
```diff
@@ -79,14 +71,16 @@ def filter_values(self, input_data: Dict[str, List[Any]]) -> Dict[str, Any]:
...
-            if key in self.query_keys
-            and isinstance(value, list)
-            and len(value)
-            and None not in value
-        }
+
+        out = {}
+        for key, values in input_data.items():
+            field = self.node.node.collection.field(key)
+            if field and key in self.query_keys and isinstance(values, list):
```
eastandwestwind, line 78: why are we using `list` instead of `List` here, in the `isinstance` check?

src/fidesops/service/connectors/query_config.py
```diff
@@ -79,14 +71,16 @@ def filter_values(self, input_data: Dict[str, List[Any]]) -> Dict[str, Any]:
...
+                cast_values = [field.cast(v) for v in values]
```
eastandwestwind, line 79: since this method now casts, is there a more descriptive name, instead of just `filter_values`? Maybe `cast_and_filter_values`

eastandwestwind (commented):
Just a couple questions!

tests/task/traversal_data.py
stevenbenjamin: apparently not, since the IDE pushed me to do it :)

src/fidesops/util/collection_util.py
stevenbenjamin: Turns out it's never being called with None. That being said, this is a good suggestion.

src/fidesops/service/connectors/query_config.py
stevenbenjamin: As far as I know the only difference is that `List` allows for generic typing. It seems that you get equivalent results for e.g.
 `isinstance([], list)` and `isinstance([], List)`

eastandwestwind (approved):
looks good, thanks!

PR #61, opened 2021-11-11
Adds validation to fides keys, updates slugify to use underscores, closes #35

### LINKED ISSUE #35: Spike: validate model keys using `FidesKey.validate` from fideslang
We changed slugify methods to allow for underscores.  All our key fields can be `fideslang.models.FidesKey`. Need to investigate how this plays with SQLAlchemy. 

https://github.com/ethyca/fides/blob/main/fidesctl/src/fideslang/validation.py#L15
Context: https://github.com/ethyca/solon/pull/144#issuecomment-948183601

eastandwestwind:
# Purpose
Validate model keys using `FidesKey.validate`

# Changes
- Validates keys using `FidesKey.validate` (accepts alphanumeric, `_`, and `.` chars)
- Adds `FidesKey` type to appropriate params / schemas
- Adds `FidesKeyMixin` to validate schemas that use `FidesKey` fields, and converts `FidesValidationError` to `ValueError`
- Updates slugify to use `_` instead of `-` for generating keys from `name`
- Updates tests and docs to using `_` instead of `-` for keys

# Ticket
https://github.com/ethyca/fidesops/issues/35


src/fidesops/db/base_class.py
```diff
@@ -23,16 +24,16 @@
 
 def get_key_from_data(data: Dict[str, Any], cls_name: str) -> str:
     """
-    Extracts key from data and slugifies. If no key, uses a slugified name.
+    Extracts key from data, validates. If no key, uses a slugified name.
 
     Will be used as the URL slug on applicable classes.
     """
-    key = slugify(data.get("key")) if data.get("key") else None
+    key = FidesKey.validate(data.get("key")) if data.get("key") else None
```
pattisdr, line 31: This highlights the importance of reaching a consensus soon on how we're going to handle versioning and releases 
https://github.com/ethyca/fides/discussions/223.

If someone had already started using fidesops, this would be a pretty big breaking change, as they might have resources created with keys that follow separate rules.

Other places affected by this change include  `quickstart.py`, the postman collections and https://github.com/ethyca/fidesdemo.

eastandwestwind: Ah, just seeing https://github.com/ethyca/fides/discussions/223 now. Taking a look

eastandwestwind: So https://github.com/ethyca/fidesops/pull/53 is also in the same boat- we'll need to decide on versioning before merging that in, too, right?

pattisdr: An argument could be made that it's still very early and we likely don't have fidesops adopters yet (that's what we were saying about #53), so these changes can go ahead and go in.  But the sooner we get to a state where our API is stable and we're being very intentional about versioning and communicating breaking changes, the better.

eastandwestwind: Created a new ticket to make sure the other assets are updated when this is merged https://github.com/ethyca/fidesops/issues/65

seanpreston (changes requested):
This is a great start. The ask here is also to use `FidesKey` itself at the schema level when defining types, similar to: https://github.com/ethyca/fides/blob/main/fidesctl/src/fideslang/models.py#L30

It might make sense to add `key` as an optional field to the `BaseSchema` since it's used on all of our models. In the same vein, we could look at adding it to the `BaseModel` too (I wonder if this would trigger a migration).

Update: looks like it does as `key` isn't _everywhere_.

```
root@34ebd50f776f:/fidesops/src# alembic revision --autogenerate -m "test"
INFO:alembic.runtime.migration:Context impl PostgresqlImpl.
INFO:alembic.runtime.migration:Will assume transactional DDL.
INFO:alembic.autogenerate.compare:Detected added column 'client.key'
INFO:alembic.autogenerate.compare:Detected added index 'ix_client_key' on '['key']'
INFO:alembic.autogenerate.compare:Detected added column 'datasetconfig.key'
INFO:alembic.autogenerate.compare:Detected added index 'ix_datasetconfig_key' on '['key']'
INFO:alembic.autogenerate.compare:Detected added column 'executionlog.key'
INFO:alembic.autogenerate.compare:Detected added index 'ix_executionlog_key' on '['key']'
INFO:alembic.autogenerate.compare:Detected added column 'privacyrequest.key'
INFO:alembic.autogenerate.compare:Detected added index 'ix_privacyrequest_key' on '['key']'
  Generating /fidesops/src/migrations/versions/1b52322ad54c_test.py ...  done
```

It could be worth adding the keys to those places for simplicity's sake here. Setting a `key` on any object could be useful.

eastandwestwind:
@seanpreston I think I'm confused because we already have a column `fides_key` for `datasetconfig.py` model, so would adding an optional `key` that's also a `FidesKey` type be confusing?

pattisdr:
I see what you're saying @eastandwestwind: it doesn't make sense to me to add keys to all models, just resources that are likely to be reused, which is what we're already doing.  I would not expect keys on privacy requests/execution logs/or clients, and because dataset configs are really us sticking fides dataset annotations in the database, it makes sense to refer to those dataset configs by their `fides_key`, not a user-generated key.

So in my opinion, I'd keep things as-is, and just set the `key` to be a `Fideskey` type on the models that currently have a `key`.

eastandwestwind:
@pattisdr ready for you again!

docs/fidesops/docs/tutorial/execute_privacy_request.md
```diff
@@ -59,7 +59,7 @@ if __name__ == "__main__":
     email = "user@example.com"
     privacy_requests = create_privacy_request(
         email=email,
-        policy_key="example-request-policy",
+        policy_key="example_request_policy",
```
pattisdr, line 62: Related privacy request guide https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/guides/privacy_requests.md#how-can-i-execute-a-privacy-request

has a sample request referencing a policy key that needs to be snake-cased

src/fidesops/db/base_class.py
```diff
@@ -23,16 +24,16 @@
 
 def get_key_from_data(data: Dict[str, Any], cls_name: str) -> str:
     """
-    Extracts key from data and slugifies. If no key, uses a slugified name.
+    Extracts key from data, validates. If no key, uses a slugified name.
```
pattisdr, line 27: I use slugify specifically to mean hyphens as separators which is no longer true - I'd do a find/replace for every instance of slugify/slugified in our code base, including renaming the slugify function.

src/fidesops/service/connectors/base_connector.py
```diff
@@ -22,7 +22,7 @@ class BaseConnector(ABC):
     SessionLocal = get_db_session()
     db = SessionLocal()
 
-    config = db.query(ConnectionConfig).filter_by(key='my-postgres-db').first()
+    config = db.query(ConnectionConfig).filter_by(key='my_postgres_db').first()
```
pattisdr, line 25: nice spotting this usage in a docstring 🔍

docs/fidesops/docs/guides/storage.md
```diff
@@ -95,7 +95,7 @@ Example response:
                 "naming": "request_id",
                 "object_name": "requests"
             },
-            "key": "s3-storage-2"
+            "key": "s3_storage_2"
```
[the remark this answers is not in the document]
pattisdr: Also L43 in this same doc has type hinting in a sample payload and uses string instead of fideskey

```json
  {
    "destinations": [
      {
        "name": str,
        "key": str (optional),
        "type": str,
        "format": str
        "details": {
          # s3
          "bucket": str,
          "naming": str,
          # onetrust
          "service_name": str,
          "onetrust_polling_hr": int,
          "onetrust_polling_day_of_week": int
        }
      }
    ]
  }
```

src/fidesops/schemas/masking/masking_strategy_description.py
```diff
@@ -1,12 +1,13 @@
...
 
+from fideslang.validation import FidesKey
 from pydantic import BaseModel
 
 
 class MaskingStrategyConfigurationDescription(BaseModel):
     """The description model for a specific configuration in a masking strategy"""
 
-    key: str
+    key: FidesKey
```
pattisdr, line 10: This one isn't true - this is just a config to explain how masking strategies should be documented, the key here has nothing to do with a fides key.

src/fidesops/db/base_class.py
pattisdr: thanks for making this ticket @eastandwestwind!

pattisdr, line 31: This error needs to be handled in our API endpoints - if a user submits a key with a dash for example, they will see an unhandled 500 error which doesn't help them correct their request.  


Multiple options here, we could catch the FidesValidationError and reraise a ValidationError in the endpoints, or we could add this check to the schema and use a[ Pydantic validator ](https://pydantic-docs.helpmanual.io/usage/validators/) on key (only if the user supplied it, and raise a ValueError where applicable - here's an example of a validator. https://github.com/ethyca/fidesops/blob/main/src/fidesops/schemas/storage/storage.py#L161.  We could even make a key mixin to use across a lot of the schemas:

Untested, something like:
```python
class FidesKeyMixin(BaseModel):

    key: Optional[FidesKey]

     @validator('key')
     def validate_key(cls, value):
           if value is None:
                  return value 
           try:
               FidesKey.validate(value)
           except FidesValidationError as exc:
               raise ValueError(exc)
           return value
```

pattisdr (changes requested):
Thanks for working on the docs/tutorial while you were here, and reticketing the quickstart/postman collections - this change touches a lot of places!

My comments won't take long to address - big picture we need to give helpful errors to the user if the key they supply is invalid instead of throwing 500's. The use of "slugify" doesn't feel correct anymore, and there are lingering uses of hyphenated keys in docstrings and misc locations in the tutorial/guides.

src/fidesops/schemas/masking/masking_strategy_description.py
eastandwestwind: ah, thanks for catching this!

src/fidesops/db/base_class.py
eastandwestwind: I like the idea of a key mixin, because then we can re-use the same validator across multiple schemas that use a `FidesKey` as `key`. Plus we get the benefit of bundling up the `FidesValidationError` as a `ValueError` so that our api endpoints don't get too overloaded with a bunch of custom errors.

src/fidesops/schemas/shared_mixins.py
```diff
@@ -0,0 +1,30 @@
...
+
+class FidesKeyMixin(BaseModel):
+    """
+    Abstract mixin to validate fields of type FidesKey
+    Converts FidesValidationError to api-friendly ValueError
+    """
+
+    _fides_key_field_names: List[str] = ["key"]
+
+    @root_validator()
```
pattisdr, line 16: Try adding `pre=True` here -

The current solution doesn't prevent unhandled 500 exceptions because `FidesKey` itself has validation on it. Because individual fields are validated before the `root_validator` runs, you will still get 500 exceptions that were raised before this code was even hit. I believe `pre=True` lets the root validator run first.

pattisdr: Also, could you add some unit tests around invalid keys used on some of the relevant API endpoints?  Doesn't look like we have any test coverage of this specifically.

pattisdr (commented):
Just a quick drive-by comment since I realized FidesKey had its own validation, so I have a suggestion to handle - can you also get your code up to date?

src/fidesops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -155,7 +156,7 @@ def create_or_update_rules(
         verify_oauth_client,
         scopes=[scopes.RULE_CREATE_OR_UPDATE],
     ),
-    policy_key: str,
+    policy_key: FidesKey,
     db: Session = Depends(deps.get_db),
     input_data: conlist(schemas.RuleCreate, max_items=50) = Body(...),  # type: ignore
```
eastandwestwind, line 161: `RuleCreate` schema contains a `FidesKey`, so this api endpoint can throw a 422 immediately on bulk create/updates instead of gracefully adding failed policies to a list and returning `succeeded` and `failed` policies (as is the current pattern in this endpoint). @pattisdr any thoughts on whether this is OK?

tests/api/v1/endpoints/test_policy_endpoints.py
```diff
@@ -316,6 +316,32 @@ def test_create_policy_with_key(
         ).first()
         pol.delete(db=db)
 
+    def test_create_policy_with_invalid_key(
```
eastandwestwind, line 319: Here is the test that demonstrates immediately throwing a 422

pattisdr: 🏆

src/fidesops/api/v1/endpoints/policy_endpoints.py
pattisdr: This isn't introducing new behavior, this is the current pattern on our endpoints, even the bulk ones, because field-level validation happens before the code in the view runs (where we're creating the succeeded/failed responses).

For example, if you tried to create a Rule whose name was a `list` you'd also get a 422 immediately.  

In my opinion it's okay to fail early here, to save us from looping through potentially many resources that aren't formatted correctly.

eastandwestwind: Got it, this makes total sense, thanks for clarifying :)

src/fidesops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -155,7 +156,7 @@ def create_or_update_rules(
         verify_oauth_client,
         scopes=[scopes.RULE_CREATE_OR_UPDATE],
     ),
-    policy_key: str,
+    policy_key: FidesKey,
```
pattisdr, line 159: Hm, a side-effect of defining this type as a `FidesKey` means that if our URL has a non-fides key element in it, they also would get an unhandled 500 error, not sure how to handle.

```
PUT /policy/my-bad-policy/rule
```

So this is making me think about how all this discussion we've had over the last day is because `FidesKey` validation throws a custom `FidesValidationError` which pydantic/fastapi have no built-in way to work with.  I'm wondering if we should override `FidesKey` within FidesOps to throw ValueErrors and avoid all these workarounds we've had to add, and fix this issue while we're at it.

src/fidesops/schemas/privacy_request.py
```diff
@@ -22,7 +23,7 @@ class PrivacyRequestCreate(BaseSchema):
     finished_processing_at: Optional[datetime]
     requested_at: datetime
     identities: List[PrivacyRequestIdentity]
-    policy_key: str
+    policy_key: FidesKey
```
pattisdr, line 26: This is one instance that isn't handled currently - an invalid policy_key in the request would cause the privacy request to fail, but this failure would be hidden because we're catching ExceptionErrors in this endpoint for some reason, so the user would have no idea why it failed.

src/fidesops/schemas/storage/storage.py
```diff
@@ -213,13 +215,13 @@ def json_validator(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
 
-class StorageDestinationResponse(BaseModel):
+class StorageDestinationResponse(FidesKeyMixin):
     """Storage Destination Response Schema"""
 
     name: str
     type: StorageType
     details: Dict[StorageDetails, Any]
-    key: str
+    key: Optional[FidesKey]
```
pattisdr, line 224: It's not optional in the response, right? We've created the key for them if it doesn't exist and are returning it.

src/fidesops/schemas/connection_configuration/connection_config.py
```diff
@@ -27,15 +29,15 @@ class Config:
...
+class ConnectionConfigurationResponse(FidesKeyMixin):
     """
     Describes the returned schema for a ConnectionConfiguration.
 
     Do *NOT* add "secrets" to this schema.
     """
 
     name: str
-    key: str
+    key: Optional[FidesKey]
```
pattisdr, line 40: Also as a Response, we create the key, so it should exist.

pattisdr (commented):
Did one more pass @eastandwestwind - getting close, maybe we can simplify this by changing what FidesKey raises internally

src/fidesops/schemas/storage/storage.py
eastandwestwind: right, I'll fix this!

src/fidesops/api/v1/endpoints/policy_endpoints.py
eastandwestwind: I'm OK with this approach. 4xx makes a lot more sense than a 5xx in this context, and it's pretty low overhead to just override `FidesKey` to throw `ValueError`

src/fidesops/schemas/dataset.py
```diff
@@ -62,14 +62,18 @@ class FidesCollectionKey(ConstrainedStr):
...
+                len(values) == 2
+                and FidesOpsKey.validate(values[0])
+                and FidesOpsKey.validate(values[0])
+            ):
+                return value
+        except ValueError:
+            raise ValueError(
+                "FidesCollection must be specified in the form 'FidesKey.FidesKey'"
+            )
+        raise ValueError(
```
eastandwestwind, line 76: I'm unsure the exact use case for this validator, so unsure if using `ValueError` is the correct approach here

pattisdr: @eastandwestwind this gets hit in the `test_dataset_yaml_format` - we could add a test that looks at the failing cases: where 1) the FidesCollectionKey is in the incorrect format, "hello", for example (because we expect a dataset name and collection name combined with a period).  and 2) if the individual pieces aren't fides keys "bad-dataset-name:bad-collection-name"

src/fidesops/schemas/dataset.py
```diff
@@ -62,14 +62,18 @@ class FidesCollectionKey(ConstrainedStr):
...
-            and FidesKey.validate(values[0])
-        ):
-            return value
-
-        raise FidesValidationError(
+        try:
+            if (
+                len(values) == 2
+                and FidesOpsKey.validate(values[0])
+                and FidesOpsKey.validate(values[0])
```
pattisdr, line 69: this was already like this, but this looks like a bug, we should be checking the second piece too [1]

src/fidesops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -6,6 +6,7 @@
 from fastapi_pagination import Page, Params
 from fastapi_pagination.bases import AbstractPage
 from fastapi_pagination.ext.sqlalchemy import paginate
+from fidesops.schemas.shared_schemas import FidesOpsKey
```
pattisdr, line 9: Importing FidesOpsKey twice

src/fidesops/schemas/dataset.py
```diff
@@ -62,14 +62,18 @@ class FidesCollectionKey(ConstrainedStr):
...
+            ):
+                return value
+        except ValueError:
+            raise ValueError(
+                "FidesCollection must be specified in the form 'FidesKey.FidesKey'"
```
pattisdr, line 74: There's a bug here, you're catching the ValueError and re-raising with a message that isn't accurate, I'd do something like this:

```python
    @classmethod
    def validate(cls, value: str) -> str:
        """
        Overrides validation to check FidesCollectionKey format, and that both the dataset
        and collection names have the FidesKey format.
        """
        values = value.split(".")
        if len(values) == 2:
            FidesOpsKey.validate(values[0])
            FidesOpsKey.validate(values[1])
            return value
        raise ValueError(
            "FidesCollection must be specified in the form 'FidesKey.FidesKey'"
        )

```

src/fidesops/schemas/dataset.py
```diff
@@ -62,14 +62,18 @@ class FidesCollectionKey(ConstrainedStr):
     def validate(cls, value: str) -> str:
```
pattisdr, line 62: `# This overrides the default method to throw the custom FidesValidationError`
no longer accurate

pattisdr (commented):
ok @eastandwestwind back to you! one more round needed because of an introduced bug, and the code needs updating again, but then I think we're good to go.

Thanks for all your hard work on this, working through different approaches.

src/fidesops/schemas/dataset.py
eastandwestwind: ohhh good point, the only case that pertains to this error message is when `len(values) != 2`

pattisdr (approved):
looks good to me @eastandwestwind 🏅

Issue #46, opened 2021-11-06
Allow a `PrivacyRequest` level encryption key to be supplied on request creation


seanpreston:
Fidesops must support the ability to supply a custom encryption key on request creation, that will be used throughout the PrivacyRequest flow. This encryption key should be used to encrypt the access data we return to the user at the end of a request before it is uploaded in either JSON or CSV format.

This encryption key can be ignored for erasure policies for now.

pattisdr:
Pieces we still need to settle on:

- How will the encryption key be specified in the request? Old planning docs suggest it should be sent as a request header using a custom ethyca header parameter, but this might be outdated.
-  How do we temporarily store the encryption key? it could be cached in Redis alongside the identities or stored encrypted itself on the PrivacyRequest, I like the latter.
- What encryption algorithm do we want to use here? We already have support for AES GCM Encryption, but there's this bit about the nonce, if we are generating it ourselves (in which case, we need to give this to the user somehow), or they are supplying that too.

pattisdr:
Updates:

- How we will specify encryption key: Via a request header with the custom parameter `x-fidesops-encryption-key`
- Encryption key will be temporarily cached like identities
- We will use AES GCM to encrypt, requiring use to supply both nonce and encryption key.

PR #66, opened 2021-11-12
Optionally Encrypt Access Request Data [#46][#83]

### LINKED ISSUE #83: AES Encryption/Decryption Endpoints Broken
Our AES GCM Encryption / Decryption endpoints do not work, partially because unit tests mock important internals.

## Expected
A PUT to ` /cryptography/encryption/aes/encrypt` with payload:

```json
{
    "key": "encryptionkeytest",
    "value": "test data",
}
```
This should return an encrypted value and the nonce used to encrypt.

## Issues:
1) The nonce we internally generate is the incorrect length, which fails:
  a) We generate a nonce using `cryptographic_util.generate_secure_random_string` which builds a string that is twice as long as the input we request.  We specify 12, and it returns a string that is 24 characters long.  Later, before we use that nonce, we require that it is 12 bytes long.

2) Our internal AES GCM encrypt/decrypt require bytes-like data, and we just pass in strings from the request body without encoding them, same issue on the decrypt endpoints

3) AES GCM encrypt/decrypt needs the encryption key to be converted to bytes, we're just trying to pass the string through.

4) Better error handling needed - if there is an error with the key that is supplied, the user will get an unhandled 500

pattisdr:
# Purpose

Fidesops must support the ability to supply a custom encryption key on request creation, that will be used throughout the PrivacyRequest flow. This encryption key should be used to encrypt the access data we return to the user at the end of a request before it is uploaded in either JSON or CSV format.

# Changes

- Use optional `encryption_key` field supplied with the Privacy Request body to trigger encrypting data (for local and s3 uploads only)
  - Temporarily store the `encryption_key` in the Redis cache
- Encrypt the contents of the results upload (for JSON and CSV files) using AES GCM encryption where applicable - concatenates the internally generated nonce with the encrypted data as "nonce" + encrypted_data.  
  - If CSV data is specified, each csv in the zipfile will be encrypted using a different nonce
- Use `secrets.token_bytes()` to generate the nonce
  - Do this same thing for our encryption endpoints
- Add additional fixes to encryption endpoints
# Ticket
Resolves https://github.com/ethyca/fidesops/issues/46
Resolves #83


src/fidesops/service/storage/storage_uploader_service.py
```diff
@@ -111,15 +105,4 @@ def _local_uploader(
...
-        json.dump(data, f, default=_handle_json_encoding)
-    return "success"
-
-
-def _handle_json_encoding(field: Any) -> str:
-    """Specify str format for datetime objects"""
-    if isinstance(field, datetime):
-        return field.strftime("%Y-%m-%dT%H:%M:%S")
-    return field
+    return upload_to_local(data, file_key, request_id)
```
pattisdr, line 108: This was just moved for circular import reasons when I was adding encryption to local storage -

src/fidesops/schemas/privacy_request.py
```diff
@@ -19,6 +23,31 @@ class PrivacyRequestCreate(BaseSchema):
...
+        nonce = values.get("nonce")
+        if nonce or encryption_key:
+            if not (nonce and encryption_key):
+                raise ValueError(
+                    "To encrypt access request data, both an encryption key and a nonce must be supplied."
+                )
+
+            # Validate supplied encryption key and nonce are acceptable
+            verify_encryption_key(encryption_key)
+            verify_nonce(nonce)
```
NevilleS, line 48: I'd prefer if we don't *require* a nonce; I'd generally not expect to provide one to a system as they are meant to be random, temporary, and never re-used. It's a bit *too easy* for a caller to just hardcode a nonce and provide it with every request so I'd like to discourage that by making the default case where we generate the nonce.

That said, we will also need to change the encrypted payload to prepend the nonce (aka initialization vector, aka IV). It's super common to just do `IV || ciphertext` as the output, so we can follow that pattern - I can show an example of how this is implemented in sqlalchemy for reference...

NevilleS (changes requested):
This is a solid implementation, but I'd like to make it *slightly* more complex by generating the nonce on the server-side and then encoding that into the payload. See comment and LMK if you want to discuss further 👍

src/fidesops/schemas/privacy_request.py
NevilleS: Sorry I missed some of the requirements discussion about this on #46!

src/fidesops/models/privacy_request.py
```diff
@@ -96,6 +97,23 @@ def cache_identity(self, identity: PrivacyRequestIdentity) -> None:
                     value,
                 )
 
+    def cache_encryption(
+        self, encryption_key: Optional[str] = None, nonce: Optional[str] = None
+    ) -> None:
+        """Sets the encryption key/nonce in the Fidesops app cache if provided"""
+        if not (encryption_key and nonce):
```
stevenbenjamin, line 104: I think this is unclear on bad input. e.g, it's probably more appropriate to raise an exception  if only one of these two is provided, since that's invalid input, which leads me to think that it'd be better to check this at the call site at `create_privacy_request`,

src/fidesops/schemas/privacy_request.py
pattisdr: 👍 reworking so we generate the nonce.

src/fidesops/tasks/storage.py
```diff
@@ -21,38 +31,79 @@
...
+        data = data.decode(config.security.ENCODING)
+
+    encryption_key = cache.get(encryption_cache_key)
+    if not encryption_key:
+        return data
+
+    encryption_key = encryption_key.encode(encoding=config.security.ENCODING)
+    nonce = cryptographic_util.generate_secure_random_string(
+        math.floor(config.security.AES_GCM_NONCE_LENGTH / 2)
+    )
```
pattisdr, line 54: This feels a little weird - our `generate_secure_random_string` method currently generates a string that is twice as long as the integer we use as the input, but the `verify_nonce` method used by our internal aes gcm `encrypt` requires that the nonce be `AES_GCM_NONCE_LENGTH` in length, so I halve that value here.

src/fidesops/schemas/privacy_request.py
pattisdr: @NevilleS @stevenbenjamin finished a refactor so we generate the nonce on our end, thanks for the comments -

Also added docs in the latest commit for how to decrypt.

src/fidesops/tasks/storage.py
pattisdr: This is related to https://github.com/ethyca/fidesops/issues/83, for a similar reason our AES Encryption/Decryption endpoints are broken.

NevilleS: I've run into this exact problem with the `secrets.token_***` methods in other projects, they're slightly confusing. For `token_hex` in particular (which is what our crypto util is using) it generates a N-byte value and then converts that to hex- with two hex characters per byte. So what you really need to do is probably just use `secrets.token_bytes`, for example: https://docs.python.org/3/library/secrets.html

As an alternative, note that `os.urandom` might be a better fit for this as it's what most projects seem to use for Python crypto, but that might just be due to timing - I think the `secrets` module is a newer addition that is designed to replace the need for that.

NevilleS: Also FYI - the nonce length *needs* to be that exact length for the AES algorithm to work (if you go ahead and try it with other lengths and remove that `verify_nonce` check it should throw an error!). For that reason, I don't think it really should be a config option either.

pattisdr: summary of separate discussion, I'm now using `secrets.token_bytes` and using b64 to encode the nonce and the encrypted data together to return to the user, thanks @NevilleS 

```python
iv = secrets.token_bytes(12)
ciphertext = <insert binary string here>
output = b64encode(iv + ciphertext)
```

```python
decoded = b64decode(output)
iv = decoded[0:12]
cipherext = decoded[12:]
```

docs/fidesops/docs/guides/privacy_requests.md
```diff
@@ -60,3 +62,59 @@ Alongside generic API interopoerability, Fidesops provides a direct integration
...
+        "encryption_key": "test--encryption"
+    }
+]
+
+```
+
+## Decrypting your access request results
+
+If you specified an encryption key, we encrypted the access result data using your key and an internally-generated `nonce` with an AES 
+algorithm in GCM mode.  The return value is a 12-byte nonce plus the encrypted data that is all b64encoded together.
```
NevilleS, line 90: I would add a visual example here, like:
```
+------------------+-------------------+
| nonce (12 bytes) | message (N bytes) |
+------------------+-------------------+
```

docs/fidesops/docs/guides/privacy_requests.md
```diff
@@ -60,3 +62,59 @@ Alongside generic API interopoerability, Fidesops provides a direct integration
 
 - Generic API interoperability: Third party services can be authorized by creating additional OAuth clients. Tokens obtained from OAuth clients can be managed and revoked at any time. Pleae see [How-To: Authenticate with OAuth](oauth.md) for more information.
 - OneTrust: Fidesops can be configured to act as (or as part of) the fulfilment layer in OneTrust's Data Subject Request automation flow. Please see [How-To: Configure OneTrust Integration](onetrust.md) for more information.
+
+
+## Encryption
+
+You can optionally encrypt your access request results by supplying an `encryption_key` string in the request body:
+We will use the supplied encryption_key to encrypt the contents of your JSON and CSV results using an AES algorithm in GCM mode.
```
NevilleS, line 70: Let's specify that it is AES-256

src/fidesops/util/cryptographic_util.py
```diff
@@ -11,11 +14,22 @@ def hash_with_salt(text: bytes, salt: bytes) -> str:
...
+    that is twice the length of the specified input"""
+    return secrets.token_hex(length)
 
 
 def generate_salt() -> str:
     """Generates a salt using bcrypt and returns a string using the configured default encoding"""
     return bcrypt.gensalt().decode(config.security.ENCODING)
+
+
+def bytes_to_str(bytestring: bytes) -> str:
```
NevilleS, line 28: I was a bit surprised that this returned a b64-encoded string, considering that Python allows you to just tuck a `bytes` array into a `str` instance whenever you like.

For that reason, maybe consider calling these helpers `bytes_to_b64_str` and `b64_str_to_bytes`?

NevilleS (approved):
I didn't review all the API and test parts, but the core security bits of the encryption & decryption look sound to me. Thanks for the thoroughness here

docs/fidesops/docs/guides/privacy_requests.md
pattisdr: good idea!

src/fidesops/util/cryptographic_util.py
pattisdr: oh yes that's much more clear

pattisdr:
thank you @stevenbenjamin !

PR #68, opened 2021-11-15
closes #36- masks params provided in % str formatter in logs, unless set otherwise

### LINKED ISSUE #36: Configurable application logging to toggle PII on/off for testing
Add additional logging configurations that would allow PII to be displayed for testing purposes.

Context: https://github.com/ethyca/solon/pull/143#discussion_r734639625

eastandwestwind:
# Purpose
Allows logging pii to be configurable via env variable

# Changes
- Removes superfluous `logging.basicConfig()` statements, since it only needs to be set at root logger level (`src/fidesops/main.py`). More details here- https://docs.python.org/3/library/logging.html#logging.basicConfig
- Adds custom log factory that masks pii values by default (see https://docs.python.org/3/library/logging.html#logging.LogRecord.getMessage), unless one of the following is true:
1. `LOG_PII` env var is set to `True`
2. The arg is explicitly white-labeled as `NotPii`
- Adds tests for custom log factory ^

# Ticket
https://github.com/ethyca/fidesops/issues/36

### DIFF
--- a/src/fidesops/util/logger.py
+++ b/src/fidesops/util/logger.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from typing import Any, Mapping, Union
+from typing import Any, Mapping, Union, Optional
 
 MASKED = "MASKED"
 
@@ -23,14 +23,16 @@ def get_fides_log_record_factory() -> Any:
         func: str = None,
         sinfo: str = None,
     ) -> logging.LogRecord:
-        masked_args: tuple[Any, ...] = tuple(_mask_pii_for_logs(arg) for arg in args)
+        env_log_pii: bool = os.getenv("LOG_PII") == "True"
+        if not env_log_pii:
+            args: tuple[Any, ...] = tuple(_mask_pii_for_logs(arg) for arg in args)
         return logging.LogRecord(
             name=name,
             level=level,
             pathname=fn,
             lineno=lno,
             msg=msg,
-            args=masked_args,
+            args=args,
             exc_info=exc_info,
             func=func,
             sinfo=sinfo,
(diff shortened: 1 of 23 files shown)


eastandwestwind:
@stevenbenjamin I made the assumption that the params in https://github.com/ethyca/fidesops/blob/main/src/fidesops/service/connectors/query_config.py#L213 can contain Pii, so I didn't whitelist these. Is that OK?

src/fidesops/util/logger.py
```diff
@@ -0,0 +1,51 @@
...
+        level: int,
+        fn: str,
+        lno: int,
+        msg: str,
+        args: Union[tuple[Any, ...], Mapping[str, Any]],
+        exc_info: Any,
+        func: str = None,
+        sinfo: str = None,
+    ) -> logging.LogRecord:
+        masked_args: tuple[Any, ...] = tuple(_mask_pii_for_logs(arg) for arg in args)
```
stevenbenjamin, line 26: This is a minor optimization, but this is going to be called _a lot_ . This calls a string comparison check with os.getenv == "True" on each parameter, be faster to do something like
```
 show_pii = os.getenv("LOG_PII") == "True"
 if not show_pii:
    ... mask ...
    ```

**Review by stevenbenjamin — CHANGES_REQUESTED** [review]
This is a really nice implementation, gives us PII masking with very limited added code complexity. 

I did find a number of places where we're raw-logging an exception, I think those might contain PII and should get masked (since we have no control over what a custom exception from another library might contain):

```
src/fidesops/core/config.py:240:            logger.error(exc)
src/fidesops/models/storage.py:106:            logger.error(exc)
src/fidesops/db/session.py:51:            logger.error(exc)
src/fidesops/api/v1/endpoints/connection_endpoints.py:194:        logger.warning(f"Connection test failed on {connection_config.key}: {str(exc)}")
src/fidesops/api/v1/endpoints/privacy_request_endpoints.py:140:            logger.error(exc)
src/fidesops/api/v1/endpoints/privacy_request_endpoints.py:147:            logger.error(exc)
src/fidesops/api/v1/endpoints/policy_endpoints.py:123:            logger.warning(f"Create/update failed for policy: {exc}")
src/fidesops/api/v1/endpoints/policy_endpoints.py:131:            logger.warning(f"Create/update failed for policy: {exc}")
```

PR #72, opened 2021-11-17
Add length field [#71]

### LINKED ISSUE #71: Add length member to Field object
Fields will need an optional length member to ensure that generated values will be insertable.

stevenbenjamin:
# Purpose
Add a length field to Field object
# Changes
- adds a length member
- validates that if exists valid length member > 0
- api endpoint test

# Ticket
Closes #71

### DIFF
--- a/src/fidesops/models/datasetconfig.py
+++ b/src/fidesops/models/datasetconfig.py
@@ -87,6 +87,7 @@ def _convert_dataset_field_to_graph(field: FidesopsDatasetField) -> Field:
     is_pk = False
     references = []
     meta_section = field.fidesops_meta
+    length = None
     if meta_section:
         identity = meta_section.identity
         if meta_section.primary_key:
@@ -117,13 +118,15 @@ def _convert_dataset_field_to_graph(field: FidesopsDatasetField) -> Field:
                     reference.dataset, ref_collection, ".".join(ref_fields)
                 )
                 references.append((address, reference.direction))
+        if meta_section.length is not None:
+            length = meta_section.length
     return Field(
         name=field.name,
         data_categories=field.data_categories,
         identity=identity,
         references=references,
         primary_key=is_pk,
-        length=field.length,
+        length=length,
     )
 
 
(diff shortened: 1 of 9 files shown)


src/fidesops/schemas/dataset.py
```diff
@@ -79,6 +91,7 @@ class FidesopsDatasetField(DatasetField):
     """Extends fideslang DatasetField model with additional Fidesops annotations"""
 
     fidesops_meta: Optional[FidesopsMeta]
+    length: Optional[int]
```
NevilleS, line 94: Would prefer we stash all these extensions in the "meta" object instead

NevilleS (changes requested):
Nitpick :)

src/fidesops/models/datasetconfig.py
```diff
@@ -117,12 +118,15 @@ def _convert_dataset_field_to_graph(field: FidesopsDatasetField) -> Field:
                     reference.dataset, ref_collection, ".".join(ref_fields)
                 )
                 references.append((address, reference.direction))
+        if meta_section.length is not None:
```
seanpreston, line 121: This check isn't doing anything. The result is the same if we just have `length = meta_section.length` like below, without the check.

src/fidesops/schemas/dataset.py
stevenbenjamin: This has been moved

src/fidesops/models/datasetconfig.py
seanpreston: This check will still pass along `0` values right? `0` is not `None`, so length will still be assigned in that case. The check for truthiness is `if meta_section.length:`

stevenbenjamin: Thought a bit more about this, and now I think I should leave it, with perhaps a comment - 

Reason is that this method should pass along a value if it is a valid int. If there’s a 0 passed in, that’s not the same as a None being passed in and it should get passed through to the Field object creation. So, we pass the value in if value is not None
In a way this is kind of moot since in the validations I’m explicitly rejecting any values <= 0, but I think that in the future it’s possible to imagine some kind of weird use case where we’d actually want to allow you to set a 0 length (maybe you want empty strings? ) and if we decided we wanted to support that, we could just change the validation and we wouldn’t have this weird buried filter that wouldn’t let the 0's go through in another part of the code.

Issue #26, opened 2021-11-02
Enabling CI for forked pull requests


pattisdr:
# What

PR's originating from my fidesops fork https://github.com/pattisdr/fidesops did not have CI run for them,[ it looks like by default, CircleCI does not build PR's from forked repositories](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories).

- https://github.com/ethyca/fidesops/pull/22
- https://github.com/ethyca/fidesops/pull/24
- https://github.com/ethyca/fidesops/pull/25

There may be some configuration to figure out here, if certain secrets need to be hidden, or to only build pull requests instead of every commit, etc.

# Why 
Allows the test suite to run for outside contributors.

iamkelllly:
For circle ci: as Dawn called out, [instructions here](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories)
For travis: we need to include a `.travis.yml` file containing the settings and build instructions, need to provide documentation for this
For github actions: users will need to [enable workflows from private repository forks](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks)

So I'm thinking this issue only needs to be adding a docs page to the documentation to include "How to enable CI for your forked repo". Any thing else you think I need @pattisdr ?

pattisdr:
To be clear, I think it's **our CI** that needs to run when a contributor submits a PR from their fork to our repo, so we need to flip some of these settings on our end.

PR #86, opened 2021-11-18
trigger workflows to run for pull_request_target event

seanpreston:
# Purpose
This PR adds the `pull_request_target` event to our workflows, which should allow PRs on forked repos to trigger workflow events (ie. CI via Github Actions).

See: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target for full docs.

I believe this change has to be made in the repo itself, before any workflows on forked PRs will run.

As @iamkelllly has mentioned here: https://github.com/ethyca/fidesops/issues/26#issuecomment-966340353, there's also something we'll need to do with the Github tokens, to stop them from being able to read secrets being stored in the underlying repository, so in lieu of completing that here I've changed the repository settings to force us to approve any workflow runs for outside collaborators, which we can choose to not do until we're sure the token issue is handled.

Screenshot 2021-11-18 at 11 28 21

# Ticket
Closes https://github.com/ethyca/fidesops/issues/26

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -1,5 +1,8 @@
 name: Run CI
-on: [push]
+on: [
+  push,
+  pull_request_target
+]
 
 jobs:
   Autoformat:


NevilleS:
I'd love a few others to weigh in here on the security side (paging @ThomasLaPiana and @PSalant726) just so we get this right. I like the idea of allowing forked PRs to run our actions for CI, but want to ensure we are very careful about how that can expose the GitHub secrets.

A couple articles I found on this:
* https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
* https://securitylab.github.com/research/github-actions-untrusted-input/
* https://securitylab.github.com/research/github-actions-building-blocks/

On my reading I think I'd prefer:
* using `pull_request` (instead of `pull_request_target`) as the trigger, which should be able to build & run unit tests without any secrets... I think?
* only ever using the docker secrets on `main`, never on any `pull_request` or `pull_request_target` workflows
* having a separate workflow for the integration tests. this can be manual for now (i.e. a core member checks out and runs the code), but we could also make it so you add a label like `safe-to-test` and that allows integration tests to run- see the example from here https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
```
    on:
      pull_request_target:
        types: [labeled]

    jobs:
      build:
        name: Build and test
        runs-on: ubuntu-latest
        if: contains(github.event.pull_request.labels.*.name, 'safe to test')
```

.github/workflows/pr_checks.yml
```diff
@@ -1,5 +1,8 @@
 name: Run CI
-on: [push]
+on: [
+  push,
+  pull_request_target
```
NevilleS, line 4: TL;DR from my comment (https://github.com/ethyca/fidesops/pull/86#issuecomment-972983287)

If you set this to `pull_request` instead of `pull_request_target` for now, we'll get 90% of the value of CI but I'd expect tests that rely on secrets to fail. That's a secure place to start from

NevilleS (changes requested):

seanpreston:
> If you set this to pull_request instead of pull_request_target for now, we'll get 90% of the value of CI but I'd expect tests that rely on secrets to fail. That's a secure place to start from

Will update.

> having a separate workflow for the integration tests. this can be manual for now (i.e. a core member checks out and runs the code), but we could also make it so you add a label like safe-to-test and that allows integration tests to run- see the example from here https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

I'd like to understand more about how the tokens are issued too. If it's the case that the tokens are short lived and can be set to only be generated on manual approval of a job run, it might not be a huge issue either way.

PSalant726 (changes requested):
@NevilleS 's suggestion about configuring the external integration tests to only run against PR's labeled `safe-to-test` (or similar) seems like a good potential route forward.

As another option, we could create an additional, similar automation that restricts the execution of the external integration tests only to PRs opened either 1) from a branch off of the repo itself (not a fork) or 2) by a member of the Ethyca GH org and/or the `@ethycadev` user group.

I don't see any way for us to safely automate the execution of any job that requires access to repository secrets when the PR being tested was opened by an untrusted user. For PRs opened by external contributors, I would recommend that an Ethyca employee first reviews the code changes here on GH, then, if everything seems safe, checkout the fork's branch locally and execute the external integration tests manually. Since all PRs require an approval from a maintainer before merging anyway, this extra step doesn't seem like _that_ much to ask.

As a last step, we should find a way to mark the external integration tests as "not run", and prevent the pipeline from displaying as passed, so we don't miss the previous manual step.

NevilleS:
From what I can see, the safe approach here for forks is to either do a
manual checkout or use something that only we approve, ie labels or some
more complex action like the `ok-to-test` action that is also linked in one
of the GitHub security articles. But yeah running our regular CI tests
without secrets is OK with me - with the collaborator approval that is,
just to also defend against spurious PRs costing us a lot of action minutes
😁

seanpreston:
To be thorough I have done some testing this afternoon by opening up a PR against this repo from a new account. I can confirm that the new account is unable to add labels:

Screenshot 2021-11-18 at 18 03 57

iamkelllly:
Given that we're referencing `pull_request` rather than `pull_request_target` then, can we update the PR title @seanpreston ?

seanpreston:
> As a last step, we should find a way to mark the external integration tests as "not run", and prevent the pipeline from displaying as passed, so we don't miss the previous manual step.

@PSalant726, it looks like this happens automatically:

Screenshot 2021-11-19 at 12 17 55

seanpreston:
I've updated this PR to factor in a couple of changes requested by both @NevilleS and @PSalant726:

- CI checks are now split into Safe and Unsafe categories
- Safe CI checks are run on every `pull_request` event. This allows forked PRs to run them (?), and is also triggered every time a commit is added to the PR (as before).
- Unsafe CI checks are triggered by the `labeled` type of `pull_request` events, and will only run if the label added contains the text `run unsafe ci checks`. NB. this category will still show a `skipped` label in the list of checks for `push` events, so the reviewer is aware that these checks have not run at the time of review.

.github/workflows/unsafe_pr_checks.yml
```diff
@@ -0,0 +1,20 @@
+name: CI Checks - Unsafe
+on:
+  push:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  Integration-Tests-External:
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'run unsafe ci checks')
```
seanpreston, line 10: This condition was a bit confusing to me on first pass, so to break down for anyone else:

`github.event.pull_request.labels.*.name` this statement is applying a filter on the `github.event.pull_request.labels` var, pull out a list of label names, e.g. `["safe to run", "dont merge", ...]`.

`contains(list, "value")` is then check if `"value"` is an element of `list`.

It seems simple when you put it that way, but if you don't know about the filter stage, it could look as though `github.event.pull_request.labels.*.name` evaluates to a string and we're checking if that string contains the PR name (which would be less robust).

https://docs.github.com/en/actions/learn-github-actions/expressions#object-filters

src/fidesops/main.py
```diff
@@ -26,6 +26,8 @@
 
 app.include_router(api_router)
 
+# abritrary change
+
```
NevilleS, line 30: This can be removed pre-merge

NevilleS (approved):
This looks good to me now.

However, with the number of actions we're running, I think we need to combine some of these jobs together to make the whole thing run faster because we're definitely burning a ton of CI minutes on every single PR

seanpreston:
> However, with the number of actions we're running, I think we need to combine some of these jobs together to make the whole thing run faster because we're definitely burning a ton of CI minutes on every single PR

The number of minutes burned might actually go down with this chance, since we're switching from running the workflows on every `push` to only events related to pull requests.

PSalant726 (approved):
@NevilleS -
> However, with the number of actions we're running, I think we need to combine some of these jobs together to make the whole thing run faster because we're definitely burning a ton of CI minutes on every single PR

Ideally the pipeline would:
- Build the container artifacts in a `build` stage, and make them available to the other stages that require them
- Execute static analysis tools and test harnesses in discreet pipeline `lint` and `test` stages, dependent on the build artifacts where needed (and, crucially, _not_ dependent where not needed)

We have an [issue in fidesctl](https://github.com/ethyca/fides/issues/228) to make a similar change.

@seanpreston -
It's not enough to block this PR for me, but I think a cleaner implementation would keep things in a single file with a generic name like `Fidesops`. The `jobs` are still discreet and can have `if` conditions applied independently.

seanpreston:
> It's not enough to block this PR for me, but I think a cleaner implementation would keep things in a single file with a generic name like Fidesops. The jobs are still discreet and can have if conditions applied independently.

I did have it like this initially. Because we're using both `push` and `pull_request` triggers, we'd end up needing custom `if: ...` conditionals on every job to be sure we're not running every job in both contexts. That was generating 15 job runs for every push which seemed excessive. There might be another way to configure it that I didn't discover, but this felt clean enough, and has the added benefit of anything unsafe being explicit and in its own file.

PR #89, opened 2021-11-18
log fix

### LINKED ISSUE #88: Stack traces caused by log masking
Log is printing out long traces from uvicorn. This is because our log masking is whitelisting
Sequence:

- fidesops.main calls uvicorn.run
- this calls a log message within uvicorn that is intended to print an int. That is, the log call is written with an int format string, that will fail with a string, e.g. `logger.warning("%d", int_parameter)
- this runs with our logging, which replaces `int_parameter` with the string `MASKED`, which throws a log format error>

sample trace excerpt:
```
 File "/usr/local/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 463, in send
self.access_logger.info(
Message: '%s - "%s %s HTTP/%s" %d'
Arguments: ('MASKED', 'MASKED', 'MASKED', 'MASKED', 'MASKED')

```

stevenbenjamin:
# Purpose
Fixes errors raised in uvicorn (and any consumed library) that uses our logger when our masking logger improperly converts values to strings.

# Changes
- logger does not mask any numeric types. 

# Ticket
Closes #88

### DIFF
--- a/src/fidesops/util/logger.py
+++ b/src/fidesops/util/logger.py
@@ -51,6 +51,6 @@ def _mask_pii_for_logs(parameter: Any) -> Any:
     Don't mask numeric values as this can throw errors in consumers
     format strings.
     """
-    if isinstance(parameter,(NotPii,Number)):
+    if isinstance(parameter, (NotPii, Number)):
         return parameter
     return MASKED
--- a/src/fidesops/util/logger.py
+++ b/src/fidesops/util/logger.py
@@ -51,6 +51,6 @@ def _mask_pii_for_logs(parameter: Any) -> Any:
     Don't mask numeric values as this can throw errors in consumers
     format strings.
     """
-    if isinstance(parameter, NotPii) or isinstance(parameter, Number):
+    if isinstance(parameter,(NotPii,Number)):
         return parameter
     return MASKED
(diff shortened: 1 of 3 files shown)


src/fidesops/util/logger.py
```diff
@@ -42,11 +43,14 @@ def factory(  # pylint: disable=R0913
...
-    :return: depending on ENV config, returns masked pii param
+    :param parameter: param that contains possible pii
+    :return: depending on ENV config, returns masked pii param.
+
+    Don't mask numeric values as this can throw errors in consumers
+    format strings.
     """
-    if isinstance(pii_text, NotPii):
-        return pii_text
+    if isinstance(parameter, (NotPii, Number)):
```
seanpreston, line 54: This looks good for now, but I wonder about the case where PII could eventually be a number. A credit card number for instance would be a good example here, and currently that wouldn't be masked.

stevenbenjamin: Agreed, with a few caveats - 
A CC should never be a number, since it can get mangled if it starts with an '0'. That being said, people do weird things in databases. It's also possible that something like an int ID might be PII since it could identify you in another system.

Ultimately this was a solution I came up with to quick-patch the issue. I think the _correct_ solution would be to switch from whitelist to blacklist - wouldn't really be much more code for us, and would not affect any embedded libraries like whitelist does. That's a bigger fix, but I think we should put it on the board.

PR #92, opened 2021-11-19
Refactor CI checks to run as a single job

### LINKED ISSUE #90: Optimize CI actions to use fewer minutes (/w less parallelization or more caching)
Right now our CI has been very helpful, but as we continue to add more tests some of the inherent inefficiencies are really adding up!

After #86 lands our checks will look something like this:
image
image

If I'm following correctly, I think that's 8 jobs, totaling 33 minutes of CI actions for one PR 😅 

Of course our actual actions (tests, linters, etc.) are quite fast - but the slow part is that each of these runners needs to `docker pull` all the dependencies and build the images every time. There's no caching between actions either so we pay that tax on every job.

Options here are:
1. Combine multiple jobs into a single job, so that we can reuse the Docker images across them
2. Setup Docker caching in our CI (I think this looks right: https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching)

NevilleS:
# Purpose
This allows the docker build cache to be shared for each check, which is the majority of the build time. This will likely reduce CI usage by a factor of 7, until we have a solution that allows running in parallel with a shared docker cache.

# Changes

- Run CI checks on `push` command (so that checks are run more than once per PR)
- Combine all checks into a single job, with multiple steps
- Rename `external_integration` to `integration_external` pytest mark for consistency with others (e.g. `integration_erasure`)

# Ticket
Closes #90

### DIFF
--- a/.github/workflows/safe_pr_checks.yml
+++ b/.github/workflows/safe_pr_checks.yml
@@ -12,6 +12,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Docker Compose Build
+        run: make compose-build
+
       - name: Format
         run: make black-ci
 
--- a/.github/workflows/safe_pr_checks.yml
+++ b/.github/workflows/safe_pr_checks.yml
@@ -1,8 +1,9 @@
 name: CI Checks - Safe
-on: [
-  push,
-  pull_request
-]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   Check-All:
--- a/.github/workflows/safe_pr_checks.yml
+++ b/.github/workflows/safe_pr_checks.yml
@@ -1,69 +1,36 @@
 name: CI Checks - Safe
 on: [
+  push,
   pull_request
 ]
 
 jobs:
-  Autoformat:
+  Check-All:
     runs-on: ubuntu-latest
     steps:
-      - run: echo "Running CI for branch ${{ github.ref }}."
-      - name: Check out repository code
+      - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


seanpreston:
We split these out intentionally a while back so we could have granularity on which part of CI was failing from this view. I'm not against changing it back, but let's make sure we're going for exactly what we want here. As I understand it the motivation here is to save the wasted build time between jobs?

NevilleS:
@seanpreston: yes, see my notes in #90 but we're basically taking 30m of CI time on every commit, which adds up fast. With this refactor, it looks like it's closer to about 9-10m. For example see this action: https://github.com/ethyca/fidesops/runs/4267532454?check_suite_focus=true
image

The "Format" step there shows as taking over 3m30 but that's because it's the first one that does a docker build. Everything else is much faster. However, if you parallelize this, you "pay" that build on every individual step so it becomes very long.

I think it's still _reasonably_ easy to see what step fails (from that "Details" view). We can parallelize the jobs again with some caching if y'all can figure out a good solution for that too.

NevilleS:
OK, this is ready for review. I think this strikes an OK balance for now. The before/after is approximately:
| workflow | total time (before) | total time (after) |
|---|---|---|
| safe checks | 29m | 8m |
| unsafe checks | 4m | 4m |
| total | 33m | 12m |

Additionally, I think I've setup both safe/unsafe to run on every merge to `main`, which I like since that'll trigger the full integration suite post-merge for PRs where we don't manually add the `run unsafe ci checks` label.

seanpreston:
> I think it's still reasonably easy to see what step fails (from that "Details" view). We can parallelize the jobs again with some caching if y'all can figure out a good solution for that too.

🙌  and if a step does fail it auto-expands the relevant section for us in the logs too.

PR #93, opened 2021-11-22
65-update postman collection and quickstart valid fides keys

### LINKED ISSUE #65: Update assets across Fides ecosystem to use valid keys on models
In https://github.com/ethyca/fidesops/pull/61 we will update all keys to comply with Fides key validator such that only alphanumeric, `_`, and `.` chars allowed. 

When that is merged, we will need to also update the following:

* `quickstart.py`
* the postman collections
* https://github.com/ethyca/fidesdemo

Context: https://github.com/ethyca/fidesops/pull/61/files#r748442186

eastandwestwind:
# Purpose
Updates postman collection and quickstart to use valid fides keys

# Changes
Updates dasherized keys to snake_case keys for the following assets:
- `quickstart.py`
- `Fidesops.postman-collection.json`

# Ticket
Closes #65

### DIFF
--- a/quickstart.py
+++ b/quickstart.py
@@ -16,7 +16,6 @@ from fidesops.models.connectionconfig import ConnectionType
 from fidesops.models.policy import ActionType
 
 logger = logging.getLogger(__name__)
-logging.basicConfig(level=logging.INFO)
 
 
 def get_access_token(client_id: str, client_secret: str) -> str:
@@ -551,10 +550,10 @@ if __name__ == "__main__":
     input()
 
     create_connection(
-        key="test-application-postgres-db", connection_type=ConnectionType.postgres
+        key="test_application_postgres_db", connection_type=ConnectionType.postgres
     )
     configure_postgres_connection(
-        key="test-application-postgres-db",
+        key="test_application_postgres_db",
         host=POSTGRES_SERVER,
         port=POSTGRES_PORT,
         dbname=POSTGRES_DB_NAME,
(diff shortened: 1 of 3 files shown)


quickstart.py
```diff
@@ -16,7 +16,6 @@
 from fidesops.models.policy import ActionType
 
 logger = logging.getLogger(__name__)
-logging.basicConfig(level=logging.INFO)
```
pattisdr, line 19: Can you add this back?  Most of the quickstart is the logged output, without this, it doesn't display very much.

quickstart.py
```diff
@@ -551,10 +550,10 @@ def print_results(request_id: str) -> None:
     input()
 
     create_connection(
-        key="test-application-postgres-db", connection_type=ConnectionType.postgres
+        key="test_application_postgres_db", connection_type=ConnectionType.postgres
```
pattisdr, line 553: Could you similarly update the [README.md](https://github.com/ethyca/fidesops#rocket-quick-start) that has the instructions for this quickstart, so the keys are snake cased?

pattisdr (commented):
Thanks for your work to keep all of our docs up to date!  Just a couple of remaining things for the quickstart requested -

quickstart.py
pattisdr: This does seem necessary for the quickstart though (test with `make quickstart`) - I wonder if it's because when I do docker exec to run `python quickstart.py` on the already running fidesops container that it doesn't have the same access to the logging config that was set at the application level, that it's a separate process.
```
quickstart: compose-build
	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
	@docker exec -it fidesops python quickstart.py
```

eastandwestwind: ohh you're totally right. I'll update this, thanks!

eastandwestwind:
back over to you @pattisdr!

pattisdr (approved):
thank you @eastandwestwind!

PR #99, opened 2021-11-23
Update pull_request_template.md

### LINKED ISSUE #84: Add small PR checklist

pattisdr:
# Purpose

Add two items to PR checklist - I'm not trying to add a bunch of checkboxes, just trying to help us all be looking out for documentation that we need to get up to date!

# Ticket
Fixes #84

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,5 +10,5 @@
 
 # Ticket
 
-Resolves #
+Fixes #
  
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,6 +2,13 @@
 
 # Changes
 
-- 
+-
+# Checklist
+
+- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo)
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -2,6 +2,13 @@
 
 # Changes
 
-- 
+-
+# Checklist
+
+- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo)
+- [ ] Good unit test/integration test coverage
```
seanpreston, line 9: I like the intention here. If we were to make this less subjective, what would that look like? Something like "Overall test coverage increased" maybe?

pattisdr: yes we can definitely rephrase - I guess I was thinking that not every PR needs tests / is testable, and when I hear "increase overall test coverage" it makes me think of increasing percentage, but if we added new code, and added new tests we probably maintained the percentage.

seanpreston: Sure thing! This will serve as a good reminder regardless, so let's merge it

PR #103, opened 2021-11-24
Add app diagram to database docs [Unticketed]

pattisdr:
# Purpose
Out of our weekly Fidesops retrospective, someone suggested we should add a DB architecture diagram to the docs.

# Changes

- Used DBeaver Community to create an [entity relationship diagram ](https://dbeaver.com/docs/wiki/ER-Diagrams/) and added to the deployment guide.  Some rearranging is required because we have several resources connected to `client` and several not, so it looks messy by default.

Screen Shot 2021-12-02 at 9 32 32 AM

# Ticket
No ticket


pattisdr:
I also tried `pgAdmin 4` and this is what it came up with, all the diagonals caused by the client table, and it was harder to rearrange then DBeaver. 
Screen Shot 2021-11-24 at 1 12 35 PM

seanpreston:
This is amazing, thanks @pattisdr. Am I right in thinking the process for updating this would be:
- Connect DBeaver to an application DB container
- Screenshot the ER diagram
- Replace `img/app_database.png` with the new screenshot

pattisdr:
yes @seanpreston, although I rearranged the diagram a bit too, the `client` table being connected to multiple tables made the diagram look messy.

pattisdr:
Going to update this diagram, new tables have been added!

iamkelllly (commented):
@pattisdr great job on the diagram, as the first time i'm actually seeing an ERD,  i have a few items i just wanted to clarify for which i might open up for a separate discussion thread (nitpicks from a former dbe) but shouldn't hold up the merging of this for 1.1:

* completely open to the fact that i may be missing something as a python noob, but does alembic version _have_ to be stored in a snakecase table? it's the only one that's inconsistently named. i personally prefer that we stick to a consistent table naming convention; if we decide for all snakecase, let's make it consistent (`alembic_version`, `storage_config`, `dataset_config`, etc). alternatively, we could opt for all camelcase (`alembicVersion`, `storageConfig`, `datasetConfig`). did we happen to decide this in advance? if not, I'm happy to open a discussion thread!
* i see non-standard datatypes (`webhookdirection`, `accesslevel`), and it may be old fashioned, but i think standard ERDs have the actual datatype and char limits? should we change these to the actual db datatypes on the ERD (`varchar(255)`, `int4`) and in the documentation we refer to/mention the enumeration? 
  * if so, i think we can automatically generate that from the CREATE script?
 
otherwise, looks good and pre- and post- webhooks designed as discussed

pattisdr:
good morning @iamkelllly thanks for reviewing!

re: the alembic table, I view this as a separate table, it was [automatically created by Alembic](https://alembic.sqlalchemy.org/en/latest/tutorial.html?highlight=alembic_version), and it's a table with one row that always just stores the latest migration run. If I knew an application was using Alembic, I would expect that table to be there, with that name, and wouldn't expect the application to mimic Alembic's choice.

pattisdr:
re: enumerated types, [this is a valid type](https://www.postgresql.org/docs/9.1/datatype-enum.html) in Postgres, restricting data to be one of the values in a set. We lose information when displaying the underlying type, making the column appear less restrictive than it is. (this also this doesn't seem to be an option for the two tools I investigated). The CREATE script also uses the enum type.

Ideally I think we'd also display the enum values, something like this https://stackoverflow.com/a/9117020, but I haven't found that DBeaver or PgAdmin4 can put this on the ER diagram.

it looks like I can manually edit the diagrams to have the underlying type, but my vote would still be to show the enum.

eastandwestwind:
Just 1 more thing @pattisdr - mind if we add https://github.com/ethyca/fidesops/pull/103#issuecomment-982459734 to documentation and/or as a checklist item in PRs? Wanna make sure maintaining this is easy and apparent. Thanks!

pattisdr:
good call @eastandwestwind, I have the checklist linking to documentation on how to update.

eastandwestwind (approved):
Thanks, looks great!

PR #104, opened 2021-11-24
Snowflake Query Execution [#73]

### LINKED ISSUE #73: Query execution for Snowflake
Requires #44 to be complete first.

We need to enable the Snowflake database connector to be able to read/write to the database itself during query execution.

seanpreston:
# Purpose

This PR contains query execution for the Snowflake external datastore.

# Changes

- Overrides the `SQLQueryConfig` class to with a `SnowflakeQueryConfig`
- Adds methods in the base `SQLQueryConfig` class to format variables that eventually end up in the query string
- Implements those methods in the `SnowflakeQueryConfig`
- Adds a dataset for our test Snowflake schema with `data_type` and `length` meta annotations
- Adds an API test that runs an access request through Snowflake
- Adds an API test that runs an erasure request through Snowflake

# Left to do
- [x] Make similar query formatting changes for erasure handling in Snowflake
- [x] Add a separate test schema (Atlas uses the current one) on Snowflake for Fidesops
- [x] Configure test data programmatically for the new Snowflake test schema in the test fixtures
- [x] More thorough testing, e.g. specifically for the `Variant` type and erasure handling
- [x] Add `type` attribute into Snowflake dataset

# Ticket
Closes #73


data/dataset/snowflake_example_test_dataset.yml
```diff
@@ -0,0 +1,214 @@
...
+              references:
+                - dataset: snowflake_example_test_dataset
+                  field: address.id
+                  direction: to
+          - name: created
+            data_categories: [system.operations]
+          - name: email
+            data_categories: [user.provided.identifiable.contact.email]
+            fidesops_meta:
+              identity: email
```
eastandwestwind, line 37: heads up that after https://github.com/ethyca/fidesops/pull/108 is merged, we'll require `data_type` on applicable fields in erasure tests

seanpreston: Thanks!

tests/fixtures.py
```diff
@@ -847,20 +892,65 @@ def dataset_config_preview(
...
+        return yaml.safe_load(file).get("dataset", [])
+
+
 @pytest.fixture
 def example_datasets() -> List[Dict]:
     example_datasets = []
     example_filenames = [
         "data/dataset/postgres_example_test_dataset.yml",
         "data/dataset/mongo_example_test_dataset.yml",
+        "data/dataset/snowflake_example_test_dataset.yml",
```
seanpreston, line 907: My plan from here out is to avoid adding any subsequent datasets to this fixture, preferring to have them loaded in by a dedicated fixture for each one.

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -129,7 +131,7 @@ def test_create_and_process_access_request(
 
     policy.delete(db=db)
     pr.delete(db=db)
-    db.expunge_all()
```
seanpreston, line 132: This line was causing downstream failures in the test teardown where FKs also needed to be cleaned up (specifically the `ConnectionConfig` -> `DatasetConfig` relation). After talking to @pattisdr I discovered this line was here as a forcing function to ensure that `pr` is no longer in the database session when the execution logs are checked for their `.privacy_request_id` attribute, a step to ensure that value is not cleared when `pr` is deleted.

I've opted to manually check whether `pr` is in the session in the line below, as this performs the same function.

tests/util/test_logger.py
```diff
@@ -1,8 +1,19 @@
+import os
+import pytest
+
 from fidesops.util import logger
 from fidesops.util.logger import NotPii, MASKED
 
 
-def test_logger_masks_pii() -> None:
+@pytest.fixture(scope="function")
+def toggle_testing_envvar() -> None:
```
seanpreston, line 9: This is useful to explicitly check areas where we use the `os.environ["TESTING"]` in production code

tests/conftest.py
```diff
@@ -38,8 +41,7 @@ def migrate_test_db() -> None:
 def db() -> Generator:
     """Return a connection to the test DB"""
     # Create the test DB enginge
-    ## This asserts that TESTING==True
-    assert os.getenv("TESTING", False)
+    assert os.getenv("TESTING") == "True"
```
seanpreston, line 44: cc @stevenbenjamin — I've updated these checks to look like this now as that's the correct convention according to `pylint`. Environment vars can only be strings so returning a `bool` as the default arg is dangerous. Hopefully this syntax is clearer too.

src/fidesops/service/connectors/query_config.py
```diff
@@ -301,13 +345,53 @@ def transform_param(p: Any) -> str:
...
+        """Returns fields surrounded by quotation marks as required by Snowflake syntax."""
+        return [f'"{field}"' for field in fields]
+
+    def format_clause_for_query(
+        self,
+        field_name: str,
+        operator: str,
+    ) -> str:
+        """Returns field names in clauses surrounded by quotation marks as required by Snowflake syntax."""
+        return f'"{field_name}" {operator} (:{field_name})'
```
seanpreston, line 372: Adding the `()` around `:{field_name}` is an interesting difference to how the more vanilla SQL connectors work. It could be that currently in our Postgres and MySQL connectors `IN` functions are accepted without parenthesis.

data/dataset/snowflake_example_test_dataset.yml
```diff
@@ -0,0 +1,229 @@
+dataset:
+  - fides_key: snowflake_example_test_dataset
+    name: Snowflake Example Test Dataset
```
pattisdr, line 3: As we're thinking about test fixtures for these different datasources, it would be nice to create some interdependency between them, like how the mongo example test dataset relies on a field in the postgres example test dataset, I don't think our datasets should all be completely independent datastores.

pattisdr: It would be nice to have the postman collection be able to make requests against our fidesops snowflake db - but the way i understand it, we wipe and repopulate it from these unit tests?

pattisdr (commented):
this looks good @seanpreston, not merging yet in case there's anything lingering you want to address but it looks good. 

code was organized and readable about why certain decisions were made.  I completed local access request testing outside of our regular unit tests - this behaved as expected and it was nice to see the queries being executed too in testing mode.

data/dataset/snowflake_example_test_dataset.yml
```diff
@@ -0,0 +1,229 @@
...
+              primary_key: True
+          - name: name
+            data_categories: [user.provided.identifiable.name]
+            fidesops_meta:
+              data_type: string
+              length: 40
+          - name: variant_eg
+            # We use this data category so we can target this column from
+            # our Snowflake tests
+            data_categories: [user.provided.identifiable.name]
```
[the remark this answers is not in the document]
seanpreston: This column maps to a Snowflake specific datatype that I've added to the schema — this is where having schemas defined in our application code would really help. as it would be obvious if that were the case. I don't think we have a `data_type` annotation for `variant` yet, so I didn't add it here, but that would have also been a clue.

The variant type is more or less a wildcard column, with some other minor nuances. https://docs.snowflake.com/en/sql-reference/data-types-semistructured.html#variant

src/fidesops/models/connectionconfig.py
```diff
@@ -88,3 +92,10 @@ def update_test_status(self, test_status: TestStatus, db: Session) -> None:
         self.last_test_timestamp = datetime.now()
         self.last_test_succeeded = test_status == TestStatus.succeeded
         self.save(db)
+
+    def delete(self, db: Session) -> Optional[Base]:
```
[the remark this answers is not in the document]
seanpreston: This implementation is intentional. Since we still have the object in memory (albeit at this point detached from a DB session) it may still be useful for reference checking attributes that were loaded into memory before the DB record was deleted.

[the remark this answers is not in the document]
pattisdr: got it!

tests/fixtures.py
```diff
@@ -256,6 +262,45 @@ def snowflake_connection_config(db: Session) -> Generator:
...
+    integration_config: Dict[str, str],
+    safe_snowflake_connection_config: ConnectionConfig,
+) -> Generator:
+    uri = integration_config.get("snowflake", {}).get("external_uri") or os.environ.get(
+        "SNOWFLAKE_TEST_URI"
+    )
+    schema = SnowflakeSchema(url=uri)
+    safe_snowflake_connection_config.secrets = schema.dict()
+    safe_snowflake_connection_config.save(db=db)
+    yield safe_snowflake_connection_config
```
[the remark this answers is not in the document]
seanpreston: This is the same object from the `safe_snowflake_connection_config` fixture, and the object is deleted as part of that fixture instead.

tests/fixtures.py
```diff
@@ -256,6 +262,45 @@ def snowflake_connection_config(db: Session) -> Generator:
     connection_config.delete(db)
 
 
+@pytest.fixture(scope="function")
+def snowflake_connection_config(
```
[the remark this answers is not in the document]
seanpreston: Good spot, I'll tidy this up in some way. Maybe `unsafe_snowflake_connection_config` or just `snowflake_connection_config` and it always has secrets if available. We don't actually use the read only config anywhere.

data/dataset/snowflake_example_test_dataset.yml
seanpreston: > we wipe and repopulate it from these unit tests?

Not quite — we're only wiping and repopulating the data we use in erasures. Everything else is part of a copy of the dataset Atlas uses for Snowflake. There are some frustrating minor differences here, for example, in Snowflake the collection is `order` and in Postgres it's `orders` plural.

I'll factor this into the testing design for sure! Thanks

data/dataset/snowflake_example_test_dataset.yml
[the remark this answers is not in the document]
pattisdr: got it, thanks!

src/fidesops/service/connectors/base_connector.py
```diff
@@ -29,6 +30,11 @@ class BaseConnector(Generic[DB_CONNECTOR_TYPE], ABC):
 
     def __init__(self, configuration: ConnectionConfig):
         self.configuration = configuration
+        # If Fidesops is running in test mode, it's OK to show
+        # parameters inside queries for debugging purposes. By
+        # default we assume that Fidesops is not running in test
+        # mode.
+        self.hide_parameters = not os.getenv("TESTING") == "True"
```
stevenbenjamin, line 37: I feel like if we're going to use this everywhere (and I think it's going to get used a lot, we've already discussed other uses like the hot-reloading flag, maybe log masking as well) we should have a place where we set a global application state and not refer to this all over the codebase with `os.getenv("TESTING")`

src/fidesops/service/connectors/query_config.py
stevenbenjamin: I don't think this is a snowflake thing, but an idiosyncracy of the the way the dialect is realized in sqlalchemy. the other dialects we've used seem to add the '()' around tuple IN values automatically.

tests/fixtures.py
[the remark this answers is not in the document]
pattisdr: @seanpreston I added this PR https://github.com/ethyca/fidesops/pull/49 to not do erasures if a node's connection config doesn't have write access -

pattisdr (approved):
looks ready to me @seanpreston

Issue #98, opened 2021-11-23
Add the HTTPS Connector class 


pattisdr:
Second up for HTTP Connectors

Add an `HTTPSConnector` class with methods to talk to a second-party API.  Add additional logic to update an identity graph, pause execution, etc.

PR #105, opened 2021-11-25
Configure PolicyPreWebhooks and PolicyPostWebhooks [#97]

### LINKED ISSUE #97: [HTTPS Connectors] Add a PolicyWebhook table and endpoints to configure webhooks
First up for HTTP Connectors 

Add a PolicyWebhook table, and API endpoints to configure setting up webhooks to run before/after a Privacy Request.  We need to be able to list webhooks, and create/update them.

Relates to #41

pattisdr:
# Purpose

Add the ability to configure a list webhooks to run _before_ a PrivacyRequest executes (`PolicyPreWebhook`) and _after_ a PrivacyRequest executes (`PolicyPostWebhook`).

This just allows you define the configuration - the ability to actually trigger these callbacks will be added in followup #98.

# Changes
- Added two tables to store webhooks: `policyprewebhook` and `policypostwebhook` - this was originally going to be a single table, but because keys need to be unique for an entire table, this could cause unexpected behavior, and I think we should think of them as two separate objects.
- Added (2) bulk endpoints to set the list of `PolicyPreWebhooks` and `PolicyPostWebhooks`.  Webhooks should be sent in the desired order. Any webhooks not in the request body will be removed.
  - PUT `/policy/{policy_key}/webhook/pre_execution`
  - PUT `/policy/{policy_key}/webhook/post_execution`
- Added (2) PATCH endpoints to update the attributes on a single webhook.  Because an update to `order` can affect the order of other webhooks, any order changes are included in the response where applicable.
  - PATCH `/policy/{policy_key}/webhook/pre_execution/{pre_webhook_key}`
  - PATCH `/policy/{policy_key}/webhook/post_execution/{post_webhook_key}`
- Added (2) GET List endpoints to view all Pre-Execution or Post-Execution webhooks for a Policy
    - GET `/policy/{policy_key}/webhook/pre_execution`
    - GET `/policy/{policy_key}/webhook/post_execution`
- Added (2) GET Detail endpoints to view a single Pre- or Post-Execution Webhook for a Policy
  - GET `/policy/{policy_key}/webhook/pre_execution/{pre_webhook_key}`
  - GET `/policy/{policy_key}/webhook/post_execution/{post_webhook_key}`
- Added (2) DELETE Detail endpoints to delete a single Pre- or Post-Execution Webhook for a Policy. Note that deleting one webhook can cause other webhooks to have their orders rearranged.  If order is changed, a summary is returned in the response.
  - DELETE `/policy/{policy_key}/webhook/pre_execution/{pre_webhook_key}`
  - DELETE `/policy/{policy_key}/webhook/post_execution/{post_webhook_key}`
- Allow setting `secrets` on an `https` ConnectionConfig, by defining the start of an `HTTPSConnector`.  Unlike other ConnectionConfigs, we don't attempt to test the connection.
-  Add documentation and postman collections
- Split out the logic that commits resources to the db on `create` and `update` so we can do this at the end of a lot of adjustments to multiple webhook resources
- Dry up code that fetches Policy and ConnectionConfig or throws a 404

# Ticket
Resolves #97


src/fidesops/api/v1/endpoints/policy_endpoints.py
```diff
@@ -62,6 +63,19 @@ def get_policy_list(
     return paginate(policies, params=params)
 
 
+def get_policy_or_error(db: Session, policy_key: FidesOpsKey) -> Policy:
+    """Helper method to load Policy or throw a 404"""
+    logger.info(f"Finding policy with key '{policy_key}'")
+    policy = Policy.get_by(db=db, field="key", value=policy_key)
+    if not policy:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
```
stevenbenjamin, line 72: There's a specific `NotFoundException` in `common_exceptions.py.

stevenbenjamin (commented):
It seems like the only reason to have separate pre and post webhook tables is to disambiguate keys, but I don't think we need to require unique string key names for webhooks - we do have some other types that don't have these keys. 

I think the requirement of having some unique name key on a webhook is unecessary, and without this we could store them all in one table. This will be more important to avoid complication as we add webhooks in other places.

src/fidesops/api/v1/endpoints/policy_endpoints.py
pattisdr: I see - no one's using this though, and I'm not sure I see the benefit of creating a `NotFoundException` - it seems like it obscures the fact that it's an HTTPException.

pattisdr:
thanks for your comments @stevenbenjamin!

I see your point, but in my opinion, it is a more intuitive user experience to treat these resources as entirely separate collections of objects, and manage them independently in the API and the database.   The main driver for this was that webhooks have an `order` attribute.  An update to a Pre-Execution Webhook's order can potentially adjust the order of _other_ Pre-Execution Webhooks but _not_ Post-Execution webhooks.  The original plan was to have a single endpoint to update both pre and post, but it felt confusing to be updating two distinct groups of resources within one request.  I then moved to having multiple endpoints but still storing the webhooks behind the scenes in a single table.  However, there was still a dependency between the two groups, in that you couldn't create a Pre-Execution webhook that had the same key as a Post-Execution webhook.

So to your point of getting rid of the key constraint entirely,  I still think we _should_ have keys for webhooks, as webhooks are resources that will be reused again and again, where folks might want a human-readable way to refer to them: for example, I could create an  "app_logout_webhook" that logs users out of my mobile app after I've removed all of their other data from my system.   This seems to differ from a PrivacyRequest which is created and used once - it would be cumbersome to require creating a key there each time.

So if Webhooks should have keys, then I'd want them to be consistent with other similar resources in that they are *unique* keys, because we've created an expectation elsewhere of "this is how a key works".

Finally, addressing the complication of adding webhooks in other places, I really only see one other type of webhook, and that's ones that are integrated into the query traversal, webhooks that are capable of acting like a datasource.  With current architecture that would be achieved by attaching a DatasetConfig to an `https` ConnectionConfig, without the need for another table.  If architecture changes, adding a third table for the third type of hook doesn't seem to be a big lift.

src/fidesops/api/v1/endpoints/policy_endpoints.py
stevenbenjamin: It _does_ inherit from HTTPException, and references the status code. THe other HTTPexceptions (e.g. AuthorizationError, AuthenticationFailure) here seem to be used, I think this is just an inconsistency in our useage - see, e.g. https://github.com/ethyca/fidesops/blob/main/src/fidesops/api/v1/endpoints/oauth_endpoints.py , where we're using the custom `AuthenticationFailure` in the first method, and then a bare `HTTPException` @ line 96. 

). My preference would be to use this everywhere we're throwing a 404, but at the very least I don't think we should delete it.

stevenbenjamin:
Re: webhooks structure, I do see your point. I'm not sure that in the long run a human-readable key ("is it "app_logout_web_hook or app-logout-webhook"?) is _that_ much simpler than "001-ADF4-33221....", and it does mean youi end up with weird uniqueness violations where you might not expect them, but it definitely helps with references when you're returning a re-ordered set, so I can see how that pushes the requirement.

src/fidesops/api/v1/endpoints/policy_endpoints.py
pattisdr: yes! i understand it inherits from HTTPException, but to me it's more useful to see the HTTPException for yourself, rather than look up the custom `NotFoundException` and verify that it will throw an HTTPException.  It's that the `NotFoundException` doesn't add anything new.  

Agreed on inconsistency - in my opinion there's no need to throw the custom `AuthenticationFailure`s either.

PR #106, opened 2021-11-29
Async privacy request

stevenbenjamin:
# Purpose
Process privacy requests asynchronously
# Changes
- Submissions to privacy requests push request processing onto a separate thread.
- Separate thread requires a separate db connection
- Reworking tests that require privacy responses

# Ticket
Closes #43 

# Additional Comments:

This issue was originally intended to move execution to asynchronous dask workers. Because of the number of changes involved, this ticket now only moves privacy request execution to a fire-and-forget model. Asynchronous dask execution should be the logical next step to this work.

Async:
- PrivacyRequest is now run using a coroutine (python async) and fire-and-forget semantics. This requires some changes to the way db sessions are handled, since their existing implementation is not set to autocommit.
This is because
a) we need to communicate with the db in the co-routine using a separate db session
b) you cannot work with a sqlalchemy object in a session that's different than the one that it is part of
c) trying to access an object that's in an uncommited state from the db results in objects that are out of sync.

There's a separate PR open to have our sessions set to autocommit=True

- To minimize coupling the PrivacyRequest runner `run` method _only_ takes the privacy request id and reconstructs everything it needs directly from a dedicated db connection.

Tests:
- tests on fire-and-forget tasks should use a returned future rather than polling the db for a completion condition. This means that :
- Tests on privacy request responses that require an actual running of the request have been moved from the dntpoint test to a test that uses the future directly and waits for its completion.
- There was some discussion of removing `finished_processing_at` from the `PrivacyRequestResponse` object in `BulkPostPrivacyRequest` but left it in, since it's defined as an optional and it's still useful for completed responses.
- There are additionaly a number of changes in our tests to remove their hardcoded dependency on our docker setup, so they can be run against local resources. These are all driven by the .toml files, which can be customized.

PR #107, opened 2021-11-30
Set AUTOCOMMIT on the DB engine [unticketed]

seanpreston:
# Purpose

This PR is an exploration into setting `AUTOCOMMIT` to our DB sessions, to help make execution of the identity graph async. As discussed with @stevenbenjamin this morning, the alternative is to set `autocommit=True` at the session level in SQLAlchemy, and remove the `.commit()` references in our code. This could be more prudent as it's using SQLAlchemy's supported bindings, instead of going straight to DBAPI.

# Changes

-
# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo)

# Ticket

Fixes #

### DIFF
--- a/src/fidesops/db/session.py
+++ b/src/fidesops/db/session.py
@@ -14,7 +14,10 @@ from fidesops.core.config import config
 logger = logging.getLogger(__name__)
 
 
-def get_db_engine(database_uri: Optional[str] = None) -> Engine:
+def get_db_engine(
+    database_uri: Optional[str] = None,
+    isolation_level: Optional[str] = "AUTOCOMMIT",
+) -> Engine:
     """
     Return a database engine. If the TESTING environment var is set the
     database engine returned will be connected to the test DB.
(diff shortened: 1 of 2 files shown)


src/fidesops/db/session.py
```diff
@@ -27,7 +30,11 @@ def get_db_engine(database_uri: Optional[str] = None) -> Engine:
...
 
-    return create_engine(database_uri, pool_pre_ping=True)
+    return create_engine(
+        database_uri,
+        pool_pre_ping=True,
+        isolation_level=isolation_level,
+    )
 
 
 def get_db_session(engine: Optional[Engine] = None) -> sessionmaker:
```
seanpreston, line 40: Just setting `autocommit=True` here causes failures because we're calling `.commit()` in a lot of areas in the code, which fails if there isn't a transaction left open in the session.

seanpreston:
This is a lower priority than the async work in https://github.com/ethyca/fidesops/pull/106 so we'll close this PR before assessing once https://github.com/ethyca/fidesops/pull/106 is merged.

PR #108, opened 2021-11-30
70 feat adds policy driven masking

### LINKED ISSUE #70: Policy-driven masking - handle new data_type and length params
The ability to mask data during the query execution needs to be driven by the specified policy. This ticket will add support for the following strategies, specifically implementing the framework to handle new `data_type` and `length` params (#100 will add support for remaining strategies):

- [   ] HMAC
- [   ] Hash
- [   ] AES Encryption
- [ x ] Null
- [ x ] Random string rewrite
- [ x ] Default string rewrite

With configuration support for format-preservation for all above strategies 

This will ensure that, during a deletion privacy request execution, we will be able to:
* mask values in the connected database by data category 
* mask data per the policy's specified masking type 
(edited)

eastandwestwind:
# Purpose
- Adds handling for new `length` and `data_type` params on `Field`
- Restricts supported masking strategies to the following:
[ ] HMAC
[ ] Hash
[ ] AES Encryption
[ x ] Null
[ x ] Random string rewrite
[ x ] Default string rewrite

# Changes
- Adds supported `data_type`s to each masking strategy
- Requires a `data_type` on `Field` for masking strategies that are not `null_rewrite`
- Adds methods to each `DataTypeConverter` to truncate val to specified `length` if provided (we only truncate `integer` and `string` data types) 

# Checklist

- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo)

# Ticket

Fixes #70

### DIFF
--- a/docs/fidesops/docs/tutorial/annotate_datasets.md
+++ b/docs/fidesops/docs/tutorial/annotate_datasets.md
@@ -55,7 +55,7 @@ Optional `fidesops_meta` fields:
   - `float`
   - `boolean`
   - `object_id`
-- `length`: if `length` is specified, in an erasure request, Fidesops will truncate the masked value to the specified `length`. This should be used if your database column has a length restriction. E.g. if a specific row has `email` of `jerry@mail.com`, and the masked value is `23982r3n8rupq8ewurnw`, and `length` is set to `10`, then the resulting update query will use the value `23982r3n8r`.
+- `length`: if `length` is specified, in an erasure request, Fidesops will truncate the end of the masked value to the specified `length`. This should be used if your database column has a length restriction. E.g. if a specific row has `email` of `jerry@mail.com`, and the masked value is `23982r3n8rupq8ewurnw`, and `length` is set to `10`, then the resulting update query will use the value `23982r3n8r`.
 
 ## Upload this Dataset to Fidesops
 
--- a/docs/fidesops/docs/tutorial/annotate_datasets.md
+++ b/docs/fidesops/docs/tutorial/annotate_datasets.md
@@ -48,6 +48,15 @@ the user by `email`, and from there, travel through other tables linked to `user
     identity: email
 ```
 
+Optional `fidesops_meta` fields:
+- `data_type`: only required for processing erasure requests, however, whenever specified, Fidesops will attempt to coerce the field's value to the type. For erasure requests in particular, Fidesops requires `data_type` to ensure that the configured masking strategy can process the value. Supported data types:
+  - `string`
+  - `integer`
+  - `float`
+  - `boolean`
+  - `object_id`
+- `length`: if `length` is specified, in an erasure request, Fidesops will truncate the masked value to the specified `length`. This should be used if your database column has a length restriction. E.g. if a specific row has `email` of `jerry@mail.com`, and the masked value is `23982r3n8rupq8ewurnw`, and `length` is set to `10`, then the resulting update query will use the value `23982r3n8r`.
+
 ## Upload this Dataset to Fidesops
 
 For more detailed information, [see the Datasets Guide](../guides/datasets.md).
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -222,6 +222,8 @@ class Field(BaseModel):
 
 @dataclass
 class MaskingOverride:
+    """Data class to store override params related to data masking"""
+
     data_type: Optional[DataType]
     length: Optional[int]
 
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -78,6 +78,7 @@ Field identities:
 from __future__ import annotations
 
 from collections import defaultdict
+from dataclasses import dataclass
 from typing import List, Optional, Tuple, Set, Dict, Literal, Any
 
 from pydantic import BaseModel
@@ -217,6 +218,12 @@ class Field(BaseModel):
                 return converter.to_value(value)
 
         return value
+
+
+@dataclass
+class MaskingOverride:
+    data_type: Optional[DataType]
+    length: Optional[int]
 
 
 class Collection(BaseModel):
--- a/src/fidesops/graph/data_type.py
+++ b/src/fidesops/graph/data_type.py
@@ -26,7 +26,7 @@ class DataTypeConverter(ABC, Generic[T]):
     def truncate(self, length: int, val: T) -> T:
         """Truncates value to given length"""
         print(
-            f"Length truncation is only supported for string and int data types. Using original masked value instead for update query."
+            "Length truncation is only supported for string data types. Using original masked value instead for update query."
         )
         return val
 
@@ -61,10 +61,6 @@ class IntTypeConverter(DataTypeConverter[int]):
     def empty_value(self) -> int:
         """Empty int value"""
         return 0
-
-    def truncate(self, length: int, val: int) -> int:
-        """Truncates value to given length"""
-        return int(str(val)[:length])
 
 
 class FloatTypeConverter(DataTypeConverter[float]):
--- a/src/fidesops/graph/data_type.py
+++ b/src/fidesops/graph/data_type.py
@@ -25,8 +25,8 @@ class DataTypeConverter(ABC, Generic[T]):
 
     def truncate(self, length: int, val: T) -> T:
         """Truncates value to given length"""
-        logger.warning(
-            f"Length truncation not supported for {T} data_type. Using original masked value instead for update query."
+        print(
+            f"Length truncation is only supported for string and int data types. Using original masked value instead for update query."
         )
         return val
 
--- a/src/fidesops/graph/data_type.py
+++ b/src/fidesops/graph/data_type.py
@@ -23,9 +23,12 @@ class DataTypeConverter(ABC, Generic[T]):
     def empty_value(self) -> T:
         """A value that represents `empty` in whatever way makes sense for type T"""
 
-    @abstractmethod
     def truncate(self, length: int, val: T) -> T:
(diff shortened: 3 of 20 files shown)


docs/fidesops/docs/tutorial/annotate_datasets.md
```diff
@@ -48,6 +48,15 @@ the user by `email`, and from there, travel through other tables linked to `user
...
 ```
 
+Optional `fidesops_meta` fields:
+- `data_type`: only required for processing erasure requests, however, whenever specified, Fidesops will attempt to coerce the field's value to the type. For erasure requests in particular, Fidesops requires `data_type` to ensure that the configured masking strategy can process the value. Supported data types:
+  - `string`
+  - `integer`
+  - `float`
+  - `boolean`
+  - `object_id`
+- `length`: if `length` is specified, in an erasure request, Fidesops will truncate the masked value to the specified `length`. This should be used if your database column has a length restriction. E.g. if a specific row has `email` of `jerry@mail.com`, and the masked value is `23982r3n8rupq8ewurnw`, and `length` is set to `10`, then the resulting update query will use the value `23982r3n8r`.
```
stevenbenjamin, line 58: I think, a bit clearer, `Fidesops will truncate the  **end of the**  masked value to the specified `length`,

src/fidesops/graph/config.py
```diff
@@ -219,6 +220,14 @@ def cast(self, value: Any) -> Optional[Any]:
         return value
 
 
+@dataclass
+class MaskingOverride:
```
stevenbenjamin, line 224: As far as I can tell, this is only being used in iteration in the query config as a temporary storage of (datatype, length). Maybe this class is not needed?

src/fidesops/graph/data_type.py
```diff
@@ -105,6 +133,13 @@ def empty_value(self) -> ObjectId:
         """Empty objectId value"""
         return ObjectId("000000000000000000000000")
 
+    def truncate(self, length: int, val: ObjectId) -> ObjectId:
+        """Truncates value to given length"""
+        logger.warning(
+            "Length truncation not supported for ObjectId data_type. Using original masked value instead for update query."
+        )
```
stevenbenjamin, line 140: Since all of these truncate methods, with the exception of `string`, are identical, they should probably be implemented in the base class and just get overridden for string.

src/fidesops/service/connectors/query_config.py
```diff
@@ -121,17 +131,56 @@ def update_value_map(self, row: Row, policy: Policy) -> Dict[str, Any]:
...
+                            f"Unable to generate a query for field {field_name} due to: data_type required on fields for the {strategy_config['strategy']} masking strategy"
+                        )
+                        continue
+                    if not strategy.data_type_supported(
+                        data_type=masking_override.data_type.name
+                    ):
+                        logger.warning(
+                            f"Unable to generate a query for field {field_name} due to: data_type of {masking_override.data_type} is not supported for the {strategy_config['strategy']} masking strategy"
+                        )
+                        continue
```
stevenbenjamin, line 166: There are 4 separate "if not" conditions in this method, makes it a bit hard to follow. Is there a way to simplify this?

src/fidesops/service/connectors/query_config.py
```diff
@@ -121,17 +131,56 @@ def update_value_map(self, row: Row, policy: Policy) -> Dict[str, Any]:
...
         """
-        rule_to_collection_fields = self.build_rule_target_fields(policy)
+        rule_to_collection_fields: Dict[
+            Rule, List[str]
+        ] = self.build_rule_target_fields(policy)
 
         value_map: Dict[str, Any] = {}
         for rule, field_names in rule_to_collection_fields.items():
             strategy_config = rule.masking_strategy
-            strategy = get_strategy(
```
stevenbenjamin, line 129: What about on #139: `for rule, field_names in rule_to_collection_fields.items() if rule.masking_strategy:` to get rid of the extra conditional on #141?

src/fidesops/service/connectors/query_config.py
```diff
@@ -121,17 +131,56 @@ def update_value_map(self, row: Row, policy: Policy) -> Dict[str, Any]:
...
+                        data_type=masking_override.data_type.name
+                    ):
+                        logger.warning(
+                            f"Unable to generate a query for field {field_name} due to: data_type of {masking_override.data_type} is not supported for the {strategy_config['strategy']} masking strategy"
+                        )
+                        continue
+                val: Any = row[field_name]
+                masked_val = strategy.mask(val)
+                logger.info(
+                    f"Generated the following masked val for field {field_name}: {masked_val}"
```
stevenbenjamin, line 170: I think this should be debug: info log on each mask will generate _a lot_ of log messages.

src/fidesops/service/masking/strategy/masking_strategy_hash.py
```diff
@@ -15,6 +15,7 @@
 
 
 HASH = "hash"
+SUPPORTED_DATA_TYPES = ["string"]
```
stevenbenjamin, line 18: If we're going to have `Supported Types`, it should probably be an abstract method or maybe a default implementation  on the parent MaskingStrategy class. I don't think this should be a module level variable.

Also, since we're repeatedly calling "x in SUPPORTED_DATA_TYPES" , making SUPPORTED_DATA_TYPES a set (since we're only creating it once) will be faster.

src/fidesops/graph/config.py
eastandwestwind: With a class, we have the benefit of type hints (vs using a `namedtuple`), see https://stackoverflow.com/a/50038614/4957420

src/fidesops/graph/data_type.py
eastandwestwind: good point!

src/fidesops/service/connectors/query_config.py
eastandwestwind: for sure. Refactoring out some separate static methods to make this easier to read

src/fidesops/service/connectors/query_config.py
eastandwestwind: perfect thanks!

src/fidesops/service/masking/strategy/masking_strategy_hash.py
eastandwestwind: I'm using the abstract method `data_type_supported` already, it's just the `SUPPORTED_DATA_TYPES` const that's specific to each module. I think a `data_type_supported` abstract method works better than a `supported_data_types` abstract method bc it leaves the specific details up to the module. i.e. for null masking, we always support the data type, so the caller of a `supported_data_types` abstract method would need to have a special case for null masking

eastandwestwind: Makes sense, I'll update!

src/fidesops/graph/data_type.py
```diff
@@ -23,9 +23,12 @@ def to_value(self, other: Any) -> Optional[T]:
     def empty_value(self) -> T:
         """A value that represents `empty` in whatever way makes sense for type T"""
 
-    @abstractmethod
     def truncate(self, length: int, val: T) -> T:
         """Truncates value to given length"""
+        logger.warning(
+            f"Length truncation not supported for {T} data_type. Using original masked value instead for update query."
+        )
+        return val
```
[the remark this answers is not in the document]
eastandwestwind: Ah right, thinking how I'd obtain type in the abstract base class here...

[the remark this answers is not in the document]
eastandwestwind: looks like it'd be a bit convoluted https://stackoverflow.com/questions/57706180/generict-base-class-how-to-get-type-of-t-from-within-instance so I'm going with diff wording for now in this message

src/fidesops/graph/data_type.py
```diff
@@ -49,6 +62,10 @@ def empty_value(self) -> int:
         """Empty int value"""
         return 0
 
+    def truncate(self, length: int, val: int) -> int:
+        """Truncates value to given length"""
+        return int(str(val)[:length])
```
stevenbenjamin, line 67: Probably not the right way to truncate an int. 
The reason for our truncation is so that we don't generate values that will fail on insert when we mask data because they won't fit in the target database. So, e.g. here are mysql types: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
These are truncated by bytes, so, e.g. if you had a mysql column defined as an INT and you tried to insert a value greater than 2147483647 it would fail.

I think this can be deferred ( for the moment at least - it was never supported in atlas AFAIK) but we shouldn't be truncating alphabetically, in any case.

eastandwestwind: got it. So would byte truncation be diff for nosql vs mysql?

src/fidesops/graph/data_type.py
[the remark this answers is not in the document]
stevenbenjamin: FYI Dawn suggested instead of typename the much simpler
```
def truncate(self, length: int, val: T) -> T:
  typename = self.__class__.__name__ 
 ```
 
Which will give you e.g. "IntTypeConverter"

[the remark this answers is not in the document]
eastandwestwind: ooo awesome idea thanks!

src/fidesops/graph/data_type.py
stevenbenjamin: I don't really know. The thing I posted was mysql specific.  All of the nosql databases I've worked with sidestep this issue by basically allowing you to write anything - e.g. dynamo _only_ uses BigDecimals in the java implementation . 
I think it shoiuld only matter with defined schema types.

Issue #101, opened 2021-11-24
[HTTPS Connectors] Run Pre-Execution and Post-Execution Policy Webhooks


pattisdr:
We are working towards allowing users to configure a list of webhooks on the Policy that would run serially before a PrivacyRequest runs (pre-execution) and then run after the PrivacyRequest runs (post-execution).

- The task here is to actually run the list of configured webhooks before and after a privacy request executes
- Also allow resuming a privacy request from a specific pre-execution webhook.
- Add follow-up cleanup of paused privacy requests to mark them as errored - if the cache has expired, even if the webhook responds some time later, we are not going to be able to finish the request.  See https://github.com/ethyca/fidesops/blob/main/src/fidesops/tasks/scheduled/tasks.py#L35-L52

Depends on https://github.com/ethyca/fidesops/issues/97 and https://github.com/ethyca/fidesops/issues/98

PR #114, opened 2021-12-07
Expand HTTPSConnector logic to hit Webhooks [#98]

### LINKED ISSUE #98: Add the HTTPS Connector class

pattisdr:
❗ Migration added. Before merge, ensure downrev does not need to be bumped.

# Purpose

Adds logic to be able to call client-defined webhooks using an HTTPS Connector.  Add an endpoint to resume executing a privacy request with a specific token if it has been paused.

Note: **Privacy request execution does not yet trigger these webhooks**, this is just the foundation.  Calling webhooks will be added here:  https://github.com/ethyca/fidesops/issues/101

# Changes
- Adds API endpoint to resume PrivacyRequest execution - verifies a specific token. 
- Adds a generic `HTTPConnector.execute` method that makes a request to a url configured on the ConnectionConfig and returns the response if one is expected. Throws a `ClientUnsuccessfulException` if the response is not a 2XX status.
- Adds a more specific `PrivacyRequest.trigger_policy_webhook(webhook)` that hits a single webhook as part of privacy request execution.  
  - Builds the webhook request 
  - Adds `reply-to-token` and `reply-to` headers for Pre-Execution requests since these can pause execution. The user should send a request to the `reply-to` URL with the `reply-to-token` when they're ready to resume.
  - Pauses privacy request execution if webhook instructs to `halt`
  - Caches any new identities received.
- Adds `PrivacyRequestRunner.run_webhooks` to run all the Pre- or Post-Execution webhooks and handle updating Privacy Request statuses if the webhook causes execution to exit.
- Adds a migration to add a `paused` PrivacyRequestStatus
- Removes obsolete HTTPS Connector code from earlier foundational work

# Checklist

- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](../docs/fidesops/docs/development/update_erd_diagram.md)

# Ticket

Closes #98


src/fidesops/models/privacy_request.py
```diff
@@ -33,9 +55,20 @@ class PrivacyRequestStatus(EnumType):
...
+    paused = "paused"
     error = "error"
 
 
+def generate_request_callback_jwe(webhook: PolicyPreWebhook) -> str:
+    """Generate a JWE to be used to resume privacy request execution."""
+    jwe = WebhookJWE(
+        webhook_id=webhook.id,
+        scopes=[PRIVACY_REQUEST_CALLBACK_RESUME],
+        iat=datetime.now().isoformat(),
```
pattisdr, line 67: Sean and I chatted about this - we're including the scopes in the token we give to the user to follow the same format as our other oauth endpoints.  However, note that we gave these scopes in the token to the user, and they are just handing it back to us.

src/fidesops/models/privacy_request.py
```diff
@@ -118,15 +161,55 @@ def get_results(self) -> Dict[str, Any]:
...
+            callback_type=webhook.prefix,
+            identities=self.get_cached_identity_data(),
+        )
+
+        headers = {}
+        is_pre_webhook = webhook.__class__ == PolicyPreWebhook
+        response_expected = webhook.direction == WebhookDirection.two_way
+        if is_pre_webhook and response_expected:
+            headers = {
+                "reply-to": f"/privacy-request/{self.id}/resume",
```
pattisdr, line 184: hmm, what's a good way to get the host here to include in the reply-to link

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -368,3 +384,24 @@ def get_request_preview_queries(
...
+    webhook: PolicyPreWebhook = Security(
+        verify_callback_oauth, scopes=[scopes.PRIVACY_REQUEST_CALLBACK_RESUME]
+    ),
+    webhook_callback: PrivacyRequestResumeFormat,
+) -> None:
+    """Resume running a privacy request after it was paused by a Pre-Execution webhook"""
+    privacy_request = get_privacy_request_or_error(db, privacy_request_id)
+    privacy_request.cache_identity(webhook_callback.derived_identities)
+
+    # TODO resume running privacy request from specific webhook
```
pattisdr, line 407: This will be connected in follow-up #101, currently no webhooks are actually run.

src/fidesops/models/privacy_request.py
```diff
@@ -118,15 +161,55 @@ def get_results(self) -> Dict[str, Any]:
...
+        response_body = SecondPartyResponseFormat(**response)
+
+        # Cache any new identities
+        if response_body.derived_identities and any(
+            [response_body.derived_identities.dict().values()]
+        ):
+            logger.info(
+                f"Updating known identities on privacy request {self.id} from webhook {webhook.key}."
+            )
+            self.cache_identity(response_body.derived_identities)
```
stevenbenjamin, line 206: This should create a union of the previously defined identity map and the newly created values - is that correct behavior? Or should we clean out the old values first?

src/fidesops/schemas/redis_cache.py
```diff
@@ -1,10 +1,16 @@
...
 
-    phone_number: Optional[str]
-    email: Optional[str]
+    phone_number: Optional[str] = None
+    email: Optional[str] = None
+
+    class Config:
+        """Only allow phone_number and email to be supplied"""
+
+        extra = Extra.forbid
```
stevenbenjamin, line 16: In theory we should be able to match any labelled data in a privacy request, as long as it matches the key looked for in the data configuration. Is it a requirement that we're limiting it?

src/fidesops/models/privacy_request.py
```diff
@@ -118,15 +161,55 @@ def get_results(self) -> Dict[str, Any]:
...
+
+        # Pause execution if instructed
+        if response_body.halt and is_pre_webhook:
+            raise PrivacyRequestPaused("Halt instruction received.")
```
stevenbenjamin, line 210: Can this log message contain the privacy request id?

src/fidesops/models/privacy_request.py
pattisdr: I was under the impression it should be a union, I'm only replacing an identity if they send in a non-null value, but I guess this is a product question @iamkelllly

src/fidesops/schemas/redis_cache.py
pattisdr: I think it's a requirement, I believe @seanpreston told me there had been a decision to just start with supporting `phone_number` and `email` identities - correct me if I'm wrong Sean.

It's worth pointing out that this doesn't change behavior.  The `PrivacyRequestIdentity` schema _already_ restricts that only phone_number and email identities are cached, this change just keeps the user from _attempting to submit_ other fields.  Otherwise, the user can submit a street_address, and believe we are adding it to the identity graph, but we are not.

src/fidesops/models/privacy_request.py
pattisdr: Well this isn't a log, it's just the exception i'm raising, and then i catch that exception in PrivacyRequestRunner.run_webhooks, and log there:, but I can add the id to the exception message.

```python
    for webhook in webhooks.order_by(webhook_cls.order):
            try:
                self.privacy_request.trigger_policy_webhook(webhook)
            except PrivacyRequestPaused:
                logging.info(
                    f"Pausing execution of privacy request {self.privacy_request.id}. Halt instruction received from webhook {webhook.key}."
                )
...
```

pattisdr:
thanks for the review @stevenbenjamin 👍

src/fidesops/schemas/redis_cache.py
stevenbenjamin: I don't think that makes sense (I'd missed that previously in the privacy request schema) - what does it add to disallow matching on e.g. social security number or any other field(s) someone might want to drive this with? That being said, agree that it's a pre-existing product decision.

pattisdr: I agree we're going to have to add support for other fields, but I guess we'll intentionally add them over time?

pattisdr:
new webhook scopes added, good catch @stevenbenjamin

PR #119, opened 2021-12-08
100 - Support Hash, HMAC, AES Masking strategies

### LINKED ISSUE #100: Add support for hash, HMAC, and AES masking strategies
The ability to mask data during the query execution needs to be driven by the specified policy. This ticket will add support for the following strategies, specifically implementing secrets management necessary for these strategies (#70 adds support for remaining strategies):

- [ x ] HMAC
- [ x ] Hash
- [ x ] AES Encryption
- [   ] Null
- [   ] Random string rewrite
- [   ] Default string rewrite

With configuration support for format-preservation for all above strategies 

This will ensure that, during a deletion privacy request execution, we will be able to:
* mask values in the connected database by data category 
* mask data per the policy's specified masking type 
(edited) 

From @NevilleS -

For all the strategies that use some random algorithms, we need to generate some values during execution and reuse them across fields, e.g.

- hash: requires a salt (aka pepper!) value
- AES encryption: requires an encryption key & nonce
- HMAC: requires an encryption key

These secrets should be randomly generated once per privacy request, used for each invocation, then discarded at the end of execution.

eastandwestwind:
# Purpose
Adds support for Hash, HMAC, AES Masking strategies

# Changes
- [Caches masking secrets](https://github.com/ethyca/fidesops/pull/119/files#diff-76af5439c4ac6f601cee84dc1ee9c4965a22524d0687c2a1c06872af004cc290R164) at the start of new privacy request
- Adds new methods on base [MaskingStrategy](https://github.com/ethyca/fidesops/pull/119/files#diff-e7cee4f5ed1aee5fa59c68ee700c573ee3be5ca565bc837e66eedc3ed74be35fR20) so that the above ^ can store secrets in the cache if necessary
- Pipes in privacy request to [query_config.py](https://github.com/ethyca/fidesops/pull/119/files#diff-fde713ede3a710d10f6a74c7a8c399bd171a278e95b74086b37d37c20390eee8R254) so that it's accessible from within masking strategies (to get appropriate secret from cache)
- Adds shared [secrets_util.py](https://github.com/ethyca/fidesops/pull/119/files#diff-ef016f642b72583730562a7c5142f2165c0b0820e0a89950ce76df51f9613c9cR1) to manage generating and retrieving secrets for masking strategies. This can be used both from standalone masking endpoint AND from create privacy request endpoints: 
- Extracts out shared [hmac_encryption_scheme.py](https://github.com/ethyca/fidesops/pull/119/files#diff-76061bf0e82ad44c4ffc9aa0ebd844cc6fe8ed7f91d25e033c14ffcc44547c6fR1) that can be used from within the HMAC strategy AND the AES strategy (to generate nonce)

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](../docs/fidesops/docs/development/update_erd_diagram.md)

# Ticket

Fixes #100


src/fidesops/api/v1/endpoints/masking_endpoints.py
```diff
@@ -30,7 +30,8 @@ def mask_value(masking_strategy: PolicyMaskingSpec, value: str) -> MaskingAPIRes
             masking_strategy.strategy, masking_strategy.configuration
         )
         logger.info(f"Starting masking with strategy {masking_strategy.strategy}")
-        masked_value = strategy.mask(value)
+        # fixme: generate secrets, but what privacy request id to use?
```
eastandwestwind, line 33: Since we call `mask()` outside the context of a privacy request, how should we handle scoping the secrets? They are typically scoped by privacy request id. This is also complicated by the fact that currently, status quo is calling cache methods directly on the `PrivacyRequest` model.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -132,6 +137,26 @@ def create_privacy_request(
...
+            logger.info(
+                f"Caching masking secrets for privacy request {privacy_request.id}"
+            )
+            erasure_rules: List[Rule] = policy.get_rules_for_action(
+                action_type=ActionType.erasure
+            )
+            unique_masking_strategies_by_name: Set[str] = set()
+            for rule in erasure_rules:
+                unique_masking_strategies_by_name.add(rule.masking_strategy.strategy)
+            # todo- could generate all secrets for all strategies
```
eastandwestwind, line 150: Currently, this bit of code only generates the secrets needed for a given request. This means that, in our `AES` strategy, we need to go aheads and generate secrets for `HMAC` in the `AES` context, because we need `HMAC` secrets to generate the nonce. These are the HMAC-in-AES-context secrets I'm referring to: https://github.com/ethyca/fidesops/pull/119/files#diff-3c3364723154032848cad20814d09705811bb0abbe49450ca21921e8af0468ccR9-R11

Alternatively, we could blindly generate all secrets for all strategies at this point, which means we won't need to generate separate `HMAC` secrets just for `AES`.

src/fidesops/core/config.py
```diff
@@ -116,6 +116,7 @@ class SecuritySettings(FidesSettings):
 
     AES_ENCRYPTION_KEY_LENGTH: int = 16
     AES_GCM_NONCE_LENGTH: int = 12
+    DEFAULT_ENCRYPTION_BYTE_LENGTH: int = 32
```
eastandwestwind, line 119: Copied over from Python secrets module default. Question is - should we have a configurable byte length per encryption strategy and secret type? (like we do for `AES_ENCRYPTION_KEY_LENGTH`).

src/fidesops/api/v1/endpoints/masking_endpoints.py
eastandwestwind: This decision will help determine:
1. If we should require a `privacy_request_id` in `mask()`. 
2. If we should require the caller of `mask()` to generate/cache needed secrets before calling `mask()`
3. If we should have the masking strategy generate/cache secrets if privacy_request_id is None

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
```diff
@@ -1,38 +1,95 @@
...
             raise ValueError(f"aes_mode {self.mode} is not supported")
 
+    @staticmethod
+    def _generate_nonce(
+        value: Optional[str], key: str, privacy_request_id: Optional[str]
+    ) -> bytes:
+        salt: str = SecretsUtil.get_secret(
+            privacy_request_id, AES_ENCRYPT, SecretType.salt_hmac
+        )
+        # fixme: replicate what Vault does with sum and remove lower bytes
```
eastandwestwind, line 53: Need to research and decide specifics on how we want to sum / trim bytes

eastandwestwind: Looks like `sum()` just computes the checksum of the HMAC strategy, not the sum (mathematical sum) as we  previously thought: Vault uses this hmac crypto package- https://pkg.go.dev/crypto/hmac

src/fidesops/service/masking/masking_processor.py
```diff
@@ -19,5 +19,6 @@ def mask(value: str, masking_strategies: List[PolicyMaskingSpec]) -> str:
         strategy = masking_strategy_factory.get_strategy(
             strategy.strategy, strategy.configuration
         )
-        masked_value = strategy.mask(masked_value)
+        # fixme: generate secrets, but what privacy request id to use?
```
eastandwestwind, line 22: Here's another place where `mask()` is called, but without context of privacy request. This bit of code doesn't seem used at the moment, so what is it for?

src/fidesops/api/v1/endpoints/masking_endpoints.py
NevilleS: This API is meant to act as a standalone masking API, so it shouldn't rely on a privacy request. I think we have two options here:
1. Require the caller supply all secrets for the masking (e.g. keys, peppers, nonces), which would allow deterministic masking
2. Generate all secrets randomly, which would make it non-deterministic

Since we don't have a very specific use case for this API yet (I imagine it being used to enable masking-as-a-service use cases in the future) I'd vote for option #2 as it is simple and easy to document and understand. It would therefore be a way to "test out" the various strategies before enabling them in an erasure policy.

src/fidesops/core/config.py
NevilleS: TBH I suspect these config values are really just constants - I don't believe you can modify the AES key or nonce length, for example... similarly, I think the byte length you're adding here also probably *has* to be 32. If none of these values are configurable, I'd prefer we move these out of the config class and declare them as constants in the encryption source files instead.

src/fidesops/service/masking/masking_processor.py
NevilleS: Probably just some legacy code from earlier testing, before we had the request execution working? If it's unused I'd recommend deleting it and refactoring the tests.

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
NevilleS: Ah interesting, that makes sense!

src/fidesops/api/v1/endpoints/masking_endpoints.py
eastandwestwind: Regarding option 2, does this mean random within a given privacy request, i.e. the `key` for `HMAC` will be different for every privacy request?

src/fidesops/core/config.py
eastandwestwind: it's configurable, and byte length can be any specified length. If not provided, the default package val is 32. For now, I'll just declare it as a const in the encryption level file

src/fidesops/api/v1/endpoints/masking_endpoints.py
eastandwestwind: answer from Neville- yes, the keys should be generated randomly on each invocation of `mask()`, so we won't need to cache anything for now.

src/fidesops/util/encryption/secrets_util.py
```diff
@@ -0,0 +1,72 @@
...
+        if privacy_request_id is not None:
+            secret: T = SecretsUtil._get_secret_from_cache(
+                privacy_request_id, masking_secret_meta
+            )
+            if not secret:
+                logger.warning(
+                    f"Secret type {masking_secret_meta.secret_type} expected from cache but was not present for masking strategy {masking_secret_meta.masking_strategy}"
+                )
+                return masking_secret_meta.generate_secret()
+                # todo- should we attempt to re-cache in this case?
```
eastandwestwind, line 30: If for some reason secret is not in cache, we can generate it on the fly. Should we also attempt to re-cache?

src/fidesops/models/privacy_request.py
```diff
@@ -168,6 +183,9 @@ def trigger_policy_webhook(self, webhook: WebhookTypes) -> None:
         Pre-Execution webhooks send headers to the webhook in case the service needs to send back instructions
         to halt.  To resume, they use send a request to the reply-to URL with the reply-to-token.
         """
+        # temp fix for circular dependency
+        from fidesops.service.connectors import HTTPSConnector, get_connector
```
eastandwestwind, line 187: not ideal, but this will take some time to address

eastandwestwind: ideally we shouldn't be making outbound calls to services from db models

pattisdr: maybe we move this method to the privacy request runner?

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -154,6 +161,28 @@ def create_privacy_request(
...
+            logger.info(
+                f"Caching masking secrets for privacy request {privacy_request.id}"
+            )
+            erasure_rules: List[Rule] = policy.get_rules_for_action(
+                action_type=ActionType.erasure
+            )
+            unique_masking_strategies_by_name: Set[str] = set()
+            for rule in erasure_rules:
+                unique_masking_strategies_by_name.add(rule.masking_strategy["strategy"])
+            for strategy_name in unique_masking_strategies_by_name:
```
seanpreston, line 174: This second loop isn't adding any value — you could remove this line entirely and still have the same working code (and just one loop)

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -154,6 +161,28 @@ def create_privacy_request(
...
+                strategy = SupportedMaskingStrategies[strategy_name].value
```
seanpreston, line 175: This will break if `strategy_name` is unsupported, and therefore not a key in `SupportedMaskingStrategies` — although I think that may already return an error at schema level instead?

src/fidesops/schemas/masking/masking_secrets.py
```diff
@@ -0,0 +1,39 @@
...
+
+    key = "key"
+    salt = "salt"
+    # the below types are used by the AES algorithm, when it calls HMAC to generate the nonce
+    key_hmac = "key_hmac"
+    salt_hmac = "salt_hmac"
+
+
+@dataclass(unsafe_hash=True)
+class MaskingSecretMeta(Generic[T]):
```
seanpreston, line 23: Why do we need `Generic[T]` here instead of the default?

src/fidesops/schemas/masking/masking_secrets.py
```diff
@@ -0,0 +1,39 @@
...
+    """Holds metadata describing one secret"""
+
+    masking_strategy: str
+    secret_type: SecretType
+    generate_secret: Callable[[int], T]
```
seanpreston, line 28: Would be good for the naming here to indicate this is a function too, since when you see `generate_secret=...` inside kwargs that isn't immediately obvious

src/fidesops/service/masking/strategy/masking_strategy.py
```diff
@@ -13,10 +13,20 @@ class MaskingStrategy(ABC):
     """Abstract base class for masking strategies"""
 
     @abstractmethod
-    def mask(self, value: Optional[str]) -> Optional[str]:
+    def mask(self, value: Optional[str], request_id: Optional[str]) -> Optional[str]:
         """Used to mask the provided value"""
         pass
 
+    @staticmethod
```
seanpreston, line 20: Should this be a static method?

src/fidesops/service/masking/strategy/masking_strategy.py
```diff
@@ -13,10 +13,20 @@ class MaskingStrategy(ABC):
...
+    @abstractmethod
+    def secrets_required() -> bool:
+        """Determines whether secrets are needed for specific masking strategy"""
+        pass
```
seanpreston, line 24: `pass` isn't required when a docstring is present — we should keep this consistent across the codebase

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
```diff
@@ -11,28 +17,53 @@
...
 
-    def mask(self, value: Optional[str]) -> str:
+    def mask(self, value: Optional[str], privacy_request_id: Optional[str]) -> str:
         if self.mode == AesEncryptionMaskingConfiguration.Mode.GCM:
-            masked: str = encrypt(value, self.key, self.nonce)
+            masking_meta: Set[MaskingSecretMeta] = self._build_masking_secret_meta()
+            key: bytes = SecretsUtil.get_or_generate_secret(
+                privacy_request_id,
+                [meta for meta in masking_meta if meta.secret_type == SecretType.key][
+                    0
```
seanpreston, line 37: Are we better off returning the `masking_meta` as a dict:
```
{
    "key": ...,
    "key_hmac": ...,
    ...,
}
```
instead of as a list and having to address `[0]` here?

Relying on `[0]` is brittle since it will error if nothing is there for any reason, and if multiple are present we'll have potentially redundant configuration.

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
```diff
@@ -60,3 +93,47 @@ def data_type_supported(data_type: Optional[str]) -> bool:
...
+            [meta for meta in masking_meta if meta.secret_type == SecretType.salt_hmac][
+                0
+            ],
+        )
+        # fixme: replicate what Vault does - remove lower bytes
+        return hmac_encrypt_return_bytes(
+            value, key, salt, HmacMaskingConfiguration.Algorithm.sha_256
+        )
+
+    @staticmethod
```
seanpreston, line 115: I'm not sure these ought to be staticmethods either

tests/util/encryption/test_secrets_util.py
```diff
@@ -0,0 +1,58 @@
...
+        secret_type=SecretType.key,
+        generate_secret=SecretsUtil.generate_secret_string,
+    )
+
+    result: str = SecretsUtil.get_or_generate_secret(None, masking_meta_key)
+    assert result
+
+def test_build_masking_secrets_for_cache() -> None:
+    # build masking secret meta for all HMAC secrets
+    masking_meta: Set[MaskingSecretMeta] = set()
```
seanpreston, line 42: It would be good to also test the way this is built with `_build_masking_secret_meta`

tests/util/encryption/test_secrets_util.py
```diff
@@ -0,0 +1,58 @@
...
+request_id = "12345"
+
+
+def test_get_secret_from_cache() -> None:
+    # build masking secret meta for HMAC key
+    masking_meta_key = MaskingSecretMeta[str](
+            masking_strategy=HMAC,
+            secret_type=SecretType.key,
+            generate_secret=SecretsUtil.generate_secret_string,
+        )
```
seanpreston, line 17: The whitespace appears to be slightly off here

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
eastandwestwind: it would error at schema level, but for true zero trust, I can do a check here too.

src/fidesops/schemas/masking/masking_secrets.py
eastandwestwind: By using a `Generic[T]`, we can pass in the expected data type that `generate_secret` returns, when we're instantiating `MaskingSecretMeta`. Is this what you mean?

src/fidesops/service/masking/strategy/masking_strategy.py
eastandwestwind: ok, I'll remove the `pass` statements from the rest of this class too

src/fidesops/models/privacy_request.py
eastandwestwind: I'll add a follow-up ticket for this, since I want to keep this PR from getting too large

src/fidesops/service/masking/strategy/masking_strategy.py
eastandwestwind: Because it doesn't rely on instance vars, I usually prefer static methods.

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
eastandwestwind: Same as above

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
eastandwestwind: ended up using `get_strategy(strategy_name, {})` which will throw `NoSuchStrategyException` if strategy does not exist

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
eastandwestwind: ended up using an inline set comprehension for brevity: `for strategy_name in {rule.masking_strategy["strategy"] for rule in erasure_rules}:`

seanpreston: You could avoid the extra iteration with a check like:
```
if `rule.masking_strategy["strategy"] in unique_masking_strategies_by_name:
    continue
```

src/fidesops/service/masking/strategy/masking_strategy.py
seanpreston: In the circumstances we're calling these methods we aren't calling them as staticmethods. If we're going to leave them static we should call them as such, it'll be confusing to developers otherwise. Consider removing the staticmethod annotation, since we're already instantiating the objects used to call the methods.

eastandwestwind: ok makes sense!

src/fidesops/service/masking/strategy/masking_strategy_aes_encrypt.py
eastandwestwind: this one, on the other hand, is only called as a static method. Is there a benefit to preferring instance methods instead here?

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
eastandwestwind: ahhh nice, makes sense to me!

src/fidesops/models/privacy_request.py
eastandwestwind: here it is: https://github.com/ethyca/fidesops/issues/134

eastandwestwind:
Back over to you @seanpreston , thanks so much for the review! Let me know your thoughts for the nonce generation piece (specifically trimming bytes)

seanpreston:
Have done some functional testing on this today by plugging the masking strategies into erasures:
- `HASH` works as expected 🎉 
- `AES` appears to be erroring. @eastandwestwind have just spent some time debugging and it looks like this is because we're trying to cache the secrets as bytes, which is causing the following error: `"'utf-8' codec can't decode byte 0xfd in position 5: invalid start byte"`
- We've worked out that some good next steps would be:
  - Write a unit test to specifically test caching the secrets for AES GCM
  - Write a unit test for an AES GCM erasure
    - This will be interesting because `request_runner_service_test.get_privacy_request_results` doesn't emulate the secrets caching, so we'll need to either add that in or move secrets caching to the `PrivacyRequestRunner` class.

All in all great work @eastandwestwind, just one little push left

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -240,6 +260,58 @@ def test_create_and_process_erasure_request_generic_category(
...
+
+    customer_found = False
+    for row in res:
+        if customer_id in row:
+            customer_found = True
+            # Check that the `email` field is `None` and that its data category
+            # ("user.provided.identifiable.contact.email") has been erased by the parent
+            # category ("user.provided.identifiable.contact")
+            assert row[1] is not "customer-2@example.com"
+            assert row[2] is not None
```
seanpreston, line 308: Just tested this, it works, thanks @eastandwestwind! One last small ask: can we test for the encrypted value here too?

PR #124, opened 2021-12-10
Run Policy Webhooks as part of Privacy Request Execution [#101]

### LINKED ISSUE #101: [HTTPS Connectors] Run Pre-Execution and Post-Execution Policy Webhooks

pattisdr:
# Purpose

Finally connect running Pre-Execution and Post-Execution Policy Webhooks as part of a Privacy Request.  Any configured PolicyPreWebhooks will run before the privacy request is executed, which can potentially halt execution or add to the identity graph.  After the query traversal has completed, run configured `PolicyPostWebhooks`. PolicyPostWebhooks can't `halt` execution, but an error will put the PrivacyRequest in an errored state.

Also connect resuming privacy request execution from a given webhook. For example, if you have three webhooks, and the first webhook needs more processing time (and replied with `halt: True`), your app should send a request to `/privacy_request/<privacy_request_id>/resume` with the `reply-to-token` the webhook was originally given.  Enable the privacy request to continue executing from the second webhook.  

# Changes

- Removes instantiating the PrivacyRequestRunner with a session, as one is generated internally.
  -  pass in the new session when running webhooks
- Fix: Update the resume_privacy_request endpoint to only allow `paused` requests to be resumed
- Fix: Exit privacy request execution if we are unable to make a connection to a webhook instead of letting it hang 
- Add running Pre- and Post-Execution webhooks as part of the `PrivacyRequestRunner.submit` flow
- When a privacy request is paused, use the `ApScheduler` to schedule a job when that redis cache expires to clean up the privacy request's status to `error` since they will no longer be able to resume that privacy request.

# Checklist

# Ticket

Fixes #101


src/fidesops/service/privacy_request/onetrust_service.py
```diff
@@ -157,7 +157,6 @@ def _create_privacy_request(  # pylint: disable=R0913
         try:
             PrivacyRequestRunner(
                 cache=get_cache(),
-                db=db,
```
stevenbenjamin, line 160: Good catch!

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -25,79 +29,89 @@
...
         privacy_request: PrivacyRequest,
     ):
         self.cache = cache
-        self.db = db
         self.privacy_request = privacy_request
 
-    def run_webhooks(
-        self, webhook_cls: WebhookTypes, after_webhook: Optional[WebhookTypes] = None
+    @staticmethod
+    def run_webhooks_and_report_status(
```
stevenbenjamin, line 58: As a general comment, would it make more sense to run the webhooks as a step in the task runner as some kind of pre-traversal task? That way the BackgroundScheduler would not be necessary.

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -25,79 +29,89 @@
...
+_scheduler = None
+
+
+def get_scheduler() -> BackgroundScheduler:
+    """Returns a BackgroundScheduler as a singleton"""
+    global _scheduler  # pylint: disable=W0603
+    if _scheduler is None:
+        _scheduler = BackgroundScheduler()
+        _scheduler.start()
+    return _scheduler
```
stevenbenjamin, line 43: Is instantiating in this way vulnerable to a race condition?

pattisdr (commented):
thanks for your comments @stevenbenjamin - I have some followup questions:

src/fidesops/service/privacy_request/request_runner_service.py
stevenbenjamin: This is a broken singleton pattern, e.g. https://stackoverflow.com/a/33000332/19479 .We're instantiating on demand here. Granted the worst that could happen is an additional scheduler can be created.

pattisdr: ah thank you for clarifying! this link is super helpful.

src/fidesops/service/privacy_request/request_runner_service.py
stevenbenjamin: Ah, thanks. I do think that if we're going to have a background scheduler it would be a nice refactor to  define it in its own file. Right now we're also using this in `tasks.initiate_scheduled_request_intake` which is being called by `main`, so it's not limited to the privacy request.

pattisdr: it was in its own file originally! I moved here for circular import reasons, it's tricky because `initiate_scheduled_request_intake` calls the privacy request runner, i'll see if i can figure out something better.

src/fidesops/service/privacy_request/request_runner_service.py
pattisdr: @stevenbenjamin when you get a chance, i've tried to fix this in the latest commit -

src/fidesops/service/privacy_request/request_runner_service.py
pattisdr: ok I've moved into its own file, plus some adjustments to not instantiate on demand!

PR #131, opened 2021-12-16
Redshift Query Execution [#74]

### LINKED ISSUE #74: Query execution for Redshift
Requires #45 to be complete first.

We need to enable the Redshift database connector to be able to read/write to the database itself during query execution.

pattisdr:
# Purpose

Add query execution support for Redshift.

# Changes
- Add `db_schema` attribute for use when setting Redshift secrets in addition to the url.  This should be used if the desired schema is not `public`.
- Override `retrieve_data` and `mask_data` on the RedshiftConnector to first set the `search_path` from the `db_schema` if applicable which persists until the connection is closed

# Checklist

# Ticket

Fixes #74


src/fidesops/service/connectors/sql_connector.py
```diff
@@ -187,6 +196,58 @@ def create_client(self) -> Engine:
...
+        """
+        query_config = self.query_config(node)
+        client = self.client()
+        stmt = query_config.generate_query(input_data, policy)
+        if stmt is None:
+            return []
+
+        logger.info(f"Starting data retrieval for {node.address}")
+        with client.connect() as connection:
+            self.set_schema(connection)
```
seanpreston, line 223: If this persists for the duration of the session, do we need to call it every time we retrieve data?

tests/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -34,7 +34,7 @@ def _reject_key(dict: Dict, key: str) -> Dict:
 def test_example_datasets(example_datasets):
     """Ensure the test fixture loads the right sample data"""
     assert example_datasets
-    assert len(example_datasets) == 3
+    assert len(example_datasets) == 4
```
seanpreston, line 37: This is maybe part of the wider testing changes. I'm not sure what value adding every new dataset to `example_datasets` brings us vs. loading them individually. Do you have a reason here or is it to respect convention?

pattisdr: this existing test just loads in all datasets, so I just changed it because tests were failing - I can just get rid of this test.

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -412,6 +415,156 @@ def test_create_and_process_erasure_request_snowflake(
...
+    connector = redshift_resources["connector"]
+    redshift_client = redshift_resources["client"]
+    with redshift_client.connect() as connection:
+        connector.set_schema(connection)
+
+        address_id = redshift_resources["address_id"]
+        stmt = f"select 'id', city, state from address where id = {address_id};"
+        res = connection.execute(stmt).all()
+        for row in res:
+            # Confirms address id/street are still there, but state fields have been replaced by null values.
```
seanpreston, line 562: Great comment. A good extension could be to write a util method for these queries that could take the query results and unpack them into a dictionary to make it easier to read

src/fidesops/service/connectors/sql_connector.py
pattisdr: Maybe i phrased it wrong, it lasts until this connection is closed -

pattisdr: I originally tried to set it more broadly, but then the schema doesn't seem to be set for individual connections -

tests/service/privacy_request/request_runner_service_test.py
pattisdr: it looks like we can access ResultProxy results like they are dictionary keys, instead of accessing them as tuple elements, I'll refactor!

src/fidesops/service/connectors/query_config.py
```diff
@@ -392,6 +392,19 @@ def get_formatted_update_stmt(
...
+class RedshiftQueryConfig(SQLQueryConfig):
+    """Generates SQL in Redshift's custom dialect."""
+
+    def get_formatted_query_string(
+        self,
+        field_list: str,
+        clauses: List[str],
+    ) -> str:
+        """Returns a query string with double quotation mark formatting for tables that have the same names as
+        Redshift reserved words."""
```
seanpreston, line 404: > same names as Redshift reserved words

This is a great edge case, thanks. Is there a way to programmatically know which words are reserved without looking at the docs? Say, if it was stored in a meta schema

pattisdr: I haven't figured out how to do that yet, but there is `from sqlalchemy_redshift.dialect import RESERVED_WORDS`

seanpreston: This is awesome. They have this for Snowflake too https://github.com/snowflakedb/snowflake-sqlalchemy/blob/master/base.py#L19 which I'll make a note to add retroactively

src/fidesops/service/connectors/sql_connector.py
```diff
@@ -187,6 +196,58 @@ def create_client(self) -> Engine:
             echo=not self.hide_parameters,
         )
 
+    def set_schema(self, connection: Connection) -> None:
+        """Sets the search_path for the duration of the session"""
+        config = RedshiftSchema(**self.configuration.secrets or {})
+        if config.db_schema:
+            logger.info("Setting Redshift search_path before retrieving data")
+            connection.execute("SET search_path TO %s" % config.db_schema)
```
stevenbenjamin, line 204: We should not send raw strings to sql - should use a sql alchemy textClause here

pattisdr: good catch thank you

tests/fixtures.py
```diff
@@ -967,6 +979,25 @@ def snowflake_example_test_dataset_config(
     dataset_config.delete(db=db)
 
 
+@pytest.fixture
```
seanpreston, line 982: We can make this a `session` level scope since the dataset won't change between tests

Issue #141, opened 2021-12-21
[Testing Improvements] Add `<env>_resources` fixtures for data creation


seanpreston:
- [ ] Replace existing data insertion methods with `pytest` fixtures

Aha! Link: https://ethyca.aha.io/features/OPS-138

PR #148, opened 2021-12-23
[Testing Improvements] Update `Makefile` and `docker-compose.integration-<env>` files to enable testing one integration env at a time

### LINKED ISSUE #138: [Testing Improvements] Update `Makefile` and `docker-compose.integration-<env>` files to enable testing one integration env at a time
This should include updating the `@pytest.marker` decorators on each test.

seanpreston:
# Purpose

This PR updates our `Makefile` commands such that we can run test environments separate of one another, or all together.

```
make pytest-integration datastores="postgres snowflake mssql"  # runs tests with the `integration_postgres` marker and uses only `docker-compose.yml` and `docker-compose.integration-postgres`
make pytest-integration  # runs tests with the `integration` marker and uses all `docker-compose.integration-...` files

make integration-env datastores=postgres  # runs the docker environment only with `docker-compose.yml` and `docker-compose.integration-postgres.yml`
make integration-env  # runs the docker environment with all `docker-compose.integration-...` files

make integration-shell datastores=postgres  # as above
make integration-shell  # as above

make quickstart datatsores="postgres mysql"
make quickstart
```

The underlying command itself has the following interface:
```
➜  fidesops git:(seanpreston-138-run-tests-independently) ✗ python3 configure_infrastructure.py -h
usage: configure_infrastructure.py [-h] [-d [DATASTORES ...]] [-t] [-p PYTEST_PATH] [-s] [-r] [-q]

optional arguments:
  -h, --help            show this help message and exit
  -d [DATASTORES ...], --datastores [DATASTORES ...]
  -t, --run_tests
  -p PYTEST_PATH, --pytest_path PYTEST_PATH
  -s, --open_shell
  -r, --run_application
  -q, --run_quickstart
```

NB. `datastores` will default to all dockerfile and external datastores if none are specified.
NB. The other flags can mostly not be used in combination — ie. the script will exit after any of `run_tests`, `open_shell`, `run_application` have been executed.
NB. `pytest_path` is only useful if `--run_tests` is also specified.

# Changes

- The updated `make ...` commands now check for a passed in `env` variable, which is used to create only the requisite infrastructure for the tests in that environment.
- NB. Anything marked `integration_erasure` is unfortunately still processed outside of this, as due to the way the integration datastores are seeded with data, we need to make sure erasure tests run separately. This will be addressed as part of https://github.com/ethyca/fidesops/issues/141.

# Workflow changes
- Anyone adding a new datastore from now on should add the Docker configuration to a file `docker-compose.integration-<env>.yml`, then add `<env>` to the vars here: https://github.com/ethyca/fidesops/pull/148/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R14-R15.
- Any custom setup for a datastore required to happen in the Python code will be run automatically if a file `tests/integration_tests/<env>-setup.py` exists.

# Ticket

Fixes #138
 
# NB

- @NevilleS + @ThomasLaPiana I've requested a review from either of you here as you're the only folks I know proficient with Make, I'm sure initially there will be *a lot* of improvements in this department. Please anyone else also chip in.
- In general I think the main improvement that would be good to find is to DRY-ify the code, since there's a lot duplicated for each command currently.
- I'm going to continue working this afternoon on seeing if specifying a list of `env`s changes things drastically — it could be that `make pytest-integration env=postgres,mysql` isn't much of a stretch from here.
- The external integration tests will fail when run here, since it's a forked repo. External integration tests can be run by checking this branch out locally, and running `make pytest-integration-external`.


Makefile
```diff
@@ -45,16 +48,70 @@ server-shell: compose-build
...
+	@path=""; \
+	for word in $(DOCKERFILE_ENVIRONMENTS); do \
+		path="$$path -f docker-compose.integration-$$word.yml"; \
+	done; \
+	docker-compose -f docker-compose.yml $$path build; \
+	docker-compose -f docker-compose.yml $$path up -d; \
+	sleep 5; \
+	setup_path=""; \
+	for word in $(DOCKERFILE_ENVIRONMENTS); do \
+		setup_path="fidesops python tests/integration_tests/$$word-setup.py"; \
```
seanpreston, line 62: These setup files need to be hyphen delineated because bash doesn't treat the underscore delineated equivalent as a variable substitution.

e.g. `setup_path="fidesops python tests/integration_tests/$$word_setup.py"; \` translates to bath looking for the file `tests/integration_tests/.py`, whereas `setup_path="fidesops python tests/integration_tests/$$word-setup.py"` substitutes `$$word` correctly into `tests/integration_tests/env-setup.py`

Makefile
```diff
@@ -45,16 +48,70 @@ server-shell: compose-build
...
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml run $(IMAGE_NAME) /bin/bash
+ifeq (,$(env))
+	@echo "no environment specified"
+	@path=""; \
+	for word in $(DOCKERFILE_ENVIRONMENTS); do \
+		path="$$path -f docker-compose.integration-$$word.yml"; \
+	done; \
+	docker-compose -f docker-compose.yml $$path build; \
+	docker-compose -f docker-compose.yml $$path up -d; \
+	sleep 5; \
```
seanpreston, line 59: This seems to run fine with a 5 second sleep but we might need to bump it up

Makefile
```diff
@@ -11,6 +11,9 @@ IMAGE_NAME := fidesops
 IMAGE := $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_TAG)
 IMAGE_LATEST := $(REGISTRY)/$(IMAGE_NAME):latest
 
+DOCKERFILE_ENVIRONMENTS := postgres mysql mongodb mssql
+EXTERNAL_ENVIRONMENTS := snowflake redshift
```
NevilleS, line 15: `ENVIRONMENT` is a really overloaded term and I think you are inviting trouble by using this and `ENV` as your variables. I'd recommend making it more specific, like `make pytest-integration INTEGRATION_DATABASES=postgres,mysql` or similar

NevilleS: And add a comment above these magic variables to describe their use

Makefile
NevilleS: These kinds of sleeps are unfortunate. I used `pg_isready` in fidesdemo to avoid it (https://github.com/ethyca/fidesdemo/blob/main/Makefile#L88), but that only works for Postgres.

I don't have any other ideas, but for clarity I'd recommend at least putting an echo statement like "Pausing for 5s to allow all integration databases to be ready..."

Makefile
NevilleS: I know this is a bit late to the game, but have you considered just forklifting this logic to a python script instead that exec's the ultimate `docker` command?

The translation would be pretty straightforward and you could use Python instead of bash, which is:
1. easier to read
2. supported on windows

...?

NevilleS: See `make docs-build` where I already broke the Makefile boundary and added a `generate_openapi.py` script for a little task like this...

NevilleS (changes requested):
First, I'd recommend renaming `env` -> `integration_databases` or similar to avoid some confusion and potential env var conflicts.

Secondly, though, see my comment about refactoring the bash logic into Python instead. I think I'd probably do that if I were you.

That said, the bash looks correct to me!

Makefile
seanpreston: Will take a look, thanks!

pattisdr:
Reviewing...

configure_infrastructure.py
```diff
@@ -0,0 +1,115 @@
...
+            # If the specified datastore is not known to us
+            os.system(f'echo "Datastore {datastore} is currently not supported"')
+
+    os.system(f'echo "infrastructure path: {path}"')
+    os.system(f"docker-compose {path} build")
+    os.system(f'echo "sleeping for: {DOCKER_WAIT} while infrastructure loads"')
+    os.system(f"sleep {DOCKER_WAIT}")
+
+    # Seed datastores with data
+    for datastore in datastores:
```
eastandwestwind, line 63: Nice, this will support datastores other than mssql in future if needed

Makefile
```diff
@@ -114,35 +108,36 @@ pytest: compose-build
...
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
-	@echo "Waiting 20s for integration containers to be ready..."
-	@sleep 20
-	@echo "Running additional setup for mssql integration tests"
-	@docker exec fidesops python tests/integration_tests/mssql_setup.py
-	@echo "Running pytest integration tests..."
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml \
-		run $(IMAGE_NAME) \
-		pytest $(pytestpath) -m "integration_erasure"
+	@path=""; \
```
eastandwestwind, line 127: why can't we re-use the `configure_infrastructure.py` for erasures?

Makefile
```diff
@@ -114,35 +108,36 @@ pytest: compose-build
...
-	@echo "Waiting 20s for integration containers to be ready..."
-	@sleep 20
-	@echo "Running additional setup for mssql integration tests"
-	@docker exec fidesops python tests/integration_tests/mssql_setup.py
-	@echo "Running pytest integration tests..."
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml \
-		run $(IMAGE_NAME) \
-		pytest $(pytestpath) -m integration
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml down --remove-orphans
+configure-infrastructure: compose-build
```
eastandwestwind, line 111: can we add this to docs? In what scenario would we use this command?

Makefile
```diff
@@ -45,23 +48,14 @@ server-shell: compose-build
...
-	@echo "Bringing up main image and images for integration testing"
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
-	@echo "Waiting 15s for integration containers to be ready..."
-	@sleep 15
-	@echo "Running additional setup for mssql integration tests"
-	@docker exec -it fidesops python tests/integration_tests/mssql_setup.py
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml logs -f -t
+	@virtualenv -p python3 fidesops_test_dispatch; \
+		source fidesops_test_dispatch/bin/activate; \
+		python configure_infrastructure.py run_application=true datastores=$(datastores)
```
eastandwestwind, line 58: nice, love the flexibility of `configure_infrastructure.py`

configure_infrastructure.py
```diff
@@ -0,0 +1,115 @@
...
+    "mssql",
+]
+EXTERNAL_DATASTORES = [
+    "snowflake",
+    "redshift",
+]
+IMAGE_NAME = "fidesops"
+
+
+def configure_infrastructure(
```
eastandwestwind, line 24: Just a thought- since this file can optional run tests and other things (on top of configure infra), is there a more appropriate name for this file?

.github/workflows/safe_pr_checks.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       - name: Unit Tests
         run: make pytest
 
-      - name: Integration Tests (Access)
-        run: make pytest-integration-access
+      - name: Integration Tests
+        run: make pytest-integration
```
eastandwestwind, line 37: so does this still only run access tests?

eastandwestwind (commented):
Great work @seanpreston, excited to have this infra in place for better testing flexibility ! Just a couple qs

pattisdr (commented):
Just early comments, a full review hasn't been completed.

Also, the most time-consuming piece of a lot of this is this section in the Dockerfile needed for MSSQL:

```
# SQL Server (MS SQL)
# https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server?view=sql-server-ver15
RUN apt-get install apt-transport-https
RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
RUN curl https://packages.microsoft.com/config/debian/10/prod.list | tee /etc/apt/sources.list.d/msprod.list
RUN apt-get update
ENV ACCEPT_EULA=y DEBIAN_FRONTEND=noninteractive
RUN apt-get -y install \
    unixodbc-dev \
    msodbcsql17 \
    mssql-tools
```

Some of this might be needed for the pyodc requirement we install later, but wondering if separating Dockerfiles might help too, so this only has to be run when working on MSSQL locally.

configure_infrastructure.py
seanpreston: It will yes, so long as we:
- configure the `docker-compose.integration-<env>.yml` file, and
- add `<env>` to the `DOCKERFILE_ENVIRONMENTS` collection atop the file

Makefile
seanpreston: We can, this is due to be handled in a subsequent ticket. The main point here is that we need to make sure we're creating test fixtures properly before we can handle erasures properly. When we handle fixture creation in the right way, we no longer need a delineation for erasure tests and they can be run at the same time as access requests (because access requests won't fail if the erasure test has run first).

Makefile
seanpreston: This is doing the same thing as `make integration-shell` so I'll remove it

.github/workflows/safe_pr_checks.yml
seanpreston: Yes, and by default it will run them for all environments

Makefile
```diff
@@ -114,35 +108,36 @@ pytest: compose-build
...
-	@echo "Bringing up the integration environment..."
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
-	@echo "Waiting 20s for integration containers to be ready..."
-	@sleep 20
-	@echo "Running additional setup for mssql integration tests"
-	@docker exec fidesops python tests/integration_tests/mssql_setup.py
-	@echo "Running pytest integration tests..."
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml \
-		run $(IMAGE_NAME) \
-		pytest $(pytestpath) -m integration
```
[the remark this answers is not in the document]
seanpreston: We can add this back for sure. I wasn't sure if it was being used in anyone's workflows, generally because spinning up all the infra to run one test is inefficient. Generally to run one test I'll use `make integration-shell` then run the `pytest ...` commands manually from inside the shell, that way you're not waiting for the infra every time.

configure_infrastructure.py
```diff
@@ -0,0 +1,115 @@
...
+    run_application: bool = False,  # Should we run the Fidesops webserver?
+    run_tests: bool = False,  # Should we run the tests after creating the infra?
+    open_shell: bool = False,  # Should we open a bash shell?
+    datastores: List[str] = [],  # Which infra should we create? If empty, we create all
+) -> None:
```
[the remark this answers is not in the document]
seanpreston: I thought about that, but since those individual functions require the infra to be configured before anything is possible, I thought maybe this was OK. I'll see what I can do

configure_infrastructure.py
```diff
@@ -0,0 +1,115 @@
...
+
+    os.system(f'echo "infrastructure path: {path}"')
+    os.system(f"docker-compose {path} build")
+    os.system(f'echo "sleeping for: {DOCKER_WAIT} while infrastructure loads"')
+    os.system(f"sleep {DOCKER_WAIT}")
+
+    # Seed datastores with data
+    for datastore in datastores:
+        if datastore in DOCKERFILE_DATASTORES:
+            setup_path = f"tests/integration_tests/{datastore}-setup.py"
```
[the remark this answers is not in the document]
seanpreston: It does, but `bash` is catching that error and printing out that the datastore has no custom config associated. I left this in like this so we had the option to configure datastore schemas like this moving forward

Makefile
eastandwestwind: ah I see, perfect!

Makefile
eastandwestwind: I see in `make integration shell` we're additionally using `open_shell=true`, so as long as we'll never need to setup the venv without the shell, as we do here

tests/integration_tests/test_mongo_task.py
```diff
@@ -98,6 +98,7 @@ def test_combined_erasure_task(
     }
 
 
+@pytest.mark.integration_mongodb
```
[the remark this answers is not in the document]
seanpreston: This is a great spot. I'm not sure whether the test would run in the current format if it was marked with both (or it could run twice)

pattisdr:
Putting a note here so I don't forget, but there's supporting documentation that will need to be updated- we reference these obsolete make commands in several places in the docs.

configure_infrastructure.py
seanpreston: There likely is, perhaps something like `dispatch_commands` is more appropriate but I couldn't think of anything obvious that made sense so I left it. Open to suggestions of course

Makefile
seanpreston: We aren't currently doing that, if we do need to set it up without the shell we can always run the command then exit the shell

Makefile
```diff
@@ -114,35 +109,36 @@ pytest: compose-build
...
-		pytest $(pytestpath) -m "integration_erasure"
+	@path=""; \
+	for word in $(DOCKERFILE_ENVIRONMENTS); do \
+		path="$$path -f docker-compose.integration-$$word.yml"; \
+	done; \
+	docker-compose -f docker-compose.yml $$path build; \
+	docker-compose -f docker-compose.yml $$path up -d; \
+	sleep 5; \
+	setup_path=""; \
+	for word in $(DOCKERFILE_ENVIRONMENTS); do \
```
seanpreston, line 136: I'm going to leave these changes in for now since they'll be changed again in the next PR anyway

tests/integration_tests/test_mongo_task.py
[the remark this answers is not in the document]
seanpreston: Just looked at changing this — was this commend in reference to the test above `test_combined_erasure_task`?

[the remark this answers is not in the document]
pattisdr: yes! i couldn't comment higher than this line

Makefile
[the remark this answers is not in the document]
seanpreston: Update: I've added this back in

tests/integration_tests/test_mongo_task.py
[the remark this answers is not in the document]
seanpreston: I remember now why I didn't do this. Right now the test only has `@pytest.mark.integration`. If we add either of the others, the test will fail if somebody runs only `make pytest-integration datastores=mongodb` or `make pytest-integration datastores=postgres` because the test will be triggered and one set of infrastructure will be unavailable. Pytest doesn't provide a way to combine markers and only run the test if both are present.

This is an issue though as this new script no longer uses `-m integration` in favour of running each specific mark individually. I'll have to add the generic one back in somehow.

An alternative could be `@pytest.mark.integration_postgres_and_mongo`, but that seems to defeat the point of splitting these out in the first place, and there's no obvious way we'd trigger it.

seanpreston:
> Putting a note here so I don't forget, but there's supporting documentation that will need to be updated- we reference these obsolete make commands in several places in the docs.

The commands have been changed such that they're backwards compatible, so what's in the docs won't break. Will update them with the new args too where appropriate thanks for the reminder

configure_infrastructure.py
eastandwestwind: Something like`datastore_infrastructure_dispatcher.py` or `command_dispatcher.py` or `infrastructure_command_dispatcher.py` comes to mind. Something that denotes we're running commands on datastores/infra, not just setting up the env.

seanpreston: Have renamed to `run_infrastructure`

tests/integration_tests/test_mongo_task.py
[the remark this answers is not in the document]
pattisdr: In short, a combined test that extends multiple databases needs to bring up all of the databases?

configure_infrastructure.py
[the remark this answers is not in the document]
pattisdr: OK - I still don't love this error that it throws, it makes it seem like all the postgres setup hasn't been completed, even though that's not true, postgres doesn't need this file, but we can leave it in if you want. 

Screen Shot 2022-01-25 at 10 16 38 AM

tests/integration_tests/test_mongo_task.py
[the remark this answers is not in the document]
seanpreston: Not technically of course, it's just hard to specify the relationship between which DBs to bring up and which tests to execute otherwise

run_infrastructure.py
```diff
@@ -0,0 +1,234 @@
...
+    if len(datastores) == 0:
+        os.system(
+            f'echo "no datastores specified, configuring infrastructure for all datastores"'
+        )
+        datastores = DOCKERFILE_DATASTORES + EXTERNAL_DATASTORES
+    else:
+        os.system(f'echo "datastores specified: {", ".join(datastores)}"')
+
+    # Configure docker-compose path
+    path: str = get_path_for_datastores(datastores)
```
pattisdr, line 51: maybe we also modify datastores to be `set(datastores)` to make sure the same datastore isn't passed in twice -

run_infrastructure.py
```diff
@@ -0,0 +1,234 @@
...
+            # If the specified datastore is not known to us
+            os.system(f'echo "Datastore {datastore} is currently not supported"')
+
+    return path
+
+
+def _run_quickstart(
+    path: str,
+    image_name: str,
+) -> None:
```
pattisdr, line 118: Can you add a docstring here?

tests/integration_tests/test_connection_configuration_integration.py
```diff
@@ -27,6 +30,7 @@ class TestPostgresConnectionPutSecretsAPI:
     def url(self, oauth_client: ClientDetail, policy, connection_config) -> str:
```
pattisdr, line 30: This `TestPostgresConnectionPutSecretsAPI` could also probably have the marks on the class instead of individual functions -

tests/integration_tests/test_integration_postgres_example.py
```diff
@@ -26,6 +26,7 @@ def postgres_example_db() -> Generator:
     engine.dispose()
 
 
+@pytest.mark.integration_postgres
```
pattisdr, line 29: Similar test in `test_integration_mongodb_example.py` could be marked with mongo - and test_integratino_mssql_example.py marked with mssql

tests/integration_tests/test_mongo_task.py
```diff
@@ -261,6 +264,7 @@ def test_composite_key_erasure(
     assert access_request_data["mongo_test:composite_pk_test"][0]["description"] == None
 
 
+@pytest.mark.integration_mongodb
```
pattisdr, line 267: class `TestRetrievingDatMongo` below could be marked at the class level, and additional marked with mongo -

tests/integration_tests/test_sql_task.py
```diff
@@ -466,6 +471,7 @@ def test_filter_on_data_categories(
     rule_target.delete(db)
 
 
+@pytest.mark.integration_postgres
```
pattisdr, line 474: Similarly, TestRetrievingData below could be marked with integration at the class level, and also marked with just `postgres`

run_infrastructure.py
```diff
@@ -0,0 +1,234 @@
...
+    elif run_quickstart:
+        return _run_quickstart(path, IMAGE_NAME)
+
+    elif run_tests:
+        # Now run the tests
+        return _run_tests(
+            datastores,
+            docker_compose_path=path,
+            pytest_path=pytest_path,
+        )
```
pattisdr, line 79: Really appreciate you taking the time to break this out - this python script is capable of doing a lot of different things, and we will probably be adding to it in the future.  I think it's a lot easier to figure out its capabilities like this, and quickly locate the piece you need to change.

run_infrastructure.py
```diff
@@ -0,0 +1,234 @@
...
+            # Otherwise only include the datastores provided
+            for datastore in datastores:
+                if len(pytest_markers) == 0:
+                    pytest_markers += f"integration_{datastore}"
+                else:
+                    pytest_markers += f" or integration_{datastore}"
+
+    environment_variables = ""
+    for datastore in EXTERNAL_DATASTORES:
+        if datastore in datastores:
```
pattisdr, line 171: In this script you do a lot of looping through one list and seeing if that value is in another list. These lists are small, so it doesn't matter, but I would loop through `set(a).intersection(b)`.

Makefile
pattisdr: You also don't appear to be using EXTERNAL_ENVIRONMENTS.

Makefile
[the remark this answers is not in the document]
pattisdr: thank you so much!

Makefile
```diff
@@ -114,35 +109,36 @@ pytest: compose-build
...
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml \
-		run $(IMAGE_NAME) \
-		pytest $(pytestpath) -m integration
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml down --remove-orphans
+configure-infrastructure: compose-build
+	@echo "creating virtualenv"
+	@virtualenv -p python3 fidesops_test_dispatch
+	@echo "activating virutalenv"; \
+		source fidesops_test_dispatch/bin/activate; \
+		python run_infrastructure.py --datastores $(datastores)
```
pattisdr, line 117: Typo in the echo here - but it sounds like you were going to remove this?

docs/fidesops/docs/development/overview.md
```diff
@@ -24,12 +24,13 @@ commands to give you different functionality.
...
 - `make integration-env` - spins up everything in make server, plus additional postgres, mongo, and mysql databases for you to execute privacy requests against.
     - Try this out locally with our [Fidesops Postman Collection](../postman/Fidesops.postman_collection.json)
+- `make integration-shell` - spins up everything in make server, plus all Docker Compose specified datastores (Postgres, MySQL, MSSQL, Mongodb) and opens a server shell. This is most useful for running `pytest -k ...` commands within the integration shell to test connected datastores.
 - `make black`, `make mypy`, and `make pylint` - auto-formats code
 - `make check-all` - runs the CI checks locally and verifies that your code meets project standards
 - `make server-shell`-  opens a shell on the Docker container, from here you can run useful commands like:
   - `ipython` - open a Python shell
 - `make pytest` - runs all unit tests except those that talk to integration databases
-- `make pytest-integration-access` - runs access integration tests.
+- `make pytest-integration` - runs access integration tests.
```
pattisdr, line 33: Can we add a few more examples of the specifics introduced in this PR? for example, we talked about how this formatting is nonobvious (for posterity: we're doing it like this b/c we'd need more custom bash code to format these arguments, but we're trying to move some of the complexity to python)

make pytest-integration datastores="postgres snowflake mssql"

pattisdr (commented):
thanks for the work here @seanpreston, nice incremental improvement to be able to run just certain external datastores at a time, plus it moves a lot of the complicated things out of the makefile and into python, which I like.

Tests are failing and PR out of date.  The merge may need a few more new integration tests annotated with their correct datastore.

seanpreston:
It's worth noting that the Unsafe won't pass here, since it's a forked PR it doesn't have access to the repo secrets even with the label

pattisdr:
> It's worth noting that the Unsafe won't pass here, since it's a forked PR it doesn't have access to the repo secrets even with the label

I didn't even notice that! Thank you Sean!

pattisdr (approved):
thanks for your work here @seanpreston!

Makefile
```diff
@@ -11,6 +11,8 @@ IMAGE_NAME := fidesops
 IMAGE := $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_TAG)
 IMAGE_LATEST := $(REGISTRY)/$(IMAGE_NAME):latest
 
+DOCKERFILE_ENVIRONMENTS := postgres mysql mongodb mssql
```
NevilleS, line 14: nit: don't forget to delete all these in the follow-up PR to handle erasures

Makefile
```diff
@@ -45,27 +47,19 @@ server-shell: compose-build
...
-	@echo "Bringing up main image and images for integration testing"
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
-	@echo "Waiting 15s for integration containers to be ready..."
-	@sleep 15
-	@echo "Running additional setup for mssql integration tests"
-	@docker exec -it fidesops python tests/integration_tests/mssql_setup.py
-	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml run $(IMAGE_NAME) /bin/bash
+	@virtualenv -p python3 fidesops_test_dispatch; \
+		source fidesops_test_dispatch/bin/activate; \
+		python run_infrastructure.py --open_shell --datastores $(datastores)
```
NevilleS, line 52: Interesting that you need the (kinda ugly) `virtualenv` wrapper here. Assuming you *need* a venv, how about this instead: `./fidesops_test_dispatch/bin/python3 run_infrastructure.py ...`

This just uses the venv's python to run the command and is something I've seen in other projects.

You could also have a *classic* make target called `integration-venv` that just creates ./fidesops_test_dispatch.

So to summarize:
```suggestion
integration-shell: compose-build integration-venv
    ./fidesops_test_dispatch/bin/python3 run_infrastructure.py --open_shell --datastores $(datastores)
    
# Create the venv as needed    
integration-venv:
  virtualenv -p python3 fidesops_test_dispatch
```

run_infrastructure.py
```diff
@@ -0,0 +1,241 @@
...
+
+def seed_initial_data(
+    datastores: List[str],
+    path: str,
+    base_image: str,
+) -> None:
+    """
+    Seed the datastores with initial data as defined in the file at `setup_path`
+    """
+    for datastore in datastores:
```
NevilleS, line 94: nit: I think you should add some more logging here like:
```
echo "Seeding initial data for all datastores..."
echo "Seeding initial data for postgres from tests/integration_tests/postgres_setup.py..."
```

etc.

run_infrastructure.py
```diff
@@ -0,0 +1,241 @@
...
+        if datastore in DOCKERFILE_DATASTORES:
+            setup_path = f"tests/integration_tests/{datastore}_setup.py"
```
NevilleS, line 96: This does seem like a better pattern than the (convenient, but dangerous) SQL script that the postgres/mysql/etc. images use to auto-populate their code. That sounds like a separate feature request, though, to "Migrate SQL setup scripts to Python scripts for Postgres, MySQL, etc.", and would be a pretty good issue for new contributors

NevilleS (approved):
LGTM, with a couple nitpicks on the Makefile and some suggestions for more helpful logging, but wouldn't consider these blockers. You may want to just open up a few more tickets in the backlog for additional polish in this area

Makefile
NevilleS: Oh actually even better, you can leverage makes own intelligence to only run the target when needed...
```
integration-shell: compose-build fidesops_test_dispatch/bin/python3
    ./fidesops_test_dispatch/bin/python3 run_infrastructure.py --open_shell --datastores $(datastores)
    
# Create the venv as needed    
fidesops_test_dispatch/bin/python3:
  virtualenv -p python3 fidesops_test_dispatch
```

run_infrastructure.py
seanpreston: Ultimately, with the next ticket, we'll move only schema creation to these files where required. The test data will always be added through Pytest fixtures so that we can run access and erasure tests together and not through separate markers

Makefile
seanpreston: If you read the logs in Github actions the virtualenv command actually fails there. This is required for local running where users have `python` set to invoke Python 2 — open to suggestions to make this better in a follow up

seanpreston:
Thanks @pattisdr @NevilleS @eastandwestwind for your reviews

PR #152, opened 2022-01-07
Fix Quickstart and Tutorial

### LINKED ISSUE #121: Fix quickstart after privacy requests moved to fire and forget
# Purpose

Privacy requests were recently moved to "fire and forget" https://github.com/ethyca/fidesops/pull/106, which breaks the quickstart (`make quickstart`). The quickstart tries to execute an access request and print the local results to the user, but it was written for synchronous requests. It's possible this broke `fidesdemo` too.

The quickstart needs to wait for the privacy request to execute in the background and then deliver the results.  There are some unit tests that have similar logic that could be borrowed.

pattisdr:
# Purpose
The quickstart (`make quickstart`) was broken by the fire and forget behavior and the tutorial was broken when we moved several endpoints from PUT -> PATCH https://github.com/ethyca/fidesops/pull/53 

# Changes
- In the quickstart, wait for the access request package to appear in fides_uploads in a while loop, since this request is no longer synchronous
- Update the relevant requests to be PUT instead of PATCH in the tutorial instructions 

# Checklist

# Ticket

Fixes #121

### DIFF
--- a/quickstart.py
+++ b/quickstart.py
@@ -3,6 +3,8 @@ A five-step fidesops quickstart
 """
 import json
 import logging
+import os
+import time
 from datetime import datetime
 from os.path import exists
 from time import sleep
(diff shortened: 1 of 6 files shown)


quickstart.py
```diff
@@ -451,11 +453,19 @@ def print_results(request_id: str) -> None:
     print it to the console if so.
     """
     results_path = f"fides_uploads/{request_id}.json"
+
+    count = 0
+    max_allowed_waiting = 10
+    while not os.path.exists(results_path) or count > max_allowed_waiting:
+        logger.info("Waiting for privacy request results...")
+        time.sleep(5)
+        count += 1  # Only loop through a reasonable number of times
```
eastandwestwind, line 462: could we print a friendly error in case the results don't get added  at all (after max wait time) for some reason?

eastandwestwind (commented):
looks good, just 1 suggestion!

quickstart.py
pattisdr: The existing script will raise a runtime error below on L472 if the results aren't found!

eastandwestwind (approved):

pattisdr:
thank you for the quick review on this @eastandwestwind ⭐

PR #153, opened 2022-01-10
MSSQL Init DB Experiments

pattisdr:
# Purpose

Experiments trying to initialize a test mssql db and populate it when we bring up the mssql container on M1.

Known issues:

1.  We can't run sql server in docker on M1 because there's no arm64 image - using azure instead
2. While PostgreSQL lets you put an init SQL script in docker-entrypoint-initdb.d, this isn't available for mssql 
3. [Sqlcmd tool not available inside the arm64 version of edge containers](https://docs.microsoft.com/en-us/azure/azure-sql-edge/disconnected-deployment)

# Changes

Three experiments:

1. First tries to create a Dockerfile where we bring up azure and then use sqlcmd to run an initialization script but it looks like I can't use sqlcmd
2. Second tries to bring up an additional container just to run a python script to populate database after all the other containers are up when running `make integration-env`.   `Errored with Failed to load /var/opt/mssql/mssql.conf ini file with error open /var/opt/mssql/mssql.conf: no such file or directory`.   Maybe this is permission issues, not sure.  This solution isn't ideal because I have to reinstall a lot of things that are in the fidesops container - it would be nice to figure out how to run this python script on the fidesops container - 
3. Running the `mssql_setup` python function which creates a mssql database and populates when we run `start_webserver`.Works but not great, as we we're combining methods to populate test databases with our core method to bring up fidesops.


pattisdr:
Seems to be solved with bringing up docker containers in detached mode including mssql one, then exec-ing into the fidesops to run the python script which creates an mssql_example db and populates, and then we do docker logs for all these containers afterwards.
```
integration-env: compose-build
	@echo "Bringing up main image and images for integration testing"
	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml up -d
	@echo "Waiting 15s for integration containers to be ready..."
	@sleep 15
	@echo "Running additional setup for mssql integration tests"
	@docker exec -it fidesops python tests/integration_tests/mssql_setup.py
	@docker-compose -f docker-compose.yml -f docker-compose.integration-test.yml logs -f -t
```

PR #154, opened 2022-01-10
Add FieldPaths to represent Nested Fields [#158]

### LINKED ISSUE #158: Add FieldPath Construct to handle Nested Fields
Add the `FieldPath` class to represent a potentially more complex nested object as a field:

Move FieldAddress from `datasetname, collectionname, fieldname` to`datasetname, collectionname, FieldPath`

`FieldAddress("databasename", "tablename", "a", "b", "c")`:

```
dbname = "databasename"
collection_name="tablename"
field_name=FieldPath("a","b","c")
```

stevenbenjamin:
# Purpose

Add the `FieldPath` construct to handle more complex nested behavior (e.g. one table referencing another on a nested object field).  Instead of referencing a field by a string name, it will be referred to by the `FieldPath` object, which can reference a specific nested field on an object if necessary.

The FieldPath object gives us more versatility for more complex object and array queries.  Note that more complex nested behavior for mongo will be included in a follow-up PR, this work just introduces the object.

## Before

A `FieldAddress` (what's used to drive the graph, to find related fields from edges , etc) field name is a _string_:

`FieldAddress("databasename", "tablename", "fieldname")`
which translates to:

```python
dbname = "databasename"
collection_name="tablename"
field_name="field_name" (str)
```

## After 

A FieldAddress can hold potentially many nested fields:

`FieldAddress("databasename", "tablename", "a", "b", "c") `

```python
dbname = "databasename"
collection_name="tablename"
field_name=FieldPath("a","b","c")
```

# Changes
Big picture when we parse yaml into `FidesopsDatasets` and from there create `Datasets`, and turn those `Datasets` into a `DatsetGraph`, and use that `DatasetGraph` to create a Traversal - these locations are using the more complex `FieldPath` instead of strings when referencing fields.   However databases don't know what to do with a `FieldPath`, so methods in the `QueryConfig` transform the `FieldPath` to its string path version by converting FieldPath("a", "b", "c") to `a.b.c.` (search for instances of `string_path` in the code to see where we make the switch).  More logic will be added here in a follow-up PR.

- `Node.contains_field` method now looks to see if a function applies on nested fields
- Edges can be built between nested fields
- An identity can be defined on a nested field
- In the yaml, a field can reference another nested field using dot notation:
```yaml
name: name
            fidesops_meta:
                references:
                    - dataset: mongo_1
                      name: users.contact_information.name.full_name
                      direction: from
```
- Traversal node children and parents are linked along `FieldPaths` not strings
- More tests added
- Variable names changed to indicate `FieldPath` construct instead of field
- Some one-name variable names and lambda usages removed in favor of list comprehensions
- Some unused methods removed
- IMPORTANT: In this first pass, when we go to make queries, a `FieldPath` is turned into its string version, `a.b.c`.  There is more logic to add here for mongo to support some of the requirements - that will be in the next PR.

# Checklist

Fixes #158

### DIFF
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool = os.getenv("TESTING") == "True"
+    is_test_mode: bool = True  # os.getenv("TESTING") == "True"
     hot_reloading: bool = os.getenv("DEV_MODE") == "True"
 
     class Config:  # pylint: disable=C0115
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool = True  # os.getenv("TESTING") == "True"
+    is_test_mode: bool =  os.getenv("TESTING") == "True"
     hot_reloading: bool = os.getenv("DEV_MODE") == "True"
 
     class Config:  # pylint: disable=C0115
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool =  os.getenv("TESTING") == "True"
+    is_test_mode: bool = os.getenv("TESTING") == "True"
     hot_reloading: bool = os.getenv("DEV_MODE") == "True"
 
     class Config:  # pylint: disable=C0115
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool = os.getenv("TESTING") == "True"
+    is_test_mode: bool = True #os.getenv("TESTING") == "True"
     hot_reloading: bool = os.getenv("DEV_MODE") == "True"
 
     class Config:  # pylint: disable=C0115
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool = True #os.getenv("TESTING") == "True"
+    is_test_mode: bool = os.getenv("TESTING") == "True"
     hot_reloading: bool = os.getenv("DEV_MODE") == "True"
 
     class Config:  # pylint: disable=C0115
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -150,9 +150,9 @@ class FieldKey:
     A scalar field is represented as a single-element tuple.
     """
 
-    def __init__(self, names: Tuple[str, ...]):
-        self.names = names
-        self.value = ".".join(names)
+    def __init__(self, *names: str):
+        self.names = tuple(names)
+        self.value = ".".join(self.names)
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, FieldKey):
@@ -169,7 +169,7 @@ class FieldKey:
         return self.value < other.value
 
     def prepend(self, key: str) -> FieldKey:
-        return FieldKey((key,) + self.names)
+        return FieldKey(*((key,) + self.names))
 
 
 class FieldAddress:
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -135,7 +135,7 @@ class CollectionAddress:
 
     def field_address(self, field: FieldPath) -> FieldAddress:
         """Create a field address appended to this collection address."""
-        return FieldAddress(self.dataset, self.collection, *field.keys)
+        return FieldAddress(self.dataset, self.collection, *field.levels)
 
 
 ROOT_COLLECTION_ADDRESS: CollectionAddress = CollectionAddress("__ROOT__", "__ROOT__")
@@ -146,31 +146,31 @@ TERMINATOR_ADDRESS = CollectionAddress("__TERMINATE__", "__TERMINATE__")
 
 class FieldPath:
     """Fields are addressable by a (possibly) nested name. This key
-    represents a field name held as a tuple of possibly descending keys.
+    represents a field name held as a tuple of possibly descending levels.
     A scalar field is represented as a single-element tuple.
     """
 
     def __init__(self, *names: str):
-        self.keys = tuple(names)
-        self.value = ".".join(self.keys)
+        self.levels = tuple(names)
+        self.string_path = ".".join(self.levels)
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, FieldPath):
             return False
-        return other.keys == self.keys
+        return other.levels == self.levels
 
     def __hash__(self) -> int:
-        return hash(self.value)
+        return hash(self.string_path)
 
     def __repr__(self) -> str:
-        return f"FieldKey({self.value})"
+        return f"FieldPath('{self.string_path}')"
 
     def __lt__(self, other: "FieldPath") -> bool:
-        return self.value < other.value
+        return self.string_path < other.string_path
 
     def prepend(self, prefix: str) -> "FieldPath":
         """Return a new field key with the prefix prepended."""
-        return FieldPath(*((prefix,) + self.keys))
+        return FieldPath(*((prefix,) + self.levels))
 
     @staticmethod
     def parse(path_str: str) -> FieldPath:
@@ -192,7 +192,7 @@ class FieldAddress:
         self.dataset = dataset
         self.collection = collection
         self.field = FieldPath(*fields)
-        self.value: str = ":".join((dataset, collection, self.field.value))
+        self.value: str = ":".join((dataset, collection, self.field.string_path))
 
     def is_member_of(self, collection_address: CollectionAddress) -> bool:
         """True if this field represents a field in the given collection address."""
@@ -258,7 +258,7 @@ class Field(BaseModel, ABC):
         """Find fields or subfields satisfying the input function"""
 
     def __repr__(self) -> str:
-        return f"{self.name}:{self.data_type_converter.name}"
+        return f"{self.__class__.__name__}(name='{self.name}', data_type='{self.data_type_converter.name}', is_array={self.is_array})"
 
 
 class ScalarField(Field):
--- a/src/fidesops/graph/config.py
+++ b/src/fidesops/graph/config.py
@@ -135,7 +135,7 @@ class CollectionAddress:
 
     def field_address(self, field: FieldKey) -> FieldAddress:
         """Create a field address appended to this collection address."""
-        return FieldAddress(self.dataset, self.collection, field)
+        return FieldAddress(self.dataset, self.collection, *field.keys)
 
 
 ROOT_COLLECTION_ADDRESS: CollectionAddress = CollectionAddress("__ROOT__", "__ROOT__")
@@ -146,18 +146,18 @@ TERMINATOR_ADDRESS = CollectionAddress("__TERMINATE__", "__TERMINATE__")
 
 class FieldKey:
     """Fields are addressable by a (possibly) nested name. This key
-    represents a field name held as a tuple of possibly descending names.
+    represents a field name held as a tuple of possibly descending keys.
     A scalar field is represented as a single-element tuple.
     """
 
     def __init__(self, *names: str):
-        self.names = tuple(names)
-        self.value = ".".join(self.names)
+        self.keys = tuple(names)
+        self.value = ".".join(self.keys)
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, FieldKey):
             return False
-        return other.names == self.names
+        return other.keys == self.keys
 
     def __hash__(self) -> int:
         return hash(self.value)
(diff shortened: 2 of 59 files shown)


src/fidesops/graph/config.py
```diff
@@ -243,11 +289,33 @@ def cast(self, value: Dict[str, Any]) -> Optional[Dict[str, Any]]:
...
+        base = {FieldPath(name): self} if f(self) else {}  # pylint: disable=no-member
+        child_dicts = merge_dicts(
+            *map(lambda field: field.collect_matching(f), self.fields.values())
+        )
+        return merge_dicts(
+            base,
+            {
+                k.prepend(name): v for k, v in child_dicts.items()
+            },  # pylint: disable=no-member
+        )
```
pattisdr, line 317: ah nice how this works for multiple levels 👍

src/fidesops/graph/config.py
```diff
@@ -233,6 +269,16 @@ def cast(self, value: Any) -> Optional[Any]:
...
+        return self if len(names) == 0 else None
+
+    def collect_matching(self, f: Callable[[Field], bool]) -> Dict[FieldPath, Field]:
+        """Find fields or subfields satisfying the input function"""
+        if f(self):
+            return {FieldPath(self.name): self}  # pylint: disable=no-member
+        return {}
+
 
 class ObjectField(Field):
```
pattisdr, line 283: Hmm, ObjectFields can have data_categories and their sub fields can each have data_categories.  How do we handle erasures for data categories where there is a mismatch between the top level and lower levels?

Say we have an ObjectField with data_category A, the has two `fields` one with data_category A and one with data_category B.  If we have a policy to erase all data with category "A" - there is a conflict here -

src/fidesops/graph/config.py
```diff
@@ -298,46 +366,49 @@ class Collection(BaseModel):
...
 
-    def identities(self) -> Dict[str, Tuple[str, ...]]:
+    def identities(self) -> Dict[FieldPath, Tuple[str, ...]]:
         """return identity pointers included in the table"""
-        flds_w_ident = filter(lambda f: f.identity, self.fields)
-        return {f.name: f.identity for f in flds_w_ident}
+        return {k: v.identity for k, v in self.field_dict.items() if v.identity}
 
-    def field(self, name: str) -> Optional[Field]:
+    def field(self, key: FieldPath) -> Optional[Field]:
```
pattisdr, line 390: Should we have another name for this "key" variable we're introducing everywhere that is a FieldPath object, maybe "field_path" instead? I'm afraid this might get confused with the other key variable we use everywhere that is a "FidesOpsKey" type.

src/fidesops/graph/config.py
```diff
@@ -143,15 +144,50 @@ def field_address(self, field: str) -> FieldAddress:
...
 
 
+class FieldPath:
+    """Fields are addressable by a (possibly) nested name. This key
+    represents a field name held as a tuple of possibly descending keys.
+    A scalar field is represented as a single-element tuple.
+    """
+
+    def __init__(self, *names: str):
+        self.keys = tuple(names)
```
pattisdr, line 154: FieldPath.keys is a confusing name to me - I think it's too general - as these are descending keys. Calling keys on a dictionary would give you top level keys for example, I just think the term is too overloaded.

Maybe "levels" instead? or similar

tests/models/test_connectionconfig.py
```diff
@@ -84,6 +84,6 @@ def test_create_connection_config_errors(self, db: Session, connection_config):
             )
         assert (
             str(exc.value)
-            == "Key my_postgres_db_1 already exists in ConnectionConfig. Keys will be snake-cased names if not provided. "
+            == "Key my_postgres_db_1 already exists in ConnectionConfig. Keys will be snake-cased keys if not provided. "
```
pattisdr, line 87: I don't think this global find and replace from names -> keys is a good idea - this edit is not accurate - if a user is creating a connection config, and they don't supply a "key" in the request_body, we do create a key from the snake_cased name.

I've made this comment elsewhere, I'd like to not use "keys" as a variable name for these nested fields, we're overloading the term.

src/fidesops/service/connectors/query_config.py
```diff
@@ -242,10 +249,15 @@ class SQLQueryConfig(QueryConfig[TextClause]):
 
     def format_fields_for_query(
         self,
-        fields: List[str],
+        keys: List[FieldPath],
```
pattisdr, line 252: again "field_paths" instead of keys would be a clearer variable name here -

pattisdr (commented):
Thanks for putting an early draft up! I like the FieldPath construct, a tuple of keys describing the different potentially nested levels. I see what you mean about it touching everything, instead of referring to a FieldAddress as a string, the field portion needs to be the more specific FieldPath which affects the graph, the traversal, the queries, etc..

Just some early comments here -

src/fidesops/graph/config.py
```diff
@@ -233,6 +269,16 @@ def cast(self, value: Any) -> Optional[Any]:
 
         return value
 
+    def getChild(self, names: Tuple[str, ...]) -> Optional["Field"]:
```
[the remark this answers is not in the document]
stevenbenjamin: yes, simplifies generalized access.

src/fidesops/graph/config.py
```diff
@@ -143,15 +144,50 @@ def field_address(self, field: str) -> FieldAddress:
...
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, FieldPath):
+            return False
+        return other.keys == self.keys
+
+    def __hash__(self) -> int:
+        return hash(self.value)
+
+    def __repr__(self) -> str:
+        return f"FieldKey({self.value})"
```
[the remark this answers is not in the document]
stevenbenjamin: Ah, that's a pre-renaming holdover. Will fix.

src/fidesops/core/config.py
```diff
@@ -174,7 +174,7 @@ class FidesopsConfig(FidesSettings):
     security: SecuritySettings
     execution: ExecutionSettings
 
-    is_test_mode: bool = os.getenv("TESTING") == "True"
+    is_test_mode: bool = True  # os.getenv("TESTING") == "True"
```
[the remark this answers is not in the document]
stevenbenjamin: done.

src/fidesops/graph/config.py
```diff
@@ -219,6 +246,15 @@ def data_type(self) -> str:
...
         return self.data_type_converter.name
 
+    def getChild(self, names: Tuple[str, ...]) -> Optional["Field"]:
+        """Return the child field or None if it is not found"""
+
+    def collect_matching(self, f: Callable[[Field], bool]) -> Dict[FieldPath, Field]:
+        """Find fields or subfields satisfying the input function"""
+
+    def __repr__(self) -> str:
+        return f"{self.name}:{self.data_type_converter.name}"
```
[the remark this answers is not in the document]
stevenbenjamin: This was just to avoid overwhelming myself in some testing. The string repr has no functional effect, removing it is not an issue of any kind.

src/fidesops/graph/config.py
stevenbenjamin: Yes. logically the top level should win, IMO. You can make an argument for more finer grained control but I think that's overkill at the risk of serious overcomplication - even testing and documenting the behavior of that would be complicated and time consuming.

I think best course would be that specifying categories in a subfield is a configuration validation error. Should return a _very_ clear message.

src/fidesops/graph/config.py
stevenbenjamin: y, again, remnant of previous name.  FieldPath I thought is more descriptive, a few "key" refs floating around still...

src/fidesops/graph/config.py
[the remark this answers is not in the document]
pattisdr: yes I like how your string repr is more concise, this would be useful for me too, just recommending one that reflects the type.

stevenbenjamin:
What remains to be done is to ensure type agreement in all parts. This is made more challenging since most of the referred to types are forms of maps, and the distinctions are not visible to the type checker.


Essentially we're moving from a version where the configuration keys are held as strings, and the data moving through the graph is held with string keys, to a system where the configuration field reference is no longer a string but a FieldPAth. This means that many internal lookups are broken (see below).


E.g. previous version:

Graph:
 FieldAddress field name is a _string_ (A field address is what's used to drive the graph, to find related fields from edges , etc etc).
 In python this looks like

FieldAddress("databasename", "tablename", "fieldname")
which translates to:
```
dbname = "databasename"
collection_name="tablename"
field_name="field_name" (str)
```
We then lookup the field name in data we pass through the system, e.g. table 1 returns

the graph look for field at address

`FieldAdddress("dbname", "tablename", "email")`

and the data returned looks like

`[{"email":"s1"}, {"email":"s2"}, {"email":"s3"}]`


Now, however, since we need to encode some logic in the fields, we'll now hold addresses like

 `FieldAddress("databasename", "tablename", "a", "b", "c")`
 
which translates to
 ```
 dbname = "databasename"
 collection_name="tablename"
 field_name=FieldPath("a","b","c")
```

This breaks in places where we're looking in the data for the matching value from the field name - since there are no values matching a FieldPath object, the traversal finds no matches.

File: src/fidesops/graph/config.py (hunk 2 above)
Comment (pattisdr) [reply src/fidesops/graph/config.py:292]: thanks for the comment, will save this for the follow-up PR

File: src/fidesops/service/connectors/query_config.py [hunk 11]
```diff
@@ -242,10 +261,15 @@ class SQLQueryConfig(QueryConfig[TextClause]):
...
     ) -> List[str]:
-        """Returns fields in a format they can be added into SQL queries."""
-        return fields
+        """Returns field pathspaths in a format they can be added into SQL queries.
+
+        This currently takes no nesting into account and only returns the
+        last value from each key. It will need to be updated to support
+        nested values.
+        """
+        return [fk.levels[-1] for fk in field_paths]
```
seanpreston, line 272: Am I correct in thinking that in the event the field isn't nested, `fk.levels[-1]` will return the only element in the list?

src/fidesops/service/connectors/query_config.py
```diff
@@ -242,10 +261,15 @@ class SQLQueryConfig(QueryConfig[TextClause]):
 
     def format_fields_for_query(
         self,
-        fields: List[str],
+        field_paths: List[FieldPath],
     ) -> List[str]:
-        """Returns fields in a format they can be added into SQL queries."""
-        return fields
+        """Returns field pathspaths in a format they can be added into SQL queries.
```
seanpreston, line 266: There's an extra `paths` in `pathspaths`

src/fidesops/service/connectors/query_config.py
```diff
@@ -277,22 +301,23 @@ def generate_query(
...
-
-        filtered_data = self.typed_filtered_values(input_data)
+        filtered_data: Dict[str, Any] = self.typed_filtered_values(input_data)
 
         if filtered_data:
             clauses = []
             query_data: Dict[str, Tuple[Any, ...]] = {}
-            formatted_fields = self.format_fields_for_query(self.fields)
+            formatted_fields: List[str] = self.format_fields_for_query(
+                list(self.field_map().keys())
```
seanpreston, line 310: Am I correct here in reading that `self.field_map().keys()` will be a list of `FieldPath`s? It looks like the key in this dict is a `FieldPath` rather than a string https://github.com/ethyca/fidesops/pull/154/files#diff-f1aba318d05064942a5f135eb61663026a735a60dcd0a2bd93cd629c40c69a02R264

src/fidesops/service/connectors/query_config.py
pattisdr: exactly, yes:

```python
>>> fruits: Tuple[str] = ("apples", "bananas", "oranges")
>>> fruits[-1]
'oranges'
>>> vegetables = ("carrots",)
>>> vegetables[-1]
'carrots'
```

src/fidesops/service/connectors/query_config.py
pattisdr: good spot

src/fidesops/service/connectors/query_config.py
pattisdr: ` list(self.field_map().keys())` does return a List of FieldPaths, but those are passed into `format_fields_for_query` which turns the list of FieldPaths into a list of strings

src/fidesops/service/connectors/query_config.py
```diff
@@ -392,9 +425,10 @@ def get_formatted_query_string(
         """Returns a query string with double quotation mark formatting as required by Snowflake syntax."""
         return f'SELECT {field_list} FROM "{self.node.node.collection.name}" WHERE {" OR ".join(clauses)}'
 
-    def format_key_map_for_update_stmt(self, key_map: Dict[str, Any]) -> List[str]:
+    def format_key_map_for_update_stmt(self, fields: List[str]) -> List[str]:
         """Adds the appropriate formatting for update statements in this datastore."""
-        return [f'"{k}" = :{k}' for k in key_map]
+        fields.sort()
```
seanpreston, line 430: What's the significance of this sorting in the flow?

pattisdr: I'm not sure, my guess is that it was added for debugging/printing that was done in earlier drafts.  

https://github.com/ethyca/fidesops/commit/e96f14a5e85ecf6023e48997304c651e7e6e20be

It shouldn't have a practical significance other than making logs easier to read.

src/fidesops/util/collection_util.py
```diff
@@ -1,18 +1,20 @@
...
-    for d in dictionaries:
-        for k, v in d.items():
-            out[k].append(v)
-    return out
+    Right-hand values take precedence. That is,
+    merge_dicts({"A": 1, "B": 2}, {"A": 2, "C": 4})
+    =>  {'A': 2, 'B': 2, 'C': 4}
+    """
+    if dicts:
+        return reduce(lambda x, y: x | y, dicts) or {}
```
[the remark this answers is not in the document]
pattisdr: can't take credit for that one!

I generally try to avoid lambdas in Python, but this is pretty slick -

seanpreston (approved):
Thanks @pattisdr, alongside a good design change this also includes many appreciated type additions and var name updates!

pattisdr:
thanks for the review @seanpreston, tests passing!

PR #157, opened 2022-01-14
82 sqlserver query execution

### LINKED ISSUE #82: Query execution for SqlServer
Requires #78  to be complete first.

We need to enable the SqlServer `ConnectionConfig` database connector to be able to read/write to the database itself during query execution.

eastandwestwind:
# Purpose
Adds Microsoft SQL Server query execution functionality

# Changes
- Alters the core query generation logic to accommodate sql server (as well as the rest of the connectors).
- Override a couple SQLConnector methods since sql server returns data in a diff format from the connection.execute(stmt) call

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)

# Ticket

Fixes #82

### DIFF
--- a/src/fidesops/service/connectors/query_config.py
+++ b/src/fidesops/service/connectors/query_config.py
@@ -17,6 +17,7 @@ from fidesops.models.policy import Policy, ActionType, Rule
 from fidesops.models.privacy_request import PrivacyRequest
 from fidesops.service.masking.strategy.masking_strategy import MaskingStrategy
 from fidesops.service.masking.strategy.masking_strategy_nullify import NULL_REWRITE
+from fidesops.util.logger import NotPii
 from fidesops.util.querytoken import QueryToken
 from fidesops.service.masking.strategy.masking_strategy_factory import (
     get_strategy,
@@ -298,6 +299,10 @@ class SQLQueryConfig(QueryConfig[TextClause]):
                     pass
             if len(clauses) > 0:
                 query_str = self.get_formatted_query_string(field_list, clauses)
+                logger.info(f"building query for {self.node.address}")
+                logger.info("query str = " + query_str)
+                logger.info("query keys = " + ''.join(query_data.keys()))
+                logger.info("query vals = " + ''.join([str(i) for i in query_data.values()]))
                 return text(query_str).params(query_data)
 
         logger.warning(
@@ -333,7 +338,6 @@ class SQLQueryConfig(QueryConfig[TextClause]):
                 f"There is not enough data to generate a valid update statement for {self.node.address}"
             )
             return None
-
         query_str = self.get_formatted_update_stmt(
             update_clauses,
             pk_clauses,
--- a/src/fidesops/service/connectors/query_config.py
+++ b/src/fidesops/service/connectors/query_config.py
@@ -17,7 +17,6 @@ from fidesops.models.policy import Policy, ActionType, Rule
 from fidesops.models.privacy_request import PrivacyRequest
 from fidesops.service.masking.strategy.masking_strategy import MaskingStrategy
 from fidesops.service.masking.strategy.masking_strategy_nullify import NULL_REWRITE
-from fidesops.util.logger import NotPii
 from fidesops.util.querytoken import QueryToken
 from fidesops.service.masking.strategy.masking_strategy_factory import (
     get_strategy,
--- a/src/fidesops/service/connectors/query_config.py
+++ b/src/fidesops/service/connectors/query_config.py
@@ -255,6 +255,13 @@ class SQLQueryConfig(QueryConfig[TextClause]):
         """Returns clauses in a format they can be added into SQL queries."""
         return f"{field_name} {operator} :{field_name}"
 
+    def format_clause_for_query_in(
+        self, field_name: str, query_data_names: List[str]
+    ) -> str:
+        """Returns clauses in a format they can be added into SQL queries."""
+        in_clause = ", ".join(query_data_names)
+        return f"{field_name} IN ({in_clause})"
+
     def get_formatted_query_string(
         self,
         field_list: str,
@@ -282,27 +289,44 @@ class SQLQueryConfig(QueryConfig[TextClause]):
 
         if filtered_data:
             clauses = []
-            query_data: Dict[str, Tuple[Any, ...]] = {}
+            query_data: List[Dict[str, Any]] = []
             formatted_fields = self.format_fields_for_query(self.fields)
             field_list = ",".join(formatted_fields)
+            from sqlalchemy import bindparam
+
             for field_name, data in filtered_data.items():
                 data = set(data)
                 if len(data) == 1:
                     clauses.append(self.format_clause_for_query(field_name, "="))
-                    query_data[field_name] = (data.pop(),)
+                    # SQL Alchemy + SQLServer errors when we use a tuple in "value"
+                    query_data.append({"name": field_name, "value": data.pop()})
                 elif len(data) > 1:
-                    clauses.append(self.format_clause_for_query(field_name, "IN"))
-                    query_data[field_name] = tuple(data)
+                    # converts to list so that we can generate indexes
+                    newdata = list(data)
+                    query_data_names: List[str] = []
+                    for i in newdata:
+                        # appending "_in_stmt_generated_" (can be any arbitrary str) so that this name has less change of conflicting with pre-existing column in table
+                        query_data_name = "_in_stmt_generated_" + str(newdata.index(i))
+                        query_data.append({"name": query_data_name, "value": i})
+                        query_data_names.append(":" + query_data_name)
+                    # SQL Alchemy + SQLServer errors when the query_str for IN statements was structured like this:
+                    # SELECT order_id,product_id,quantity FROM order_item WHERE order_id IN :some-params-in-tuple
+                    # Current structure of query_str:
+                    # SELECT order_id,product_id,quantity FROM order_item WHERE order_id IN (:_in_stmt_generated_0, :_in_stmt_generated_1, :_in_stmt_generated_2)
+                    clauses.append(
+                        self.format_clause_for_query_in(field_name, query_data_names)
+                    )
                 else:
                     #  if there's no data, create no clause
                     pass
             if len(clauses) > 0:
                 query_str = self.get_formatted_query_string(field_list, clauses)
                 logger.info(f"building query for {self.node.address}")
-                logger.info("query str = " + query_str)
-                logger.info("query keys = " + ''.join(query_data.keys()))
-                logger.info("query vals = " + ''.join([str(i) for i in query_data.values()]))
-                return text(query_str).params(query_data)
+                stmt = text(query_str)
+                for i in query_data:
+                    # technically don't need to use this bindparams() method, we can go back to using just params()
+                    stmt = stmt.bindparams(bindparam(i["name"], value=i["value"]))
+                return stmt
 
         logger.warning(
             f"There is not enough data to generate a valid query for {self.node.address}"
--- a/src/fidesops/service/connectors/sql_connector.py
+++ b/src/fidesops/service/connectors/sql_connector.py
@@ -362,9 +362,9 @@ class MicrosoftSQLServerConnector(SQLConnector):
         # todo: log everything for mssql- config option
         # todo- try passing in schema
 
-        from sqlalchemy.dialects import mssql
-        full_stmt = stmt.compile(compile_kwargs={"literal_binds": True})
-        import pdb; pdb.set_trace()
+        # from sqlalchemy.dialects import mssql
+        # full_stmt = stmt.compile(compile_kwargs={"literal_binds": True})
+        # import pdb; pdb.set_trace()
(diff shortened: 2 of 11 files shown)


src/fidesops/service/connectors/query_config.py
```diff
@@ -255,6 +255,13 @@ def format_clause_for_query(
         """Returns clauses in a format they can be added into SQL queries."""
         return f"{field_name} {operator} :{field_name}"
 
+    def format_clause_for_query_in(
+        self, field_name: str, query_data_names: List[str]
+    ) -> str:
+        """Returns clauses in a format they can be added into SQL queries."""
+        in_clause = ", ".join(query_data_names)
+        return f"{field_name} IN ({in_clause})"
+
```
seanpreston, line 264: This method will currently live on every `SQLQueryConfig`, even where we don't need special `IN` clause formatting, which would be confusing for any developer checking the class def.

To stick with convention, could we instead:
- create a new `MicrosoftSQLServerQueryConfig(SQLQueryConfig)`  class
- override `format_clause_for_query()` on that class that returns the correctly formatted clause where `operator="IN"`
- add `query_config()` to the `MicrosoftSQLServerConnector` class and have it return the new `MicrosoftSQLServerQueryConfig`

src/fidesops/service/connectors/query_config.py
```diff
@@ -282,23 +289,44 @@ def generate_query(
...
+                        query_data_name = "_in_stmt_generated_" + str(newdata.index(i))
+                        query_data.append({"name": query_data_name, "value": i})
+                        query_data_names.append(":" + query_data_name)
+                    # SQL Alchemy + SQLServer errors when the query_str for IN statements was structured like this:
+                    # SELECT order_id,product_id,quantity FROM order_item WHERE order_id IN :some-params-in-tuple
+                    # Current structure of query_str:
+                    # SELECT order_id,product_id,quantity FROM order_item WHERE order_id IN (:_in_stmt_generated_0, :_in_stmt_generated_1, :_in_stmt_generated_2)
+                    clauses.append(
+                        self.format_clause_for_query_in(field_name, query_data_names)
+                    )
```
seanpreston, line 318: All this custom query generation logic could also be moved to the new `MicrosoftSQLServerQueryConfig`, which is the only place it's required

seanpreston:
Hey @eastandwestwind I'd say you're approach is sound, like you said in the standup it's mostly just tidy up work

src/fidesops/service/connectors/query_config.py
eastandwestwind: See my comment here- https://github.com/ethyca/fidesops/pull/157/files/35c47ce05d225c529754494d500602f29dd9615b#r786866514 

If we can use the same way to build queries for all connectors, I personally would prefer that, since it'll be less code to maintain

pattisdr: My general hang-up with this is that you've added to the `generate_query` complexity quite a bit, also making it diverge even more from the Mongo implementation too.  I'm against bloating the `SQLQueryConfig` class with a bunch of handling necessary for single datastores only.  Your approach doesn't necessarily mean "less code to maintain" - it means more datastores need to be scrutinized on every edit, especially because I'm not confident of our test coverage across datastores. 

But yes, it can be useful to better generalize on the parent class, but watch the complexity tradeoff.

eastandwestwind: Hmm yeah good point. I'm open to either, but wanted to make sure it was clear that this isn't necessarily SQLServer-specific, but instead a different (and more complex!) way to build queries such that the query generation step works for all connectors. I can make a new SQLServer-specific class, and just make a note that it can work across connectors, but more thorough testing will be needed, particularly once we have a better testing framework in place

src/fidesops/service/connectors/query_config.py
```diff
@@ -366,6 +366,69 @@ def dry_run_query(self) -> Optional[str]:
...
+
+    1. This increases code complexity for building queries
+    2. Lack of thorough testing framework to ensure all connectors will in fact work using this code
+    """
+
+    @staticmethod
+    def format_clause_for_query_in(field_name: str, query_data_names: List[str]) -> str:
+        """Returns clauses in a format they can be added into SQL queries."""
+        in_clause = ", ".join(query_data_names)
+        return f"{field_name} IN ({in_clause})"
```
seanpreston, line 382: Re-implementing https://github.com/ethyca/fidesops/pull/157/files#diff-8287ffddccc208436200313a2d60459cccd653f2fe1c303754fb11db4c090785R250-R256 would still be better for consistency — did we think about flattening `in_clause` outside of the method and passing it in? That way the interface could stay the same

src/fidesops/service/connectors/query_config.py
```diff
@@ -396,6 +397,72 @@ def dry_run_query(self) -> Optional[str]:
...
+
+        filtered_data = self.typed_filtered_values(input_data)
+
+        if filtered_data:
+            clauses = []
+            query_data: Dict[str, Tuple[Any, ...]] = {}
+            formatted_fields = self.format_fields_for_query(self.fields)
+            field_list = ",".join(formatted_fields)
+
+            for field_name, data in filtered_data.items():
```
pattisdr, line 430: Can you update the `field_name` variable to be `string_path`? It's a new change to the other `generate_query` methods to reflect that this can potentially be a dot separated name, representing a nested field.

eastandwestwind: ah yes!

pattisdr: ⭐

eastandwestwind: I've renamed other occurrences I saw in `query_config` as well

pattisdr: thank you! still lots to do here around nested objects - these queries aren't actually supported yet, so a lot of these places still subject to change over the next few weeks!

src/fidesops/service/connectors/query_config.py
```diff
@@ -396,6 +398,73 @@ def dry_run_query(self) -> Optional[str]:
...
+    Generates SQL valid for SQLServer. This way of building queries should also work for every other connector,
+    but SQLServer is separated due to increased code complexity for building queries
+    """
+
+    def format_clause_for_query(
+        self, string_path: str, operator: str, operand: str
+    ) -> str:
+        """Returns clauses in a format they can be added into SQL queries."""
+        if operator == "IN":
+            return f"{string_path} IN ({operand})"
```
seanpreston, line 412: Do we need a `:` before `{operand}` or will this already be there?

src/fidesops/service/connectors/query_config.py
```diff
@@ -410,12 +479,10 @@ def format_fields_for_query(
         return [f'"{field_path.levels[-1]}"' for field_path in field_paths]
 
     def format_clause_for_query(
-        self,
-        field_name: str,
-        operator: str,
+        self, string_path: str, operator: str, operand: str
```
seanpreston, line 482: Nit: could we leave trailing commas on the end of function args so that black will continue to put them each on newlines?

src/fidesops/service/connectors/sql_connector.py
```diff
@@ -340,8 +343,33 @@ def create_client(self) -> Engine:
         """Returns a SQLAlchemy Engine that can be used to interact with a MicrosoftSQLServer database"""
         config = MicrosoftSQLServerSchema(**self.configuration.secrets or {})
         uri = config.url or self.build_uri()
-        return create_engine(
-            uri,
-            hide_parameters=self.hide_parameters,
-            echo=not self.hide_parameters,
-        )
+        return create_engine(uri, hide_parameters=False, echo_pool="debug")
```
seanpreston, line 346: We probably want to set `hide_parameters` and `echo` back to their original values here?

src/fidesops/service/connectors/query_config.py
eastandwestwind: it will already be there

src/fidesops/service/connectors/sql_connector.py
eastandwestwind: good catch!

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -172,6 +173,57 @@ def test_create_and_process_access_request(
...
+    # Both pre-execution webhooks and both post-execution webhooks were called
+    assert trigger_webhook_mock.call_count == 4
+
+    for webhook in policy_pre_execution_webhooks:
+        webhook.delete(db=db)
+
+    for webhook in policy_post_execution_webhooks:
+        webhook.delete(db=db)
+
+    policy.delete(db=db)
```
seanpreston, line 221: If we delete fixtures outside of their teardown methods this can lead to errors depending on the order the tests run in, we should remove this. The same for those `webhook.delete()` lines above too

tests/task/traversal_data.py
```diff
@@ -241,7 +242,8 @@ def sample_traversal() -> Traversal:
         name="mysql", collections=[customers, addresses, users], connection_key="mysql"
     )
     postgres = Dataset(name="postgres", collections=[orders], connection_key="postgres")
+    mssql = Dataset(name="mssql", collections=[addresses], connection_key="mssql")
 
-    graph = DatasetGraph(mysql, postgres)
+    graph = DatasetGraph(mysql, postgres, mssql)
```
seanpreston, line 247: Do we need a follow-up to add `mongodb`, `redshift`, and `snowflake` here too? It looks like this has been overlooked

eastandwestwind: We already have `mongo` traversal tests (this specific line is associated with SQL traversals). But I don't see any reason we can't add `redshift` and `snowflake`, too. Unless there's something specific to external connectors that makes it more difficult to test?

tests/service/privacy_request/request_runner_service_test.py
eastandwestwind: Hmm we were doing this before, in `test_create_and_process_access_request`, unsure if @pattisdr has any more context into why we're manually deleting webhooks vs using fixtures? My fixture knowledge isn't strong here.

seanpreston: It's more relevant to `session` level fixtures, where if we delete them before they're due to be re-used they won't get created again between tests.

Screenshot 2022-01-20 at 13 20 40

In this instance we'd be able to avoid nested the delete inside `try / except` blocks, which is better because test data creation and deletion should happen consistently in the same place where possible

pattisdr: @eastandwestwind, it looks like in `test_create_and_process_access_request`, we're not deleting the privacy request for cleanup purposes, we're trying to verify a consequence of deleting the privacy request - in other words, the deletion was part of the test.  We're checking that even in the event a privacy request is removed, its execution logs still remain for record-keeping.  Generally, we should be relying on fixture teardown.  

Perhaps this means for `test_create_and_process_access_request` too, we should create a separate test that just creates the resources in the test and then deletes them, so we're not tearing down a fixture.

eastandwestwind: Got it, thanks, @seanpreston and @pattisdr for the clarification! 

For the new `test_create_and_process_access_request_mssql` test, I don't necessarily have to assert for logs after deleting privacy request, so I can just remove that bit (L:215-224).

tests/fixtures.py
```diff
@@ -1076,6 +1100,7 @@ def mysql_example_test_dataset_config(
     db: Session,
     example_datasets: List[Dict],
 ) -> Generator:
+    # todo: this references the incorrect dataset
```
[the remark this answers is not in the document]
eastandwestwind: Also, I plan on addressing some peculiarities in `test_graph_task.py`. I'm seeing things like:

`ConnectionConfig(key="mysql", connection_type=ConnectionType.postgres),`

and:

```
 TaskResources(
            EMPTY_REQUEST,
            Policy(),
            [
                ConnectionConfig(key="mysql", connection_type=ConnectionType.mongodb),
                ConnectionConfig(
                    key="postgres", connection_type=ConnectionType.mongodb
                ),
            ],
        ),
```

Let's add this to the same ticket

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -207,22 +207,9 @@ def test_create_and_process_access_request_mssql(
...
     assert trigger_webhook_mock.call_count == 4
 
-    for webhook in policy_pre_execution_webhooks:
-        webhook.delete(db=db)
-
-    for webhook in policy_post_execution_webhooks:
-        webhook.delete(db=db)
-
-    policy.delete(db=db)
-    pr.delete(db=db)
```
seanpreston: It's generally good practise to tidy up what's created as part of the test itself, which in this case `pr` is

eastandwestwind: Fixed! Back over to you @seanpreston

seanpreston (approved):
Thanks for all the edits @eastandwestwind!

Issue #129, opened 2021-12-14
[Array + JSON handling] Erasures in complex datatypes for MongoDB


seanpreston:
Add support for erasing data from complex datatypes (objects + arrays) in MongoDB.

pattisdr:
Note access requests addressed here https://github.com/ethyca/fidesops/pull/165 and some light testing of nested erasures completed. This ticket should make sure there are no other additional requirements we need to consider and complete more detailed testing.

pattisdr:
FYI Current mongo work that I'm doing for accessing data in complex arrays involves a design for erasures, this is being looked at currently.

PR #165, opened 2022-01-21
Add support for Nested Queries in Mongo [Access Requests] [#128]

### LINKED ISSUE #128: [Array + JSON handling] Query nested data from objects from Mongo
This increment describes the ability to:
- Discover data on a node via a nested identity. In other words, one collection has its identity data in a nested object

```
dataset:
  - fides_key: example
    name: Users
    collections:
      - name: user
        fields:
          - name: id
            data_categories: [system.operations]
            fidesops_meta:
              primary_key: True
          - name: username
            data_categories: [user.provided.identifiable]
          - name: email
            data_categories: [user.provided.identifiable]
            fidesops_meta:
              identity: email
          - name: comments
            fidesos_meta:
              data_type: object[]
            - fields:
              - name: id
                data_categories: [system.operations]
              - name: content
                data_categories: [user.provided]
              - name: email
                data_categories: [user.provided]
                fidesops_meta:
                  identity: email
```
— here if `joe@example.com` submits a privacy request, and that email is only found in objects within `user.comments`, then that data can be retrieved as an identity.

- Reference one node from another via a nested object field

```
dataset:
  - fides_key: example
    name: Users
    collections:
      - name: mailing_list
        fields:
          - name: id
            data_categories: [system.operations]
            fidesops_meta:
              primary_key: True
          - name: emails
            data_categories: [user.derived]
            fidesos_meta:
              data_type: object[]
              fields:
                - name: email
                  references:
                    - dataset: example
                      field: user.email
                      direction: to

      - name: user
        fields:
          - name: id
            data_categories: [system.operations]
            fidesops_meta:
              primary_key: True
          - name: email
            data_categories: [user.provided]
            fidesos_meta:
              data_type: string
```
— here `mailing_list.emails[].email` points to `user.email`

- Retrieve data with a data_category that applies to a nested object field, for example we want to get phone numbers, and mongo_example::customer_details::preferences.backups.phone field has a data category of [user.provided.identifiable.contact.phone_number]

```
dataset:
  - fides_key: example
    name: Users
    collections:
      - name: user
        fields:
          - name: id
            data_categories: [system.operations]
            fidesops_meta:
              primary_key: True
          - name: username
            data_categories: [user.provided.identifiable]
          - name: email
            data_categories: [user.provided.identifiable]
            fidesops_meta:
              identity: email
          - name: comments
            fidesos_meta:
              data_type: object[]
            - fields:
              - name: id
                data_categories: [system.operations]
              - name: content
                data_categories: [user.provided]
              - name: email
                data_categories: [user.provided]
                fidesops_meta:
                  identity: email
```
— here any query for `user` data will retrieve `comment.content` and `comment.email`

Other requirements:
- [ ] Add logic to the Dataset validation to ensure fields with `data_type: object` cannot have `data_categories: [...]`

NB. Erasure via nested object fields will happen as part of a subsequent increment
NB. This increment is Mongo specific for now, and can be extended to other data stores at a future point

pattisdr:
# Purpose

Finish adding support for nested objects in MongoDB, building off of Steven's work in https://github.com/ethyca/fidesops/pull/154 to add `FieldPath` objects to represent nested fields.  Note, this adds support for nested *access* requests.  Some light testing and adjustments were made for erasures on nested objects, but more comprehensive testing needed there.

# Changes
- [x] **Fix querying data on a nested object** - this wasn't quite working yet with FieldPaths from the last PR, so we're fixing that here (queries were being built whose projection queried for every nested level, causing collisions). Now we query on top-level fields.
  - Specifically, we alter the second value in the tuple returned by`MongoQueryConfig.generate_query`, which specifies which fields to return.
  - [x] Related: **Allow filtering data categories on nested results.** (`filter_data_categories`) We over-select when querying data, and filter to just the data categories specified on the Policy before uploading to the user.  Previously, a category on nested data would be overlooked. Now if a query returns an object, and we only need one field in that object, we will strip off the other fields so the object contains only the field matching the data category.
- [x] **Allow one collection to reference another collection via a nested object** in Mongo.  For example, you could take a nested id from collection and use that to locate a record in another collection with this id that is nested inside another field.
  - This was a change to `GraphTask.to_dask_input_data` to retrieve nested data from one collection and pass it as flattened data into the next collection.
- [x] **Allow identity data to be specified on a nested object.**  For example, using an email as seed data that is buried in a nested object is supported.
  - This already worked after FieldPaths, so supporting tests only were added.
- [x] Add Fidesops validation to **disallow specifying data categories on object fields** - because an object's data category and its nested data catagories could conflict, we're choosing to make data categories be specified at the lowest level.
  - Added validation at both the `FidesopsDatasetField` and `ObjectField` level (latter redundant, but good for testing to make sure we're testing the right things when declaring ObjectFields directly)

## Also:
- Expand mongo dataset in postman collection for local end-to-end experimentation on nested objects

# Checklist

# Ticket

Fixes #128


src/fidesops/task/graph_task.py
```diff
@@ -410,16 +410,17 @@ def filter_data_categories(
...
-            )
+        # Normalize nested data into a flat dataframe
+        df = pd.json_normalize(results, sep=".")
+        # Only keep dataframe columns that match the target field paths
+        df = df[[field_path.string_path for field_path in target_field_paths]]
+        # Turn the filtered results back into a list of dictionaries
+        filtered_flattened_results = df.to_dict(orient="records")
+        for row in filtered_flattened_results:
+            # For each row, unflatten the dictionary
+            filtered_access_results[node_address].append(unflatten_dict(row))
```
pattisdr, line 424: Using pandas `json_normalize` which we also use when we flatten data for CSV upload - wanted to get more use out of this tool since we include it as a dependency!

Unfortunately, I couldn't find a method that does the opposite of `json_normalize` and rebuilds the data in the nested format we want to return to the user, so I added an `unflatten_dict` util.

src/fidesops/service/connectors/query_config.py
```diff
@@ -178,7 +182,7 @@ def update_value_map(
                         f"Unable to generate a query for field {rule_field_path.string_path}: data_type is either not present on the field or not supported for the {strategy_config['strategy']} masking strategy. Received data type: {masking_override.data_type_converter.name}"
                     )
                     continue
-                val: Any = row[rule_field_path.string_path]
+                val: Any = rule_field_path.retrieve_from(row)
```
pattisdr, line 185: This adjustment lets us get a potentially nested value that was previously retrieved so we can mask it.

src/fidesops/service/connectors/query_config.py
```diff
@@ -320,9 +324,6 @@ def generate_query(
                         self.format_clause_for_query(string_path, "IN", string_path)
                     )
                     query_data[string_path] = tuple(data)
-                else:
-                    #  if there's no data, create no clause
-                    pass
```
pattisdr, line 325: Removed as it's not necessary.

src/fidesops/service/connectors/query_config.py
```diff
@@ -553,13 +554,8 @@ def transform_query_pairs(pairs: Dict[str, Any]) -> Dict[str, Any]:
         if input_data:
             filtered_data: Dict[str, Any] = self.typed_filtered_values(input_data)
             if filtered_data:
-                field_list = {
-                    field_path.string_path: 1
-                    for field_path, field in self.field_map().items()
-                }
```
pattisdr, line 559: The `field_list` portion here is used to generate the projection of the mongo query which specifies which fields we want from mongo.  `self.field_map` flattens every field and every level of a nested field.

The projection being generated was something like this: notice that `preferences.notifications` and `preferences.notifications.security_emails` are both included. This would cause mongo to throw errors about path collisions. 

```
db.getCollection('customer_details').find({"customer_id": 1}, {'_id': 1, 'preferences.notifications.security_emails': 1, 'preferences.notifications.informational_emails': 1, 'preferences.notifications': 1})
```

Below, I generate the `field_list` using top-level fields instead.  I'm also moving the definition of `field_list` closer to where it's being used.

src/fidesops/task/graph_task.py
```diff
@@ -152,11 +160,7 @@ def to_dask_input_data(self, *data: List[Row]) -> Dict[str, List[Any]]:
 
             for row in rowset:
                 for foreign_field_path, local_field_path in field_mappings:
-                    new_value = (
-                        row[foreign_field_path.string_path]
-                        if foreign_field_path.string_path in row
-                        else None
-                    )
+                    new_value = foreign_field_path.retrieve_from(row)
```
pattisdr, line 163: Data is returned from mongo in a potentially nested format.  This fix lets us retrieve potentially nested data before passing as flattened format into a downstream collection that needs potentially nested data as input.

Docstring updated.

src/fidesops/task/graph_task.py
```diff
@@ -410,16 +410,17 @@ def filter_data_categories(
...
-        for row in results:
-            filtered_access_results[node_address].append(
-                {
-                    field: result
-                    for field, result in row.items()
-                    if field in {target.string_path for target in target_fields}
-                }
-            )
+        # Normalize nested data into a flat dataframe
+        df = pd.json_normalize(results, sep=".")
```
eastandwestwind, line 417: ahh dataframes takes me back to my R days. Could we include the `DataFrame` return type for clarity here?

src/fidesops/util/nested_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def _create_dict() -> defaultdict:
+        """Can create a defaultdict at every level"""
+        return defaultdict(_create_dict)
+
+    unpacked_results = _create_dict()
+
+    for key, value in input_row.items():
+        levels = key.split(separator)
+        subdict = unpacked_results
```
eastandwestwind, line 29: Woah this is a super elegant way of doing this!

src/fidesops/service/connectors/query_config.py
pattisdr: This is just a Python cleanup, if you are specifying if/elif/elif, and your "else" case is "do nothing", you don't need it. 

```python3
def animal_sounds(animal:str):
    if animal == "dog":
        print("woof")
    elif animal == "cat":
        print("meow")
```

```bash
>>> animal_sounds("dog")
woof
>>> animal_sounds("cat")
meow
>>> animal_sounds("toad")
```

Thanks for pointing out this is elsewhere, I'll clean it up.

src/fidesops/task/graph_task.py
```diff
@@ -377,7 +381,7 @@ def termination_fn(*dependent_values: int) -> Tuple[int, ...]:
 def filter_data_categories(
     access_request_results: Dict[str, Optional[Any]],
     target_categories: Set[str],
-    graph: DatasetGraph,
+    data_category_fields: Dict[str, Dict[str, List[FieldPath]]],
```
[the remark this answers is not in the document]
pattisdr: good point, maybe there's a compromise here - I don't really want to go overboard with a pydantic schema, this is just an intermediate format that has collected which Fields are associated with which data categories for each collection.

Maybe if I refactor so the keys are `CollectionAddresses`, and I am more specific about the inner Dict, something like:

```
data_category_fields: Dict[CollectionAddress, Dict[FidesOpsKey, List[FieldPath]]]
```

src/fidesops/task/graph_task.py
pattisdr: helpful thank you

src/fidesops/util/nested_utils.py
pattisdr: thank you!!

src/fidesops/service/connectors/query_config.py
eastandwestwind: oh right, good point 😁

src/fidesops/task/graph_task.py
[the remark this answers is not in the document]
eastandwestwind: oo I'd like that. I think my main issue was not immediately being about to trace what those `str` types were referring to.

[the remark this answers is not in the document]
eastandwestwind: if it's not too large of a refactor, of course, otherwise feel free to make another ticket for this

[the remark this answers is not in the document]
pattisdr: it was a quick change, thanks @eastandwestwind!

pattisdr:
Thanks for the comments @eastandwestwind !

eastandwestwind:
@pattisdr is there a follow-up ticket for more comprehensive erasure testing?

pattisdr:
@eastandwestwind good question, yes, this ticket was just supposed to be access requests, and Sean created a separate ticket for erasures https://github.com/ethyca/fidesops/issues/129.  However, a failing test necessitated that erasures be addressed at least in part here, so tests have been added confirming a nested erasure was successful.  On the follow-up ticket we should verify that there aren't any other requirements we've missed and add more detailed tests.

pattisdr:
For posterity, issue described above was unrelated - sending a request to `/privacy-request/preview/` to build the queries for one dataset, that references another dataset, needs both datasets listed in the request body.  This is non-obvious, and throws a 500 when encountered. Small tweak added to raise a 400 and explain how to improve request.

Issue created to add guides for this endpoint because it's really useful and I hadn't used it until today.

docs/fidesops/docs/development/overview.md
```diff
@@ -32,7 +32,7 @@ commands to give you different functionality.
 - `make pytest-integration-access` - runs access integration tests.
 - `make pytest-integration-erasure` - runs erasure integration tests.
 - `make reset-db` - tears down the Fideops postgres db, then recreates and re-runs migrations.
-- `make quickstart` - runs a quick, five second quickstart that talks to the Fidesops API to execute privacy requests
+- `make quickstart` - runs a quick, five minute quickstart that talks to the Fidesops API to execute privacy requests
```
eastandwestwind, line 35: 😄

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -384,9 +389,15 @@ def get_request_preview_queries(
...
-            *[dataset.get_graph() for dataset in dataset_configs]
-        )
+        try:
+            dataset_graph: DatasetGraph = DatasetGraph(
+                *[dataset.get_graph() for dataset in dataset_configs]
+            )
+        except ValidationError as exc:
+            raise HTTPException(
+                status_code=HTTP_400_BAD_REQUEST,
+                detail=f"{exc}. Make sure all referenced datasets are included in the request body.",
```
eastandwestwind, line 399: thank you for this detailed error message!

eastandwestwind (approved):
Great work on this @pattisdr ! I love the thorough testing you've put in place here, in addition to the last-minute add to ensure the `/privacy-request/preview/` endpoint returns clearer error messages.

src/fidesops/graph/config.py
```diff
@@ -187,6 +188,19 @@ def parse(path_str: str) -> FieldPath:
...
+        Example:
+            input_data = {"A": {"B": {"C": 2, "D": 3}}}
+            field_path = FieldPath("A", "B", "C")
+            field_path.retrieve_from(input_data) = 2
+        """
+        sub_data = copy.deepcopy(input_data)
+        for level in self.levels:
+            sub_data = sub_data[level]
+        return sub_data
+
```
stevenbenjamin, line 209: This is vulnerable to missing levels - in the documented case `retrieve_from({"a":{"b":{}} ` will raise an error. There's also a pydash method that handles this, see [here](https://pydash.readthedocs.io/en/latest/api.html?highlight=get_#pydash.objects.get)

pattisdr: Good point @stevenbenjamin, there's a similar vulnerability in `filter_data_categories`, this pydash method looks great.

src/fidesops/graph/config.py
```diff
@@ -294,6 +314,21 @@ class ObjectField(Field):
...
+        cls: "ObjectField", value: Optional[List[FidesOpsKey]]
+    ) -> Optional[List[FidesOpsKey]]:
+        """To prevent mismatches between data categories on an ObjectField and a nested ScalarField, only
+        allow data categories to be defined on the individual fields.
+
+        This shouldn't be hit unless an ObjectField is declared directly.
+        """
+        if value:
+            raise ValueError(
+                "ObjectFields cannot be given data_categories; annotate the sub-fields instead."
```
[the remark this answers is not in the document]
pattisdr: ⭐

src/fidesops/util/nested_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        for level in levels[:-1]:
+            subdict = subdict[level]
+
+        try:
+            subdict[levels[-1]] = value
```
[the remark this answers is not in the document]
pattisdr: Thanks for the comment!

I think an IndexError is unlikely, as accessing the `-1` index always returns the last element in the list.  This could fail if the list `levels` was empty.  But levels is formed when we split the dictionary key on the separator - even something like ` "".split('.')` returns a list with one element, an empty string.

The only example I've thought of is `unflatten_dict({"": "hi"}, separator=None)`. We do specify in the signature that separator shouldn't be None, and the key of that dict is an empty string.

So maybe we leave this IndexError in, I think if this occurred, unflatten_dict would be used in a way it wasn't intended. I could be missing something, too. 

At any rate, I've added more tests to the unflatten_dict examples, and added a condition requiring that this is only run on completely flattened dicts, not partially flattened dicts to avoid surprising results.

PR #168, opened 2022-01-24
136 mysql query execution

### LINKED ISSUE #136: Query Execution for MySQL
This was never formally addressed.

Add integration access and erasure tests for MySQL.  

We have the connector logic, but no tests for the traversal portion.  Currently MySQL is just using base SQL Connector logic, but we may be missing some core MySQL differences, because we haven't formally assessed this datastore unlike postgres, snowflake, etc.

eastandwestwind:
# Purpose
Adds query execution support for MySQL Connector.

# Changes
- Adds integration access and erasure tests for MySQL, and 1 [missing one](https://github.com/ethyca/fidesops/pull/168/files#diff-e6aeca78065db09a4852ebc1c7a584c6c1839c5dbe002e1ca90fe96dd8d18c78R338) for MsSQL
- Overrides SQLConnector where appropriate to handle `LegacyCursorResult` return type from SQLAlchemy.
- Updates Postman Collection and Postman Docs to reflect MySQL-specific config

# Checklist

# Ticket

Fixes #136


src/fidesops/service/connectors/sql_connector.py
```diff
@@ -183,6 +183,32 @@ def create_client(self) -> Engine:
...
+    # Overrides BaseConnector.cursor_result_to_rows
+    @staticmethod
+    def cursor_result_to_rows(results: LegacyCursorResult) -> List[Row]:
+        """Convert SQLAlchemy results to a list of dictionaries"""
+        columns: List[Column] = results.cursor.description
+        l = len(columns)
+        rows = []
+        for row_tuple in results:
+            rows.append({columns[i][0]: row_tuple[i] for i in range(l)})
+        return rows
```
pattisdr, line 195: I see this is elsewhere, but Python's `range` creates a sequence of numbers that we're iterating over, but we already have the columns themselves that we can use to iterate over, so it feels a bit roundabout.

 I might rewrite as:
```python
rows.append({col[0]: row_tuple[count] for count, col in enumerate(columns)})
```
where `enumerate` gives you both the column value whose loop we're on, as well as the current count.

src/fidesops/service/connectors/sql_connector.py
```diff
@@ -183,6 +183,32 @@ def create_client(self) -> Engine:
...
+        query_config = self.query_config(node)
+        client = self.client()
+        stmt: Optional[TextClause] = query_config.generate_query(input_data, policy)
+        if stmt is None:
+            return []
+        logger.info(f"Starting data retrieval for {node.address}")
+        with client.connect() as connection:
+            # fixme: update mssql type too to LegacyCursorResult
+            results: LegacyCursorResult = connection.execute(stmt)
+            return MySQLConnector.cursor_result_to_rows(results)
```
pattisdr, line 210: Unless I'm missing something, this override seems to just call `MySQL.cursor_result_to_rows` instead of `SQLConnector.cursor_result_to_rows`. However, cursor_result_to_rows is a staticmethod and we can call that from the instance itself.

Instead of the overrides you've added here for MySQL and MSSQL `retrieve_data`, you should be able to get rid of it, and have the parent `SQLConnector.retrieve data` return `self.cursor_result_to_rows`.  When `MySQL.retrieve_data` is run, `self` will be an instance of the MySQL Connector, so the subclassed `cursor_result_to_rows` should be selected.

tests/integration_tests/test_sql_task.py
```diff
@@ -333,6 +333,89 @@ def test_sql_access_request_task(db, policy, integration_postgres_config) -> Non
         > 0
     )
 
+# todo- add mssql test
```
pattisdr, line 336: is this a follow-up TODO? does it need a ticket?

tests/integration_tests/test_sql_task.py
```diff
@@ -253,10 +253,10 @@ def test_sql_erasure_task(db, postgres_inserts, integration_postgres_config):
 
 
 @pytest.mark.integration
-def test_sql_access_request_task(db, policy, integration_postgres_config) -> None:
+def test_postgres_access_request_task(db, policy, integration_postgres_config) -> None:
 
     privacy_request = PrivacyRequest(
-        id=f"test_sql_access_request_task_{random.randint(0, 1000)}"
+        id=f"test_postgres_access_request_task_{random.randint(0, 1000)}"
```
pattisdr, line 259: Nice clarity here -

tests/integration_tests/test_sql_task.py
```diff
@@ -333,6 +333,89 @@ def test_sql_access_request_task(db, policy, integration_postgres_config) -> Non
...
+    )
+
+    v = graph_task.run_access_request(
+        privacy_request,
+        policy,
+        integration_db_graph("my_mysql_db_1"),
+        [connection_config_mysql],
+        {"email": "customer-1@example.com"},
+    )
+
```
pattisdr, line 352: Would you mind adding a separate postman folder with a MySQL collection for local end-to-end testing? Ideally it would create a connection, add the secrets, and create a MySQL dataset. I moved the MSSQL calls into their own folder, so the folder with the "minimum calls to create an access request" doesn't get too bloated.

pattisdr (commented):
thank you @eastandwestwind, well-organized, easy to follow, just a few changes requested ⭐

src/fidesops/service/connectors/sql_connector.py
eastandwestwind: niiice, learning lots of new python things today!

tests/integration_tests/test_sql_task.py
eastandwestwind: I'll go ahead and add this now!

tests/integration_tests/test_sql_task.py
eastandwestwind: Definitely, good call!

eastandwestwind:
Thanks for the comments @pattisdr ! Ready for review again

pattisdr (approved):
Wonderful @eastandwestwind!

More of a general comment, related to Sean's testing work, it will be good when we are testing each datastore against every data type; we may be missing datastore-specific handling in places.  But this PR definitely demonstrates basic MySQL access/erasures in line with how we introduced the other datastores and we can continue to improve our test matrix. Thanks again for adding MySQL to the postman collection!

PR #185, opened 2022-01-27
Consolidate `erasure` and `access` tests into `integration` tests

seanpreston:
This PR:
- [x] Adds `/fixtures` as a top level dir for more specific fixture files, e.g. `/fixtures/fixtures_postgres.py`
- [x] Moves schema population SQL code into pytest fixtures
- [x] Removes the `integration_erasure` pytest markers
- [x] Adds appropriate `integration_{env}` pytest markets
- [x] Cleans up overlap in integration env fixtures
- [ ] Adds data seeding scripts for all datastores so we can still have data when we run `make integration-env`


eastandwestwind (commented):
Just 1 cleanup task. Nice work @seanpreston !

tests/integration_tests/test_integration_postgres_example.py
```diff
@@ -9,26 +13,26 @@
...
-    engine = get_db_engine(database_uri=example_postgres_uri)
-    logger.debug(f"Connecting to PostgreSQL example database at: {engine.url}")
-    SessionLocal = get_db_session(engine=engine)
-    the_session = SessionLocal()
-    # Setup above...
-    yield the_session
-    # Teardown below...
-    the_session.close()
-    engine.dispose()
+# @pytest.fixture(scope="session")
```
eastandwestwind, line 16: Can we remove this if not needed?

tests/integration_tests/test_integration_postgres_example.py
```diff
@@ -49,4 +53,9 @@ def test_postgres_example_data(postgres_example_db):
         # templating as much as possible. instead, use the table() helper to
         # dynamically generate the FROM clause for each table_name
         count_sql = select(func.count()).select_from(table(table_name))
-        assert postgres_example_db.execute(count_sql).scalar() == expected_count
+        actual_count = postgres_integration_db.execute(count_sql).scalar()
+        # if actual_count != expected_count:
```
eastandwestwind, line 57: another leftover bit from testing

docker-compose.integration-postgres.yml
```diff
@@ -13,7 +23,3 @@ services:
       - 6432
     ports:
       - "0.0.0.0:6432:5432"
-    volumes:
-      # NOTE: consider running initialization SQL in Python instead of this
-      # (see "Initialization scripts" section at https://hub.docker.com/_/postgres)
-      - ./data/sql/postgres_example.sql:/docker-entrypoint-initdb.d/postgres_example.sql
```
eastandwestwind, line 19: Does this break our postman flow, since the SQL required to create / populate the db is only loaded in with a test fixture? I haven't tested, but when we make integration-env, can we run a successful access request against Postgres?

eastandwestwind (commented):
Lots of good work on this @seanpreston ! I like that now we can specify specific datastores to run with erasures.

Did a 1st pass just going through the code, and have some Qs.

data/sql/postgres_example.sql
```diff
@@ -1,18 +1,4 @@
 -- Example Postgres schema matching the dataset in public/data/dataset/postgres_example_dataset.yml
-DROP TABLE IF EXISTS public.report;
```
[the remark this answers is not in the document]
seanpreston: We are dropping tables before creating them, we're just doing it a different way now.

Postgres doesn't like `DROP DATABASE ...` in the context of transactions. This doesn't play nicely with our SQLAlchemy sessions. This isn't an issue anywhere else in the codebase because well, we'd never want to execute a statement like that. In this case I've moved the drop DB logic into the fixture itself, this is predominantly to get around the issue where `TRUNCATE ...` statements were locking the DB indefinitely in the tests.

docker-compose.integration-postgres.yml
```diff
@@ -4,6 +4,16 @@ services:
       - postgres_example
 
   postgres_example:
+    command:
+      [
+        "postgres",
+        "-c",
+        "logging_collector=on",
```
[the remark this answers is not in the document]
seanpreston: I see no harm in leaving these on for the utility. This is the integration test DB so gets trashed repeatedly so we don't need to worry about logfile sizes maxing out.

docker-compose.integration-postgres.yml
seanpreston: Great spot, this does. I've added this TODO to rectify that before we merge:
Screenshot 2022-02-18 at 14 30 49

tests/integration_tests/setup_scripts/postgres_setup.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    session = SessionLocal()
+
+    if database_exists(session.bind.url):
+        # Postgres cannot drop databases from within a transaction block, so
+        # we should drop the DB this way instead
+        drop_database(session.bind.url)
+    create_database(session.bind.url)
+
+    with open("./data/sql/postgres_example.sql", "r") as query_file:
```
eastandwestwind, line 63: 🙌

eastandwestwind (approved):
Looking good, thanks for fixing the setup for postgres/mongo so that our postman testing still works!

PR #188, opened 2022-02-03
Correctly format and use MariaDB fixtures

seanpreston:
# Purpose

This PR uses the MariaDB fixtures as part of the tests, and fixes some SQL syntax errors in their execution


eastandwestwind (approved):
looks good, thanks for fixing this up!

PR #189, opened 2022-02-04
76 Connection config BigQuery

### LINKED ISSUE #76: [DB Compatibility] `ConnectionConfig` for BigQuery
Fidesops must support connecting to a BigQuery instance. Test credentials can be obtained from @iamkelllly 

This issue represents just the connection, query building for BigQuery will come in a subsequent issue.

eastandwestwind:
# Purpose
Adds connection support for BigQuery

# Changes
- Adds connector for BigQuery
- Adds db migration for new `bigquery` enum
- Adds integration tests- as part of this, adds helper function to populate our test BigQuery schema with data
- Updates postman collection
- Updates docs to include example for `PATCH {{host}}/connection/{{bigquery_key}}/secret`

# TODO- add github secrets for CI

# Checklist

# Ticket

Fixes #76


tests/integration_tests/test_external_database_connections.py
```diff
@@ -106,3 +106,22 @@ def test_snowflake_example_data(snowflake_test_engine):
         "service_request",
         "visit",
     ]
+
+@pytest.mark.integration_external
+@pytest.mark.integration_bigquery
```
galvana, line 111: Are there any other tests that we should mark with the `integration_bigquery` annotation?

tests/fixtures/bigquery_fixtures.py
```diff
@@ -0,0 +1,304 @@
+import logging
```
galvana, line 1: I like that the fixtures for BigQuery are separated from the rest. It makes it easier to work with 👍

tests/fixtures/bigquery_fixtures.py
```diff
@@ -0,0 +1,304 @@
...
+
+    connector: BigQueryConnector = get_connector(connection_config)
+    engine = connector.client()
+    yield engine
+    engine.dispose()
+
+
+def seed_bigquery_integration_db(bigquery_integration_engine) -> None:
+    """
+    Currently unused.
```
galvana, line 97: I see this is currently unused but does it make sense to move these SQL statements under `/data/sql` in the future? Is that the directory we use for seed data?

eastandwestwind: I'm open to it! Some caveats:

* Those SQL statement files in `/data/sql` are typically referenced by the corresponding `docker-compose.integration-[datastore].yml` files
* If we wanted to run this seed SQL, it will be more complex to load in and run the SQL via python, as opposed to if it were in a python method. Alternatively we could copy/paste the SQL directly in bigquery console, but only if the user has access to the console. I don't anticipating having to run this code often, but it's good to have an easy way to run it just in case.
* we do have other seed data in the corresponding `[datastore]-fixtures,py` files, so this pattern follows convention

tests/integration_tests/test_external_database_connections.py
eastandwestwind: Not in this PR- in the BigQuery query generation there are more.

notice that https://github.com/ethyca/fidesops/pull/189/files#diff-0fa654d3dbcc4abd3ae1889caf2316f7efa649d6bb7ec77abe6261fd16d35d6bR689 doesn't actually connect to a bigquery instance, so we won't need to mark it as an integration test

tests/fixtures/bigquery_fixtures.py
galvana: Thanks for the explanation, it makes sense to me now why it's here. Let's not overcomplicate things then and keep this where it is.

galvana (approved):

Issue #193, opened 2022-02-08
[Array + JSON Handling] Make Element Matching Configurable for Arrays


pattisdr:
### Is your feature request related to a specific problem?

This is being split off of https://github.com/ethyca/fidesops/issues/146, as that PR is getting large.  This also might be worth postponing addressing until we do erasures on arrays.

In array use case discussions, it was brought up that the user might want to configure some array behavior.  Specifically, introducing array support adds some ambiguity into how queries should be built.  A query on an array field (+ field is a `reference` field or contains identity data) may match multiple indices in an array, or multiple subdocuments inside the array.   The default behavior I've built into #146  is that only matched array indices are returned on a document, and only those matched values are used to make subsequent queries.   This is a more conservative approach, in that it selects less/masks less, but may not cover all user's data structures.

(This is separate from the after-the-fact filtering we do on data categories.  This is related to what pieces of array data we return and/or use for other queries).

### Describe the solution you'd like

Allow the user to configure in the dataset yaml that all values in the discovered array field should be returned and/or used to make subsequent queries, if they don't want the default.  Specifically there's a `filter_element_match` method that we run (we do this in code, not at the database level) that removes non matched elements from arrays, but we could be more targeted as to which fields we run this on.

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

pattisdr:
For example, say for the "people" collection we run a query that says "return records where users.id = 1".   
![untitled (1)](https://user-images.githubusercontent.com/9755598/153283374-2b71533d-d20e-4a06-ae27-3f45a7cba58f.png)
That query returns the first record, {"users": ["id": 1, "ccn": 12345, "name": "Dawn"}, {"id": 2, "ccn": 54321", "name": "Jane"}], "employees": []} because it contains the id:1  somewhere in there.   Say we're supposed to use the ccn here to lookup records in the credit card collection.  Because this is an array, filter out the non-matched elements: `{"users": ["id": 1, "ccn": 12345, "name": "Dawn"}], "employees": []} `, save that as the result for the people collection, and then pass this data to the credit card collection to only search for credit cards where the id is 12345.

iamkelllly:
> Specifically there's a `filter_element_match` method that we run (we do this in code, not at the database level) that removes non matched elements from arrays, but we could be more targeted as to which fields we run this on.

[...]

PR #194, opened 2022-02-08
Add support for Array Access Requests in MongoDB [#146]

### LINKED ISSUE #146: [Array + JSON Handling] Query nested data from arrays from Mongo
When given a Dataset specification containing an array:

```
          - name: tags
            fidesops_meta:
                data_type: string[]
            data_categories: [user.provided]
```

we have a few separate use cases:
- `tags` can contain a root identity
- `tags` cannot contain a root identity

Fidesops must:
- Treat each element within the array as each data category within `data_categories` for access and retrieval purposes — e.g.
```
["dawn@ethyca.com", "sean@ethyca.com", "steven@ethyca.com"]
```
must mask all values if the field is not an identity, resulting in:
```
["masked", "masked", "masked"]
```

- Be able to mask and retrieve an identity from within an array — e.g.
```
["dawn@ethyca.com", "sean@ethyca.com", "steven@ethyca.com"]
```
must be able to mask only `steven@ethyca.com` if the field is an identity, resulting in:
```
["dawn@ethyca.com", "sean@ethyca.com", "masked"]
```

NB. Here I'm using the word `masked` to denote the result of the policy's masking strategy applied to the value.

Questions:
- Is it feasible to support nested arrays at this stage? — e.g.
```
[
  ["dawn@ethyca.com", "sean@ethyca.com"],
  ["steven@ethyca.com"]
]
```
must result in:
```
[
  ["dawn@ethyca.com", "sean@ethyca.com"],
  ["masked"]
]
```
if the field is an identity and `steven@ethyca.com` is passed as the value, and
```
[
  ["masked", "masked"],
  ["masked"]
]
```
if not.

I understand that support will also need to be added at the Dataset level here too (https://github.com/ethyca/fidesops/blob/0157344397ae4ebca56a8b8169e362b2b17ae198/tests/util/test_dataset_yaml.py#L36)

pattisdr:
👉   A lot of the diff are new tests/rearranged tests/new dataset annotations/docs etc.

# Purpose

Adds array support for access requests in MongoDB. Note that erasures on arrays are **definitely not addressed** here and will be worked on in #129.  You should not attempt to run erasures on collections with arrays yet.

# Changes

Three primary pieces are involved in adding better array support - 1) Treating arrays as embedded collections - we only return data that is matched in arrays (there is interest in making this configurable https://github.com/ethyca/fidesops/issues/193)  . 2) Consolidating array results from one collection to feed into a query on downstream collections. Also remember arrays of objects are a thing! 3) Filtering access request results for data categories on array data. More information below:

- Expands example mongo dataset and initialized data in MongoDB to include more complex object and array examples
- On `graph_task.access_request`, after we've run an access request on a collection, I use a new `filter_element_match` method to only keep matched array indices and matched objects embedded in arrays where applicable. I then save this modified data in the cache, and then pass it onto the next collection for use in building more queries downstream.  I'm doing this in code so it will be more broadly applicable to other datastores besides MongoDB, plus this would be expensive query to run db-side. See guides for more info why.
- In `to_dask_input_data` (which takes inputs from potentially multiple collections and formats to pass into the current collection), run `consolidate_query_matches` to combine data from potentially multiple scalar values, arrays, or scalar values with arrays of embedded objects,  into a flat list.  The next collection will run a query to see if the input fields match any of the consolidated data.
-  Updates `filter_data_categories` (which selects relevant data categories from each record to return to the user) to be able to select matching data from arrays.  We were using `json_normalize` after we added nested objects, but now we need something a little more custom.  I'm using a new method `select_and_save_field` to recurse through both objects and arrays to get the targeted data.
- Some recursive functions added to support the above, since we're retrieving data from unknown levels.
- Lots of tests, organization of tests, tests of the individual recursion pieces. `tests/integration_tests/test_mongo_task.py:: test_array_querying_mongo`  has the most comprehensive end-to-end test checking different array use cases.

# Checklist

- [X] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)
- [X] Good unit test/integration test coverage

# Ticket

Fixes #146


src/fidesops/task/graph_task.py
```diff
@@ -226,9 +233,23 @@ def log_end(
...
+            logger.info(
+                f"Filtering row in {self.traversal_node.node.address} for matching array elements."
+            )
+            filter_element_match(
+                row,
+                {
+                    FieldPath.parse(field): inputs
+                    for field, inputs in formatted_input_data.items()
+                },
+            )
```
pattisdr, line 252: This is a primary change: I filter out non-matching sub-documents and array elements in code, this is also a function that we may make configurable https://github.com/ethyca/fidesops/issues/193, although I think the default behavior makes sense, in that it views array data as nested documents, and only matched nested documents are used to find subsequent documents.

src/fidesops/task/graph_task.py
```diff
@@ -417,18 +445,11 @@ def filter_data_categories(
...
-        )
-        for row in filtered_flattened_results:
-            # For each row, unflatten the dictionary
-            filtered_access_results[node_address].append(unflatten_dict(row))
+        for row in results:
+            filtered_results: Dict[str, Any] = {}
+            for field_path in target_field_paths:
+                select_and_save_field(filtered_results, row, field_path)
+            remove_empty_objects(filtered_results)
+            filtered_access_results[node_address].append(filtered_results)
```
pattisdr, line 453: This is another primary change, `json_normalize` doesn't do what we need for array data, so I added a recursive `select_and_save_field` method that incrementally builds a new result with just the matched data categories on scalar, dictionary fields, array fields, or arrays of objects.

src/fidesops/task/graph_task.py
```diff
@@ -160,11 +162,16 @@ def to_dask_input_data(self, *data: List[Row]) -> Dict[str, List[Any]]:
...
             for row in rowset:
                 for foreign_field_path, local_field_path in field_mappings:
-                    new_value = foreign_field_path.retrieve_from(row)
-                    if new_value:
-                        append(output, local_field_path.string_path, new_value)
+                    new_values: List = consolidate_query_matches(
+                        row=row, target_path=foreign_field_path
+                    )
+                    if new_values:
+                        append(output, local_field_path.string_path, new_values)
```
pattisdr, line 174: This is the final primary change. Before, we were using `foreign_field_path.retrieve_from(row)` to get the data from the record and pass into the next collection. However, array data doesn't necessarily have a direct path, and array data can have *multiple* matched paths. So we get the multiple matches, and consolidate them into a single list.

src/fidesops/task/refine_target_path.py
```diff
@@ -0,0 +1,147 @@
...
+    was just one path, we return one list.
+
+    :Example:
+    In this example, path ["A", "B", "C"] points to two elements that match values "F" or "G".  We update
+    the path to insert the indices to locate the appropriate value.
+
+    refine_target_path({"A": {"B": [{"C": "D"}, {"C": "F"}, {"C": "G"}]}}, ["A", "B", "C"], only=["F", "G"])
+
+    [["A", "B", 1, "C"], ["A", "B", 2, "C"]]
+    """
```
pattisdr, line 70: This method (which is a recursive method called as part of `filter_element_match`) should also be useful in making targeted edits for erasures when i get there.  I can take the field_path, build a more refined field path to specific element(s) in the record, mask just those specific values, and write the whole record back. In mongo, we can also write queries to update specific elements in arrays, but other datasets may not have this capability - in general I've added several Python methods to do some of this in code rather than at the database level.

pattisdr: I use a lot of recursion in this PR to crawl through arrays and dictionaries of unknown depth - I generally try to avoid recursive solutions in Python - Python doesn't have tail call recursion, there's generally a way to implement it with an iterative solution, recursion is limited to 1000 function calls by default, and the language has a strong preference for clear, readable solutions.   

To help with the readable part, I've tried to add a lot of detail to the docstrings with examples.  There may be ways to simplify the recursion, but a lot of these pieces were added to cover different array use cases.

src/fidesops/task/refine_target_path.py
```diff
@@ -0,0 +1,147 @@
...
+                results.extend(current_result)
+            else:
+                results.append(current_result)
+
+    return results[0] if len(results) == 1 else results
+
+
+def _match_found(elem: Any, only: List[Any]) -> bool:
+    """The given scalar element matches one of the input values"""
+    return elem in only
```
pattisdr, line 125: A corner case surfaced in testing today related to type coercion, and the cons of doing this filtering in code instead of at the database level (although I still think there are a lot of good reasons to do it in code).

Say the query was `select * from customer where id = '1'` as a string, but the field is an integer field.  In several cases, the database language is capable of returning the record where id is 1, even though it was queried as a string.  If data of one type is being used to find data of another type (and the database language was able to make that conversion), our equality filters here will return False because the data is not the same type - the data is 1, and python is looking for string `1`.

Thinking more on this today..

tests/integration_tests/test_mongo_task.py
```diff
@@ -474,11 +495,130 @@ def test_filter_on_data_categories_mongo(
...
+            "children": ["Erica Example"],
+        }
+    ]
+
+    # items in array mongo_test:customer_details.travel_identifiers used to lookup matching array elements
+    # in mongo_test:flights:passenger_information.passenger_ids.  passenger_information.full_name has relevant
+    # data category.
+    assert filtered_identifiable["mongo_test:flights"][0] == {
+        "passenger_information": {"full_name": "Jane Customer"}
+    }
```
seanpreston, line 570: If we're only expecting `filtered_identifiable["mongo_test:flights"]` to have one element, it would be good to assert that here also

src/fidesops/task/filter_element_match.py
```diff
@@ -0,0 +1,127 @@
...
+    The given row was retrieved from a dataset because an element in row["A"] matched 2 or an element in
+    row["C"]["D"] matched 5.  row["A"] is filtered to just contain the matching element, and row["C"] is filtered
+    to just contain the objects where "D" = 5.
+
+    filter_element_match(
+        row={"A": [1, 2, 3], "B": 2, "C": [{"D": 3, "E": 4}, {"D": 5, "E": 6}, {"D": 5, "E": 7}]},
+        query_paths={FieldPath("A"): [2], FieldPath("C, "D"): 5}
+    )
+
+    {"A": [2], "B": 2, "C": [{"D": 5, "E": 6}, {"D": 5, "E": 7}]}
```
seanpreston, line 38: You've done a great job explaining what's being filtered here, and how that works. It might also be worth stating that paths not in `query_paths` won't be filtered at all? I was initially confused as to why `"B": 2` was in the results for instance

src/fidesops/task/filter_element_match.py
```diff
@@ -0,0 +1,127 @@
...
+    Used by filter_element_match, remove array elements from row that are not specified in preserve_indices
+
+    :param row: Record retrieved from a dataset
+    :param preserve_indices: A dictionary of dot-separated paths to arrays, where the values are the list of indices
+    we want to *keep*
+    :return: A filtered row that has removed non-specified indices from arrays
+
+    :Example:
+    The first element in row["A"]["B"] was the only one specified to preserve, so we remove the other two.
+    _remove_paths_from_row(_remove_paths_from_row({"A": {"B": [{"C": "D"}, {"C": "F"}, {"C": "G"}]}}, {"A.B": [0]})
```
seanpreston, line 64: Would be good to include a simpler example where the function is only called once, too

src/fidesops/task/filter_results.py
```diff
@@ -0,0 +1,79 @@
...
+                saved[key] = select_and_save_field(
+                    saved[key], row[key], FieldPath(*target_path.levels[1:])
+                )
+    return saved
+
+
+RecursiveRow = Union[Dict[Any, Any], List[Any]]
+
+
+def remove_empty_objects(row: RecursiveRow) -> RecursiveRow:
```
seanpreston, line 55: Why aren't we removing empty arrays too?

src/fidesops/task/filter_element_match.py
pattisdr: oh sorry typo in the docstring, thank you

tests/integration_tests/test_mongo_task.py
pattisdr: good idea, I can add that assertion

src/fidesops/task/filter_element_match.py
pattisdr: An important point too is that it's only array elements that are being removed, so if {"B": 2} was in the `query_path`, there should be no effect. The goal of `filter_element_match` is to remove non-matched elements in _arrays_.

src/fidesops/task/filter_results.py
pattisdr: An earlier commit removed the arrays, and I reverted it because my primary reason for removing empty dicts is because I created them. 

There are cases when I can preemptively add dicts to the results in `select_and_save_field`, expecting data to be retrieved in that row.  However, sometimes a row may not have data on a field path, so I created an empty dict for no reason.  I am trying to delete dicts that I created.   I don't think I am creating empty arrays here, so I didn't want to delete an empty array that was in the user data.  Although I can remove it too, it's probably not adding anything, this would be mostly restoring an old commit.

data/nosql/mongo-init.js
```diff
@@ -16,33 +16,51 @@ db.customer_details.insert([
         "customer_id": 1,
```
seanpreston, line 16: Not for part of this PR, but for this to match we'll need to convert these to `"1"`, since that's the type in `db.customer.inset(...)`

tests/integration_tests/test_mongo_task.py
pattisdr: added!

data/nosql/mongo-init.js
pattisdr: It might actually be good to get the mismatched types to work (using integer to find a string, although I think that statement inserts it as float, maybe we insert it as an integer instead).  We in theory support some limited type coercion to use data of one type to find data of another type, but don't have a lot of examples of it. 

This is somewhat related to my type coercion commit here https://github.com/ethyca/fidesops/pull/194/commits/741c9dd34a2e5ed43e2333eb5ae3fb507f58dc57

src/fidesops/task/refine_target_path.py
pattisdr: addressed here https://github.com/ethyca/fidesops/pull/194/commits/741c9dd34a2e5ed43e2333eb5ae3fb507f58dc57

tests/integration_tests/test_mongo_task.py
```diff
@@ -490,7 +631,12 @@ def connector(self, integration_mongodb_config):
     def traversal_node(self, example_datasets, integration_mongodb_config):
         dataset = FidesopsDataset(**example_datasets[1])
         graph = convert_dataset_to_graph(dataset, integration_mongodb_config.key)
-        node = Node(graph, graph.collections[0])  # customer collection
+        customer_details_collection = [
```
seanpreston, line 634: I don't think this will ever be significant since `graph.collections` is limited to the number of collections in the datastore. We could save ourselves some iterations by using:
```
customer_details_collection = None
for collection in graph.collections:
    if collection.name == "customer_details":
        customer_details_collection = collection
        break
```

tests/task/test_consolidate_query_matches.py
```diff
@@ -0,0 +1,50 @@
...
+    target_path = FieldPath("A")
+    assert consolidate_query_matches(input_data, target_path) == [1, 2, 3]
+
+    # Array of embedded objects have multiple matching sub-paths merged
+    field_path = FieldPath("A", "B")
+    input_data = {"A": [{"B": 1, "C": 2}, {"B": 3, "C": 4}, {"B": 5, "C": 6}]}
+    assert consolidate_query_matches(input_data, field_path) == [1, 3, 5]
+
+    # Nested array returned
+    input_data = {"A": {"B": {"C": [9, 8, 7]}}, "D": {"E": {"F"}}}
```
seanpreston, line 22: Is the intention to pass `{"F"}` in as a set here?

pattisdr: nope, but I was just adding extra keys to this dictionary - that path isn't crawled. I'll fix.

src/fidesops/task/filter_results.py
pattisdr: OK removing empty arrays now, this also has a cascading effect, removing an empty dict or empty array can cause higher levels in the path to be removed as well (this was already happening).

data/nosql/mongo-init.js
pattisdr: I've made them the same types for now as I'm not really testing the customer collection, but this could be improved in the future.

src/fidesops/task/filter_element_match.py
pattisdr: I added a few points that this is for arrays, this docstring is being rewritten too in the array erasure PR as some of these methods are being used for multiple purposes.

pattisdr:
Addressing failing tests ^

src/fidesops/task/filter_results.py
```diff
@@ -0,0 +1,83 @@
...
+logger = logging.getLogger(__name__)
+
+
+def select_and_save_field(saved: Any, row: Any, target_path: FieldPath) -> Dict:
+    """Extract the data located along the given `target_path` from the row and add to the "saved" dictionary.
+
+    Entire rows are returned from your collections; this function will incrementally just pull the PII from the rows,
+    by retrieving data along target_paths to relevant data categories.
+
+    To use, you might start with a "saved" that is an empty dict, and loop through a list of FieldPaths you want,
```
seanpreston, line 15: > you might start with a "saved"

Do you mean to say "saved" collection?

pattisdr: Ha I meant the variable is called "saved", just like I refer to the "row" or the "target_path", but yes this is confusing, I'll rephrase.

src/fidesops/task/filter_element_match.py
```diff
@@ -0,0 +1,127 @@
...
+    for path, preserve_index_list in desc_path_length.items():
+        matched_array: List = pydash.objects.get(row, path)
+        if matched_array is None:
+            # This case shouldn't happen - if this gets logged, we've done something wrong
+            logger.info(
+                f"_remove_paths_from_row call: Path {path} in row {row} not found."
+            )
+            continue
+        # Loop through array in reverse to delete indices
+        for i, _ in reversed(list(enumerate(matched_array))):
```
seanpreston, line 86: Why are we `enumerate()`ing `matched_array` only to ignore the index?

seanpreston (approved):
Thanks for the hard work here @pattisdr — in particular shout out to the elegant recursive methods, and nicely organised tasks (and their corresponding tests)

pattisdr:
thank you for reviewing @seanpreston

PR #215, opened 2022-02-11
80 Query Execution BigQuery

### LINKED ISSUE #80: [DB Compatibility] Query execution for scalar values in BigQuery
Requires #76  to be complete first.

We need to enable the BigQuery `ConnectionConfig` db connector to be able to read/write to the database itself during query execution.

eastandwestwind:
# Purpose
Adds query execution (access and erasure) support for BigQuery

# Changes
- Adds BigQuery-specific query config. BigQuery, like Microsoft SQLServer, does not like Tuples in the query string. Otherwise, tuples are converted to quotes. E.g. SQL statements will be formatted like `select * from table.address where "id" = 123` instead of the correct format `select * from table.address where id = 123` 
- For BigQuery Erasures only, constructs update statements using the `update` method in sqlalchemy (https://docs.sqlalchemy.org/en/14/tutorial/data_update.html#the-update-sql-expression-construct) instead of using `TextCaluse`
- Adds integration tests for BigQuery

# Checklist

# Ticket

Fixes #80
 
# Testing

Manually made access request in Postman, obtaining the following results:

```
{"bigquery_example_test_dataset:customer": [{"email": "customer-1@example.com", "name": "John Customer"}], "postgres_example:address": [{"city": "Exampletown", "house": 123, "state": "NY", "street": "Example Street", "zip": "12345"}, {"city": "Exampletown", "house": 4, "state": "NY", "street": "Example Lane", "zip": "12321"}], "bigquery_example_test_dataset:address": [{"city": "Exampletown", "house": "123", "state": "NY", "street": "Example Street", "zip": "12345"}], "bigquery_example_test_dataset:visit": [{"email": "customer-1@example.com"}], "postgres_example:payment_card": [{"ccn": 123456789, "code": 321, "name": "Example Card 1"}], "postgres_example:service_request": [{"alt_email": "customer-1-alt@example.com"}], "bigquery_example_test_dataset:service_request": [{"alt_email": "customer-1-alt@example.com"}], "bigquery_example_test_dataset:payment_card": [{"ccn": 123456789, "code": 321, "name": "Example Card 1"}], "mongo_test:customer_details": [{"gender": "male", "birthday": "1988-01-10T00:00:00"}], "postgres_example:visit": [{"email": "customer-1@example.com"}], "postgres_example:customer": [{"email": "customer-1@example.com", "name": "John Customer"}]}
```


pattisdr (commented):
thanks for your continued efforts on this front @eastandwestwind.  My comments are around consolidation and clarity, plus there are test failures after running the external tests.

src/fidesops/service/connectors/query_config.py
```diff
@@ -520,6 +520,69 @@ def get_formatted_query_string(
         return f'SELECT {field_list} FROM "{self.node.node.collection.name}" WHERE {" OR ".join(clauses)}'
 
 
+class BigQueryQueryConfig(SQLQueryConfig):
+    """
+    Generates SQL valid for BigQuery
+    """
```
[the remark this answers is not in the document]
eastandwestwind: Excellent idea, I'll update this such that both MSSQL and Bigquery can inherit shared methods

eastandwestwind:
Thanks for the review @pattisdr , back over to you!

src/fidesops/service/connectors/query_config.py
```diff
@@ -440,6 +455,14 @@ def generate_query(  # pylint: disable=R0914
         return None
 
 
+class MicrosoftWithoutTuplesSQLServerQueryConfig(
```
pattisdr, line 458: Just a quick note - I did like the original MicrosoftSQLServerQueryConfig name here, same for BigQuery -

eastandwestwind: ah, this was a mistake in refactoring the name!

pattisdr (commented):
Thanks for the reorganization here!  Wonderful docstrings on the methods in the common class that clearly explain the need for the overrides 🏆  I see you're still working on tests.

eastandwestwind:
Discovered that erasure tests were failing due to `TextClause` not properly handing None value for insert into Bigquery:

```
>>> s = text("update dataset.customer set name = :x where id = :y")
>>> engine.execute(s, x=None, y=3)
    raise exceptions.ProgrammingError(
google.cloud.bigquery.dbapi.exceptions.ProgrammingError: Encountered parameter x with  value None of unexpected type.
```

Using `update` instead works:

```
>>> engine = create_engine('bigquery://project/dataset, credentials_info=c)
>>> table = Table('dataset.customer', MetaData(bind=engine), autoload=True)
>>> table.update().where(table.c.id==3).values(name=None)
<sqlalchemy.sql.dml.Update object at 0x102d91ac0>
```

I've updated/refactored BigQuery mask data to use `update`. It's possible in the future to update all connectors to use `update`, as its code complexity is lower than using text clause.

src/fidesops/service/connectors/query_config.py
```diff
@@ -497,6 +520,42 @@ def get_formatted_query_string(
...
+        if not valid:
+            logger.warning(
+                f"There is not enough data to generate a valid update statement for {self.node.address}"
+            )
+            return None
+
+        table = Table(
+            self.node.address.collection, MetaData(bind=client), autoload=True
+        )
+        pk_clauses = [
```
pattisdr, line 553: Can you add a type to pk_clauses?

src/fidesops/service/connectors/sql_connector.py
```diff
@@ -357,6 +359,32 @@ def create_client(self) -> Engine:
...
+        rows: List[Row],
+    ) -> int:
+        """Execute a masking request. Returns the number of records masked"""
+        query_config = self.query_config(node)
+        update_ct = 0
+        client = self.client()
+        for row in rows:
+            update_stmt: Optional[Executable] = query_config.generate_update(
+                row, policy, request, client
+            )
```
pattisdr, line 381: Hmm, I've gone back and forth on this, whether a new "generate_update" method is more clear, as you've demonstrated, or alternatively, we override `BigQueryQueryConfig.generate_update_stmt` to return a Sqlalchemy Update object, which can be executed just like a TextClause, so we don't have to override `mask_data`.   And we have `generate_update_stmt` types to expand to be any type of Executable. What are your thoughts?

src/fidesops/service/connectors/query_config.py
```diff
@@ -497,6 +520,42 @@ def get_formatted_query_string(
...
+    """
+    Generates SQL valid for BigQuery
+    """
+
+    def generate_update(
+        self, row: Row, policy: Policy, request: PrivacyRequest, client: Engine
+    ) -> Optional[Update]:
+        """
+        Using TextClause to insert 'None' values into BigQuery throws an exception, so we use update clause instead
+        """
```
pattisdr, line 533: Can you add to this description that it returns a SQLAlchemy Update object, that the update itself is not occurring here?  mostly for clarification because sqlalchemy's syntax itself can be confusing - table.update().where(*pk_clauses).values(**update_value_map)

src/fidesops/service/connectors/query_config.py
```diff
@@ -497,6 +520,42 @@ def get_formatted_query_string(
...
+            )
+            return None
+
+        table = Table(
+            self.node.address.collection, MetaData(bind=client), autoload=True
+        )
+        pk_clauses = [
+            getattr(table.c, k) == v for k, v in non_empty_primary_keys.items()
+        ]
+        return table.update().where(*pk_clauses).values(**update_value_map)
```
pattisdr, line 556: Very nice solution to this @eastandwestwind, the SQLAlchemy table update method is much more readable

pattisdr (commented):
Great debugging here, Catherine, just a few more comments -

tests/fixtures/bigquery_fixtures.py
```diff
@@ -28,9 +43,9 @@ def bigquery_connection_config(db: Session) -> Generator:
         },
     )
     # Pulling from integration config file or GitHub secrets
-    keyfile_creds = integration_config.get("bigquery", {}).get("keyfile_creds") or os.environ.get(
+    keyfile_creds = integration_config.get("bigquery", {}).get("keyfile_creds") or ast.literal_eval(os.environ.get(
```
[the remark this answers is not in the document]
eastandwestwind: yup it is!

data/dataset/bigquery_example_test_dataset.yml
```diff
@@ -114,7 +114,7 @@ dataset:
             fidesops_meta:
               references:
                 - dataset: bigquery_example_test_dataset
-                  field: order.id
+                  field: orders.id
```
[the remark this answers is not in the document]
eastandwestwind: So I'm confused why this hasn't blown up. BigQuery has `order` as a reserved keyword, but we've been able to create a table named `order` and make selects on it. Maybe it's because of the way SQLalchemy concatenates it with the dataset (schema). Do you think I should implement something similar to Redshift to be on the safe side still?

run_infrastructure.py
```diff
@@ -9,7 +9,7 @@
     List,
 )
 
-DOCKER_WAIT = 3
+DOCKER_WAIT = 15
```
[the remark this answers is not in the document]
eastandwestwind: This was previously causing the build to fail, but what I'll try is using a longer `DOCKER_WAIT`, but keep the `wait = min(DOCKER_WAIT * len(datastores), 15)`

data/dataset/bigquery_example_test_dataset.yml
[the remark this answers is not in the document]
pattisdr: Hmm, I ran an access request with a dataset that had an `order` collection in it, which generated the query - `SELECT customer_id,id,shipping_address_id FROM order WHERE customer_id = %(customer_id)s` 

and ran into this issue:
```
 (google.cloud.bigquery.dbapi.exceptions.DatabaseError) 400 Syntax error: Unexpected keyword ORDER at [1:48]+
```

src/fidesops/service/connectors/sql_connector.py
eastandwestwind: My thinking is `generate_update_stmt` was misleading naming in the first place, since it returned a `Optional[TextClause]` instead of using the `update` method of sqlalchemy. So here I'd like to delineate more obviously from that TextClause than an override provides. 

The main thing, though, is we need an additional param, `client`, passed in to `generate_update` to generate the Update object for BigQuery (as opposed to the base `generate_update_stmt` which does not require `client`).

src/fidesops/service/connectors/query_config.py
eastandwestwind: agreed, good idea!

src/fidesops/service/connectors/sql_connector.py
pattisdr: hm, yeah good point

data/dataset/bigquery_example_test_dataset.yml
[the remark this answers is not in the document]
eastandwestwind: Thanks so much for testing this! I'll work on a fix now

pattisdr (approved):
Looks good to me @eastandwestwind!

PR #223, opened 2022-02-16
[#218] Don't always install MSSQL deps

### LINKED ISSUE #218: Remove MSSQL specific installations from Dockerfile when not needed
There are many new steps in the Fidesops container's Dockerfile required for interfacing with MSSQL (https://github.com/ethyca/fidesops/blob/main/Dockerfile#L15-L25). They add to the installation time significantly but aren't always required, so we should strip them out in CI where we know Fidesops won't be interfacing with MSSQL.

seanpreston:
# Purpose

This PR stops the `fidesops` container installing the MSSQL dependencies in the `Dockerfile` if MSSQL is not being used in the integration env.

It's important to note that no matter what, we'll always need to install these dependencies when `fidesops` gets run in production, so that any users can connect an MSSQL datastore.

# Changes

- Adds `MSSQL_REQUIRED` as a `Dockerfile` arg
- Passes `MSSQL_REQUIRED` into `docker compose build` commands where appropriate

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)

# Ticket

Fixes #218


Dockerfile
```diff
@@ -12,25 +14,30 @@ RUN apt-get update && \
...
-    mssql-tools
+    mssql-tools ; fi
 
 # Update pip and install requirements
-COPY requirements.txt dev-requirements.txt ./
+COPY requirements.txt dev-requirements.txt mssql-requirements.txt ./
 RUN pip install -U pip  \
     && pip install 'cryptography~=3.4.8' \
     && pip install snowflake-connector-python --no-use-pep517  \
     && pip install -r requirements.txt -r dev-requirements.txt
```
seanpreston, line 37: We'll want to change this too such that we're not installing `dev-requirements.txt` in a production context

Makefile
```diff
@@ -46,17 +46,17 @@ server: compose-build
 server-shell: compose-build
 	@docker-compose run $(IMAGE_NAME) /bin/bash
 
-integration-shell: compose-build
```
seanpreston, line 49: Marking `compose-build` as a pre-requisite for the scripts to run was causing `docker compose build` to be run twice, taking up time when running any of these commands both locally and in CI.

Makefile
```diff
@@ -159,7 +161,7 @@ clean:
 compose-build:
 	@echo "Tearing down the docker compose images, network, etc..."
 	@docker-compose down --remove-orphans
-	@docker-compose build
+	@docker-compose build --build-arg REQUIRE_MSSQL="true"
```
eastandwestwind, line 164: why is this hard-coded to `true`? Particularly, not sure if `pytest-integration-external` needs mssql deps

seanpreston: When we run Fidesops in production, we'll always want to install the MSSQL deps so that a user can connect an MSSQL instance if they wish. This `compose-build` task is used by the `Makefile`, and not the `run_infrastructure.py` script that `make pytest-integration-external` will eventually use.

Good spot though, we don't need this here in that instance!

PR #226, opened 2022-02-17
Add MongoDB Support for Array Masking [#129]

### LINKED ISSUE #129: [Array + JSON handling] Erasures in complex datatypes for MongoDB

pattisdr:
# Purpose

Allow data to be masked within arrays, as well as data within objects within arrays (Nested object erasure support was already introduced here https://github.com/ethyca/fidesops/pull/165).  Worth noting, individual elements within arrays are targeted and masked, not the entire array.  Support is for MongoDB only, but the most of this work is done in code, and can be used to help support other database implementations in the future.

Example: 
```json
{
  "id": 1,
   "name": "MASKED",
   "children": ["MASKED", "MASKED"],
   "logins": [
       {"date": "2021-01-01", "chat_name": "MASKED"},
       {"date": "2021-01-01", "chat_name": "MASKED"},
       {"date": "2021-01-01", "chat_name": "MASKED"}
    ],
   "identification": {
       "internal_identifier": "123456",
       "ssn": "MASKED",
       "email": "MASKED"
    }
}
```
# Changes Overview
Two primary pieces were needed to support array masking here: building paths to all array elements we need to mask, and having access requests (which cache data that we later use for erasures) preserve the indices of elements we should and shouldn't mask where applicable.

### Cached Access Requests Results need to preserve elements that shouldn't be masked

Access requests on arrays may potentially return data with array elements removed.  However, when running an erasure, we need to know the exact index where those matched elements were located, making this data no longer safe for erasures. To resolve, access requests on a node cache an additional version of the queried data, where we preserve all elements and just indicate that some shouldn't be masked (for erasures), in addition to the version we already cache that has unmatched  elements removed (for access requests).  

  **Original data**
```json
{
   "flight_id":"12345",
   "passengers":[
      {
         "name":"Joe",
         "seat":"6A"
      },
      {
         "name":"Jessie",
         "seat":"3B"
      },
      {
         "name":"Josh",
         "seat":"11C"
      },
   ],
   "destination":"ORD",
   "origin":"DFW"
}
```

**Cached version that we'll use for erasures - NEW**

This both keeps us from building any paths to data at the first and third elements of the passengers array, but also preserves that the index we care about is index 1.

```json
{
   "flight_id":"12345",
   "passengers":[
      "FIDESOPS_DO_NOT_MASK"
      {
         "name":"Jessie",
         "seat":"3B"
      },
      "FIDESOPS_DO_NOT_MASK"
   ],
   "destination":"ORD",
   "origin":"DFW"
}
```
**Cached version that we already return for access requests.** 
The access request only targeted data at the first element of the above array, so that is the only element we return here 

```json
{
   "flight_id":"12345",
   "passengers":[
      {
         "name":"Jessie",
         "seat":"3B"
      },
   ],
   "destination":"ORD",
   "origin":"DFW"
}
```
### Build paths to all array elements we need to mask

To mask array data, we need to build a mapping of every applicable index in that array to the masked value.  Paths targeting array data will be expanded to every possible path. In this example, the object located at the second position of the `passengers` array is going to have its `name` and `seat` fields masked.  Previously, we couldn't build index paths.   

We can use these paths directly in Mongo queries - for other databases, we may need to cache the entire original result, and use the paths here to replace certain data, and then write the entire complex field back.

```python
{
  "passengers.1.name": "abcdefh",
  "passengers.1.seat": "ghijklmn",
}
```

# Detailed Changes

- For caching a separate version of the data for erasures, adds a new parameter to `filter_element_match` (introduced in https://github.com/ethyca/fidesops/pull/194) which allows you to specify that non-matched array elements should be replaced with placeholder text instead of deleted 
- To allow us to build a query targeting array indices, `QueryConfig.update_value_map` was expanded to build paths to all possible array indices instead of singular paths
  - This reuses `build_refined_target_paths` (introduced in https://github.com/ethyca/fidesops/pull/194) modified to optionally expand target paths to all possible paths instead of just paths that just match specific target values.

# Checklist

# Ticket

Fixes #129


src/fidesops/task/filter_element_match.py
```diff
@@ -85,7 +109,10 @@ def _remove_paths_from_row(
         # Loop through array in reverse to delete indices
         for i, _ in reversed(list(enumerate(matched_array))):
             if i not in preserve_index_list:
```
seanpreston, line 111: We might want a comment here explaining that "this is referring to which index in the list to preserve, not which elements in the dataset to preserve" since in the erasures, any items preserved in this list will be masked (if I've understood that correctly?)

tests/fixtures/application_fixtures.py
```diff
@@ -914,3 +918,16 @@ def sample_data():
...
+def mongo_example_db() -> Generator:
+    """Return a connection to the MongoDB example DB"""
+    uri = "mongodb://mongo_user:mongo_pass@mongodb_example/mongo_test"
+
+    client = MongoClient(uri, serverSelectionTimeoutMS=5000)
+    logger.debug(f"Connecting to MongoDB example database at: {uri}")
+    # Setup above...
+    yield client
+    # Teardown below...
+    client.close()
```
seanpreston, line 933: We already have a fixture `integration_mongodb_connector` for this.

pattisdr: ah I was just moving the existing fixture to a place where I could use it more easily, but I'm not sure what changes were made in the fixture PR so I'll take a closer look and get this cleaned up.

src/fidesops/task/filter_element_match.py
pattisdr: Correct, it's which index we want to preserve, I will try to improve clarity.  Maybe I shouldn't even be talking about access/erasure here in this method at all, this should be a standalone helper method that is just removing/replacing elements in arrays as part of a post-processing step.

Importantly, this is also used for access requests too, basically if it's "preserved" here, it's data we care about, it's data we may return in an access request, or mask in an erasure request.

seanpreston: That this method would return `row` in two different contexts feels like a potential foot-gun to me.

To confirm my understanding:
- in the context of an access request, it's fine to remove the unwanted elements from `row`, because the data is being returned, and we don't need to preserve list order
- in the context of an erasure request, we'll need to go back and mask over these elements in the datastore, so we need to preserve the indexes in the list in order to effectively erase the data at those indexes

Especially since we're storing the results separately in either case (placeholder vs. regular), does this mean we're caching a lot of unnecessary PII considering we may only ever need one of those versions?

pattisdr: > To confirm my understanding:

> - in the context of an access request, it's fine to remove the unwanted elements from row, because the data is being returned, and we don't need to preserve list order
> - in the context of an erasure request, we'll need to go back and mask over these elements in the datastore, so we need to preserve the indexes in the list in order to effectively erase the data at those indexes




Correct. In addition, on the access request, we remove the unwanted elements from the array because data in that array is potentially used to locate data on other collections downstream in the graph.

pattisdr: Can you clarify what you mean by "we may only ever need one of those versions?"  as in, the policy only has access rules defined or only has erasure rules defined?

src/fidesops/task/refine_target_path.py
```diff
@@ -120,8 +147,17 @@ def _enter_array(array: List[Any], field_path: List[str], only: List[Any]) -> Li
...
+    """Returns True if the given element is considered a match."""
+    if elem == FIDESOPS_DO_NOT_MASK_INDEX:
+        # This element has already been marked in post-processing as one that shouldn't be considered a match
+        return False
+
+    if not only:
+        # If no values are specified to match, we return all possible paths
+        return True
+
+    # Otherwise, does this element match a value in only?
```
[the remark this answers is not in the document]
seanpreston: It's more transparent in testing too this way, good choice 🙌

src/fidesops/task/refine_target_path.py
```diff
@@ -48,33 +65,40 @@ def build_incoming_refined_target_paths(
...
         current_elem = row[current_level]
-    except KeyError:  # FieldPath not found in record, this is expected to happen when data doesn't exist in collection
-        return []
-    except (IndexError, TypeError):  # No field path or invalid field path
-        logger.warning(f"Error with locating {target_path} on row")
+    except (IndexError, TypeError, KeyError):
+        # KeyError - FieldPath not found, expected when data doesn't exist in collection
+        # IndexError - No FieldPath
+        # TypeError - Bad FieldPath, expected when we're trying to select a path we've already eliminated
         return []
```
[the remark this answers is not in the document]
seanpreston: It would be good to continue warning here I think. Another option could be to use a different class of object (that can still exist in our Redis cache) for placeholders, then the delineation would be clearer since they wouldn't both be strings

tests/integration_tests/test_mongo_task.py
```diff
@@ -502,6 +591,48 @@ def test_object_querying_mongo(
...
+    privacy_request = PrivacyRequest(id=f"test_mongo_task_{random.randint(0,1000)}")
+
+    mongo_dataset, postgres_dataset = combined_mongo_postgresql_graph(
+        integration_postgres_config, integration_mongodb_config
+    )
+    graph = DatasetGraph(mongo_dataset, postgres_dataset)
+
+    access_request_results = graph_task.run_access_request(
+        privacy_request,
+        policy,
```
seanpreston, line 607: Just to note that this is using the empty `policy` that's defined on line 46, rather that the `policy` or `erasure_policy` fixture. I don't think it's significant here because the erasure data will get cached regardless, so maybe that's intentional?

pattisdr: Good eye - you're right that the policy isn't significant for this test, but the overlap between the empty policy and the fixture policy isn't great, I'll rename the empty policy.

tests/fixtures/application_fixtures.py
pattisdr: Fixture removed in favor of integration_mongodb_connector

src/fidesops/task/filter_element_match.py
pattisdr: updates - I've tried to clarify in code comments that this is a list of indices we want to preserve

data/dataset/mongo_example_test_dataset.yml
```diff
@@ -162,10 +162,15 @@ dataset:
...
               data_type: integer
       - name: conversations
-        fidesops_meta:
-          data_type: object[]
         fields:
+          - name: _id
+            data_categories: [ system.operations ]
+            fidesops_meta:
+              primary_key: True
+              data_type: object_id
```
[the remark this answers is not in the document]
seanpreston: I saw you added some clarifying comments to the docs on this point too — thanks!

src/fidesops/task/filter_element_match.py
seanpreston: > Can you clarify what you mean by "we may only ever need one of those versions?" as in, the policy only has access rules defined or only has erasure rules defined?

When I wrote that I was under the impression it is the case that erasures will only ever use the placeholder data. As I understand it now, the access data may well be used for further lookups down the chain. So we do in fact need both

seanpreston (approved):
Thanks @pattisdr 🎉

pattisdr:
thanks for reviewing @seanpreston ⭐

src/fidesops/task/refine_target_path.py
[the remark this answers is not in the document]
pattisdr: OK, restored the warning, again this warning is expected if we are masking an array where some of the elements have been marked with placeholder text, so they will be false positives.  We can tweak later if we are seeing this too much.

PR #236, opened 2022-02-28
Fix Fields Affected on Execution Logs [#144]

### LINKED ISSUE #144: DB ExecutionLogs should only print affected data categories for erasures
Born from the discussion on #38 , we found that `fields_affected` in the logs includes `data_categories` that were not specified on the erasure rule. This is misleading and incorrect. 

The logs `fields_affected` should only output categories that were, in fact, masked by fidesops.  This issue should correct the log output and for any other place that the logs incorrectly print what was affected, correct that as well.

pattisdr:
# Purpose

Update `ExecutionLog.fields_affected` to only be the fields specified for all the erasure rules or all of the access rules.

When a PrivacyRequest is executed, every collection in the graph has a series of ExecutionLogs saved in the fidesops application database with a high-level summary of what occurred for troubleshooting.  For successful execution logs, under the `fields_affected` we've been recording every field on that collection instead of only the fields relevant for that particular policy + action_type (erasure/access).

# Changes

- Add a new method `build_affected_field_logs` for use when we create a completed ExecutionLog for a given node.  For all of the rules of a given action_type (erasure or access) on the specified policy, it gets all of the fields for the requested data_categories.  Note that this doesn't mean data actually existed on that particular collection, but it's what we looked for. We don't look at data actually returned, because a collection can return multiple rows, and one row might return/mask all specified fields, the next row may only have data for some of the fields, so this is just a high-level summary of all fields considered for the policy.

# Checklist

# Ticket

Fixes #144


NevilleS:
More of a question than feedback @pattisdr: if we wanted to add the actual DB queries that were run to these logs, how might we do that? For example, the logs today are organized by field (e.g. every field affected is added to results) but if we wanted to understand how they were affected, how might you include that context?

pattisdr:
@NevilleS Big picture, I'd probably cache the queries as we constructed them instead of reconstructing the details for the logs at the end. I might add an array field to `ExecutionLog` to store the list of queries made on that collection, and maybe change the `ExecutionLog.fields_affected` column to be to `ExecutionLog.fields_targeted` to more accurately reflect its contents.   There might be some more discussion around whether `ExecutionLogs` are being organized at the proper level.  I'd need product input on redaction requirements before persisting the queries. At least for access requests, the queries wouldn't be that useful if we couldn't persist the "where" clause or equivalent.

---
More details:

`ExecutionLogs` are currently created at the Collection level - one log per collection per life cycle step.  This one `ExecutionLog` may encompass multiple Policy Rules, and regard multiple rows.  The `ExecutionLog.fields_affected` is really a summation of all the fields that were potentially affected on this collection given the Policy and Dataset - a field may have been touched on one row, and not another for example, that breakdown is not included currently.  To meet the given requirements, we've been able to build the `fields_affected` just using the Policy/Dataset. 

If we want to store actual queries, we should probably cache these queries as we make them instead of waiting until after the fact to build them.  Access request queries are less informative than you might think; we select all fields on a given collection that meet certain conditions, and then filter them potentially multiple ways at the end, for each separate Policy Rule.  So the query might be (`select city, state, address, id, birthday, email from user where id = 1`;) but we uploaded two SAR packages! One that included city/state/address and another that included just birthday because there were two separate rules defined, that each targeted different categories.  Here, `fields_affected` is more useful than the query itself, because we know that birthday/city/state/address fields are all potentially affected from inspecting the Policy/Dataset.  Erasure queries on the other hand, are more targeted and could add value to the ExecutionLogs, depending on how much information needed to be redacted.

docs/fidesops/docs/guides/reporting.md
```diff
@@ -77,60 +78,64 @@ Check out the [API docs here](/fidesops/api#operations-Privacy_Requests-get_requ
 
 ## View Individual Privacy Request Log Details
 
-Use the `verbose` query param to see more details about individual queries run as part of the Privacy Request along
-with individual statuses. 
+Use the `verbose` query param to see more details about individual collections visited as part of the Privacy Request along
+with individual statuses. Individual collection statuses include `in_processing`, `retrying`, `complete` or `error`.
```
seanpreston, line 82: Should we add `paused` here too?

docs/fidesops/docs/guides/reporting.md
```diff
@@ -77,60 +78,64 @@ Check out the [API docs here](/fidesops/api#operations-Privacy_Requests-get_requ
...
+You may see multiple logs for each collection as they reach different steps in the lifecycle.  
 
 `verbose` will embed a “results” key in the response, with execution logs grouped by dataset name.  In the example below,
-we have two datasets: `my-mongo-db` and `my-postgres-db`. There is one execution log for my-mongo-db and two execution
-logs for my-postgres-db.  The embedded execution logs are automatically truncated at 50 logs, so to view the entire 
-list of logs, visit the execution logs endpoint separately.
+we have two datasets: `my-mongo-db` and `my-postgres-db`. There are two execution logs for my-mongo-db (when the `flights`
```
seanpreston, line 86: Should have `s around the second `my-mongo-db`

docs/fidesops/docs/guides/reporting.md
pattisdr: Paused isn't a status for an ExecutionLog, it's just a status for the PrivacyRequest.

docs/fidesops/docs/guides/reporting.md
pattisdr: Fixed!

pattisdr:
comments addressed @seanpreston

seanpreston:
Thanks for that @pattisdr, great work!

seanpreston:
Another question with regard to:

> if we wanted to add the actual DB queries that were run to these logs, how might we do that?

What value will this offer Fidesops? Are there instances where the queries we execute will differ from what's returned by the preview?

docs/fidesops/docs/guides/reporting.md
seanpreston: Oh Of course. If a request is paused it's not executing anything. Gotcha

NevilleS:
> What value will this offer Fidesops? Are there instances where the queries we execute will differ from what's returned by the preview?

That's fair, the preview is *probably* what I want. The use case I'm imagining is really about getting users _comfortable_ with what fidesops is doing when a request is executed. Making it really clear which queries are running is what I'd want to see as an engineer implementing this on my system.

@pattisdr thanks for those detailed notes and it's helpful to see how you're thinking about it. I don't have a specific feature request quite yet, but I'm looking at the types of questions people ask when they first start learning about the system, and understanding more about the logging is a big one!

Issue #182, opened 2022-01-26
[SaaS Connectors] API endpoints for SaaS connector configuration


seanpreston:
# Purpose
Modify or add API endpoints to fulfill the following:

1. Create a ConnectionConfig of type saas
2. Configure the connection with a saas_config (yaml)
3. Upload a corresponding dataset yaml for the saas connector
4. Run a privacy request via API for the configured SaaS Connector

PR #237, opened 2022-02-28
176 mask data implementation

### LINKED ISSUE #176: [SaaS Connector Engine] Define mask_data implementation
Building off of the first version of the SaaS connector engine - we need to define the implementation for handling erasure requests (right to be forgotten) to third party applications

Show that in the event of an erasure Request, we can retrieve data from a third party application (can validate using stripe, zendesk, hubspot or slack)

### Additional context
For this first pass, we don't need to implement features for pagination, retry handling, parameter preprocessing, response post processing. This can be as vanilla of a DELETE (perhaps UPDATE) request as possible.

galvana:
# Purpose
Adds the `mask_data` implementation for SaaS connectors based on the SaaS config. Dynamically builds the update request with a masked payload.

# Changes
1. Added `mask_data` implementation to saas_connector
2. Added `generate_update_stmt` and `prepare_update_params` to `SaaSQueryConfig`

# Checklist

Closes #176


tests/fixtures/saas_fixtures.py
```diff
@@ -99,6 +111,112 @@ def connection_config_saas(
     connection_config.delete(db)
 
 
+@pytest.fixture(scope="function")
+def erasure_policy_string_rewrite(
```
seanpreston, line 115: It would be useful for these policies to live alongside the others in `application_fixtures.py`, since they could be used in any test in the application and not only in the SaaS integration tests. That way anyone looking at policies will see every policy we have, and be less likely to add a new one if it's not needed.

tests/fixtures/saas_fixtures.py
```diff
@@ -109,3 +227,29 @@ def mailchimp_account_email():
     return pydash.get(saas_config, "mailchimp.account_email") or os.environ.get(
         "MAILCHIMP_ACCOUNT_EMAIL"
     )
+
+
+@pytest.fixture(scope="function")
+def reset_saas_data(connection_config_saas, mailchimp_account_email) -> Generator:
```
seanpreston, line 233: It would be helpful to indicate that this test is Mailchimp specific in the name.

tests/fixtures/saas_fixtures.py
```diff
@@ -109,3 +227,29 @@ def mailchimp_account_email():
...
+        {"query": mailchimp_account_email},
+        {},
+    )
+    response = connector.create_client().send(request)
+    body = response.json()
+    member = body["exact_matches"]["members"][0]
+    yield member
+    request = (
+        "PUT",
+        f'/3.0/lists/{member["list_id"]}/members/{member["id"]}',
```
seanpreston, line 251: I might well be missing something here — why is it useful to add the member to a list in the teardown on this fixture?

In order to keep this as idempotent as we can, would it be better to add the data we need to Mailchimp in the setup, then remove it again in the teardown? We might also want to use `session` scoped fixtures for this instead of `function` to avoid making these API calls at the start and end of every test.

tests/service/privacy_request/request_runner_service_test.py
```diff
@@ -337,6 +341,47 @@ def test_create_and_process_access_request_saas(
...
+    data = {
+        "requested_at": "2021-08-30T16:09:37.359Z",
+        "policy_key": erasure_policy_hmac.key,
+        "identity": {"email": customer_email},
+    }
+
+    pr = get_privacy_request_results(db, erasure_policy_hmac, cache, data)
+
+    connector = SaaSConnector(connection_config_saas)
+    request = ("GET", "/3.0/search-members", {"query": mailchimp_account_email}, {})
```
seanpreston, line 367: This request is being sent by the `reset_saas_data` fixture's setup too. Until we have an intentionally designed SaaS integrations testing solution in place, it'd be good to try and keep the number of requests sent to external services to a minimum so as to keep the test run time down, and rate limits to as low loadings as possible.

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
+                    params[param.name] = pydash.get(
+                        param_values, param.references[0].field
+                    )
+                elif param.identity:
+                    params[param.name] = pydash.get(param_values, param.identity)
+            elif param.type == "path":
+                path = path.replace(
+                    f"<{param.name}>",
+                    pydash.get(param_values, param.references[0].field),
+                )
```
pattisdr, line 696: There is a lot of duplicated logic from `prepare_query_params` - it seems like prepare_query_params could also use the pydash methods you're using here.  Is there any way to consolidate?

I may not fully understand the extent of the differences that made you feel you needed to separate the two.

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
+        path: str = request.path
+        params: Dict[str, Any] = {}
+
+        for param in request.request_params:
+            if param.type == "query":
+                if param.default_value:
+                    params[param.name] = param.default_value
+                elif param.references:
+                    params[param.name] = pydash.get(
+                        param_values, param.references[0].field
```
pattisdr, line 688: Does this need to take into account the direction of the reference, whether "from" or "two" has been defined?  What if there are multiple references? I am a little unclear about what it is happening here.

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
+                    )
+                elif param.identity:
+                    params[param.name] = pydash.get(param_values, param.identity)
+            elif param.type == "path":
+                path = path.replace(
+                    f"<{param.name}>",
+                    pydash.get(param_values, param.references[0].field),
```
pattisdr, line 695: Why is this getting the reference field but prepare_query_params is getting the param.name?

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
+
     def generate_update_stmt(
         self, row: Row, policy: Policy, request: PrivacyRequest
-    ) -> Optional[List[SaaSRequestParams]]:
-        return None
+    ) -> SaaSRequestParams:
+        update_value_map: Dict[str, Any] = self.update_value_map(row, policy, request)
+        body = paths_to_dict(update_value_map)
+        current_request = self.get_request_by_action("update")
+        collection_name = self.node.address.collection
```
pattisdr, line 706: Can you add some missing type hints for clarity?

data/saas/config/mailchimp_config.yml
```diff
@@ -69,7 +69,22 @@ saas_config:
...
+          data_path: exact_matches.members
+        update:
+          path: /3.0/lists/<list_id>/members/<subscriber_hash>
+          request_params:
+            - name: list_id
+              type: path
+              references:
+                - dataset: mailchimp_connector_example
+                  field: member.list_id
+                  direction: from
```
pattisdr, line 81: All your references are in the "from" direction, but they can also be specified to point the other way, does this affect your implementation?

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -98,14 +101,6 @@ def create_client(self) -> AuthenticatedClient:
         logger.info(f"Creating client to {uri}")
         return AuthenticatedClient(uri, self.client_config, self.secrets)
 
-    @staticmethod
-    def get_value_by_path(dictionary: Dict, path: str) -> Dict:
-        """Helper method to extract an arbitrary data path from a given dictionary"""
-        value = dictionary
-        for key in path.split("/"):
-            value = value[key]
-        return value
```
pattisdr, line 107: Nice swap to use pydash instead of this custom method - pydash has some great features.

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -49,14 +51,15 @@ def get_authenticated_request(
...
-        req = Request(url=f"{self.uri}{path}", params=params, data=data).prepare()
+        (method, path, params, data) = request_params
+        req = Request(
+            method=method, url=f"{self.uri}{path}", params=params, data=data
+        ).prepare()
         return self.add_authentication(req, self.client_config.authentication)
 
-    def get(self, request_params: SaaSRequestParams) -> Response:
+    def send(self, request_params: SaaSRequestParams) -> Response:
         """Builds and executes an authenticated GET request"""
```
pattisdr, line 61: Can you update docstring to be more accurate?

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -142,6 +135,15 @@ def mask_data(
         rows: List[Row],
     ) -> int:
         """Execute a masking request. Return the number of rows that have been updated"""
+        query_config = self.query_config(node)
+        prepared_requests = [
+            query_config.generate_update_stmt(row, policy, request) for row in rows
+        ]
+        rows_updated = 0
+        for prepared_request in prepared_requests:
+            self.client().send(prepared_request)
```
pattisdr, line 144: What happens if you get a non-200 response from this request?

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -120,13 +115,11 @@ def retrieve_data(
 
         rows: List[Row] = []
         for prepared_request in prepared_requests:
-            response = self.client().get(prepared_request)
+            response = self.client().send(prepared_request)
```
pattisdr, line 118: What happens if you get a non-200 response in `retrieve_data`?

src/fidesops/service/connectors/query_config.py
```diff
@@ -597,11 +598,11 @@ def dry_run_query(self) -> Optional[str]:
         return None
 
 
-SaaSRequestParams = Tuple[str, Dict[str, Any], Dict[str, Any]]
+SaaSRequestParams = Tuple[Literal["GET", "PUT"], str, Dict[str, Any], Optional[str]]
 """Custom type to represent a tuple of path, params, and body values for a SaaS request"""
 
 
-class SaaSQueryConfig(QueryConfig[List[SaaSRequestParams]]):
+class SaaSQueryConfig(QueryConfig[Union[List[SaaSRequestParams], SaaSRequestParams]]):
```
pattisdr, line 605: I don't understand the new type hinting , what is this?

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
             for value in reference_values:
                 request_params.append(
-                    self.prepare_params(current_request, {string_path: value})
+                    self.prepare_query_params(current_request, {string_path: value})
                 )
         logger.info(f"Populated request params for {current_request.path}")
         return request_params
 
+    @staticmethod
+    def prepare_update_params(
```
pattisdr, line 673: It would be useful to see some unit tests on your internals - `prepare_update_params` and `prepare_query_params`, you have a nice end-to-end test but I wonder if we're missing corner cases

data/saas/config/mailchimp_config.yml
```diff
@@ -69,7 +69,22 @@ saas_config:
               type: query
               identity: email
               data_type: string
-          data_path: exact_matches/members
+          data_path: exact_matches.members
+        update:
+          path: /3.0/lists/<list_id>/members/<subscriber_hash>
+          request_params:
+            - name: list_id
+              type: path
```
pattisdr, line 77: Are there examples of "body" request params? Can those be nested?

pattisdr (commented):
Pass complete @galvana, I have questions about whether this is general enough to work for other API's, but a lot of this is coming from not having a good understanding of the API's we'll be integrating with. I think I should start with digging into mailchimp's API and understand that better first.

Generally this felt like a great proof of concept for saas masking, but few corner cases are explored here.

tests/service/connectors/test_queryconfig.py
```diff
@@ -623,27 +625,67 @@ def test_generate_query(self, policy, combined_traversal, connection_config_saas
...
         # static path with multiple query params with default values
         config = SaaSQueryConfig(conversations, endpoints)
-        prepared_requests = config.generate_query({"placeholder": ["customer-1@example.com"]}, policy)
+        prepared_requests = config.generate_query(
+            {"placeholder": ["customer-1@example.com"]}, policy
+        )
         assert prepared_requests[0] == (
+            "GET",
             "/3.0/conversations",
             {"count": 1000, "placeholder": "customer-1@example.com"},
```
pattisdr, line 646: I missed this, what is this placeholder query param? I can't seem to find it in the docs

src/fidesops/service/connectors/query_config.py
```diff
@@ -597,11 +598,11 @@ def dry_run_query(self) -> Optional[str]:
         return None
 
 
-SaaSRequestParams = Tuple[str, Dict[str, Any], Dict[str, Any]]
+SaaSRequestParams = Tuple[Literal["GET", "PUT"], str, Dict[str, Any], Optional[str]]
 """Custom type to represent a tuple of path, params, and body values for a SaaS request"""
```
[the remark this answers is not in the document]
galvana: Updated
```Custom type to represent a tuple of HTTP method, path, params, and body values for a SaaS request```

File: src/fidesops/service/connectors/query_config.py (hunk 14 above)
Comment (galvana) [reply src/fidesops/service/connectors/query_config.py:605]: Generate query returns `Optional[List[SaaSRequestParams]]` and generate_update_stmt returns `SaaSRequestParams` but QueryConfig defines both of these methods as returning generic type `T`. This was the only way I could think of to accommodate both of these return types.

File: src/fidesops/service/connectors/query_config.py [hunk 19]
```diff
@@ -644,10 +644,8 @@ def prepare_params(
                     params[param.name] = param_values[param.name]
             elif param.type == "path":
                 path = path.replace(f"<{param.name}>", param_values[param.name])
-            elif param.type == "body":
-                data[param.name] = param_values[param.name]
 
-        return path, params, data
+        return ("GET", path, params, None)
```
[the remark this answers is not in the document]
galvana: Fixed.

src/fidesops/service/connectors/query_config.py
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
+                    )
+                elif param.identity:
+                    params[param.name] = pydash.get(param_values, param.identity)
+            elif param.type == "path":
+                path = path.replace(
+                    f"<{param.name}>",
+                    pydash.get(param_values, param.references[0].field),
+                )
+
+        return ("PUT", path, params, json.dumps(body))
```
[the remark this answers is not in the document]
galvana: Fixed.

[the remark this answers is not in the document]
galvana: Fixed.

src/fidesops/service/connectors/saas_connector.py
galvana: ```Builds and executes an authenticated request. The HTTP method is determined by the request_params.```

tests/fixtures/saas_fixtures.py
galvana: Changed fixture name to `reset_mailchimp_data`

tests/fixtures/saas_fixtures.py
galvana: At a high level, we're reading the current `member` value from Mailchimp and restoring it in the teardown. I'm using this as part of every Mailchimp erasure to restore the original value after asserting the erasure works. I also did this to avoid having to specify `list_id` and `id` in the test, using this approach we can just read it from the `GET` request and use those values to put the member back in the specific location in the teardown.

tests/service/connectors/test_queryconfig.py
galvana: The placeholder was used to create an artificial link to the `conversations` node. Since this endpoint doesn't have any dependencies on identity or references, this was the way I could get this node to be considered reachable without making changes to the traversal logic.

Ideally, it would be nice to have nodes like these (without dependencies) to be treated as root nodes and still be considered reachable.

tests/service/privacy_request/request_runner_service_test.py
galvana: Keeping this in mind for the future. I'd like to develop a better test suite for SaaS connectors where we mock most of the interactions with a given API and only have a single integration test with the actual API to verify that nothing has changed in the API spec.

tests/fixtures/saas_fixtures.py
galvana: Good idea, moved these to `application_fixtures.py`

data/saas/config/mailchimp_config.yml
galvana: The way I understood direction to work is that if we specify the direction as `from` in the SaaS config, then the SaaS connector nodes would receive the necessary data from the referenced datasets as part of the graph task execution. I didn't see a need to specify any references as `to` since those datasets could just reference the SaaS dataset using a `from` reference. Is this a valid understanding/assumption?

src/fidesops/service/connectors/query_config.py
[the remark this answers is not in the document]
galvana: Updated
```Custom type to represent a tuple of HTTP method, path, params, and body values for a SaaS request```

File: src/fidesops/service/connectors/query_config.py (hunk 14 above)
Comment (galvana) [reply src/fidesops/service/connectors/query_config.py:605]: generate_query returns `Optional[List[SaaSRequestParams]]` and generate_update_stmt returns `SaaSRequestParams` but QueryConfig expects both of these to return the same generic type `T`. This type hint was the only way I could have both of these methods return their own type.

File: src/fidesops/service/connectors/query_config.py [hunk 21]
```diff
@@ -666,15 +664,47 @@ def generate_query(
...
                 )
         logger.info(f"Populated request params for {current_request.path}")
         return request_params
 
+    @staticmethod
+    def prepare_update_params(
+        request: SaaSRequest, param_values: Dict[str, Any], body: Dict[str, Any]
+    ) -> SaaSRequestParams:
+        """
+        Populates the placeholders in the request with the given param values
```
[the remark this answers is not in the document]
galvana: Updated
``` 
Populates the placeholders in the request with the given
param values and converts the body dict into a JSON string
to include in the body of the request.

File: data/saas/config/mailchimp_config.yml (hunk 9 above)
Comment (pattisdr) [reply data/saas/config/mailchimp_config.yml:81]: You can use `from`, my question to rephrase is, does it break if someone tries to use `to` with Saas configs, our other datasets have both `to` and `from` directions between their collections, but is it a requirement for saas configs that only the "from" direction is used?

File: src/fidesops/service/connectors/query_config.py (hunk 8 above)
Comment (galvana) [reply src/fidesops/service/connectors/query_config.py:706]: Added!

File: src/fidesops/util/saas_util.py [hunk 22]
```diff
@@ -48,3 +49,32 @@ def merge_datasets(dataset: Dataset, config_dataset: Dataset) -> Dataset:
...
+
+    becomes
+
+    {
+        "object": {
+            "first_property": "one",
+            "second_property" "two"
+        }
+    }
+    """
```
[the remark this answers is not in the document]
galvana: Thanks for this! Unflatten was the keyword I was missing when I was researching this. I hope you don't mind, I borrowed the error handling from your implementation and I augmented/cleaned up the test cases using yours as a reference. With that said, should this be moved back to it's own util class since this has the potential of being used outside of the SaaS work?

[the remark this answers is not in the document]
pattisdr: either way - we can move it later if someone else wants to use it too

data/saas/config/mailchimp_config.yml
galvana: I have removed "body" request_params from the SaaS config since I didn't see any use cases for this in the current legacy adaptors. We can scope this out in the future if there is a need.

tests/service/connectors/test_queryconfig.py
pattisdr: ah ok this would be a great thing to document in the saas config guides -

data/saas/config/mailchimp_config.yml
galvana: I'll add this to #182 since that's where I have the endpoints/tests.

src/fidesops/service/connectors/saas_connector.py
galvana: Going to fix this in #182

src/fidesops/service/connectors/saas_connector.py
galvana: Going to fix this in #182

src/fidesops/service/connectors/query_config.py
galvana: Restricting request_param references to only use the `from` direction

tests/fixtures/saas_fixtures.py
seanpreston: I can see why this approach would be appealing, because it's the lightest touch on the external API possible. I'd worry that it relying on data being present on Mailchimp leaves it susceptible to interference — any change in the data on Mailchimp will affect whether any test using this fixture passes or fails.

data/saas/config/mailchimp_config.yml
galvana: Merged in validation from main branch.

src/fidesops/service/connectors/saas_connector.py
galvana: Fixed with changes merged in from main branch.

src/fidesops/service/connectors/saas_connector.py
galvana: Fixed with changes merged in from main branch.

data/saas/config/mailchimp_config.yml
pattisdr: thanks @galvana

src/fidesops/service/connectors/saas_connector.py
pattisdr: thanks for the update!

src/fidesops/service/connectors/query_config.py
pattisdr: @galvana, OK, right, this highlights some differences, because with our other implementations `generate_query` returns a single statement and `generate_update_stmt` returns a single statement, but for saas configs, we might need to make multiple API requests.

For improved organization, you might rename `prepare_query_params` to be `generate_query`, and then `generate_query` becomes `generate_requests` or something like that.  That way `generate_query` and `generate_update_stmt` both return a single "query" or "request".

data/saas/config/mailchimp_config.yml
pattisdr: ok thanks for clarifying

src/fidesops/service/connectors/query_config.py
galvana: The `param_values` param in `prepare_query_params` is a map of request_param.names to values. So if a request_param is defined as

```
request_params:
  - name: query
    type: query
    identity: email
    data_type: string
```
we can just use the name value `query` as the dictionary key

However the `param_values` param in `prepare_update_params` is a map of fields to values. So for something like this
```
- name: subscriber_hash
  type: path
  references:
    - dataset: mailchimp_connector_example
      field: member.id
      direction: from
```
We have to use the `field` value as the key instead of name (`subscriber_hash` in this case)

pattisdr: thanks for the clarification!  The type signatures being the same makes this confusing, is the param_values type for `prepare_update_params` possibly `Dict[str, Row]`? I might make variable names more distinct for `prepare_update_params` and add more detail to the docstring.

src/fidesops/service/connectors/query_config.py
galvana: That's a nice way to clean this up, I implemented your recommendation.

src/fidesops/service/connectors/query_config.py
pattisdr: answered above -

src/fidesops/service/connectors/query_config.py
```diff
@@ -694,56 +695,83 @@ def get_request_by_action(self, action: str) -> SaaSRequest:
...
-    ) -> Optional[List[SaaSRequestParams]]:
+    def generate_update_stmt(
+        self, row: Row, policy: Policy, request: PrivacyRequest
+    ) -> SaaSRequestParams:
         """
-        This returns the query/path params needed to make an API call.
-        This is the API equivalent of building the components of a database
-        query statement (select statement, where clause, limit, offset, etc.)
+        Takes a row and masks the fields based on the policy. This masked row is then added as the
+        body to a dynamically generated SaaS request.
```
pattisdr, line 745: Thanks for the improved docstring! - one minor point, we haven't actually made the request yet, and the first bit indicates otherwise

src/fidesops/service/connectors/query_config.py
galvana: I have some unit tests under the `TestSaaSQueryConfig` class and I can add some more as we discover new use cases in the next body of work.

data/saas/config/mailchimp_config.yml
```diff
@@ -20,68 +20,59 @@ saas_config:
...
+              identity: email
+    - name: member
+      requests:
+        read:
+          path: /3.0/search-members
+          request_params:
+            - name: query
+              type: query
+              identity: email
+              data_type: string
```
pattisdr, line 62: Is this data_type supposed to be in this file? it looks like you're putting data_types in the other one.

src/fidesops/service/connectors/query_config.py
```diff
@@ -694,56 +695,83 @@ def get_request_by_action(self, action: str) -> SaaSRequest:
...
+    ) -> SaaSRequestParams:
         """
-        This returns the query/path params needed to make an API call.
-        This is the API equivalent of building the components of a database
-        query statement (select statement, where clause, limit, offset, etc.)
+        Prepares the update request by masking the fields in the row data based on the policy.
+        This masked row is then added as the body to a dynamically generated SaaS request.
         """
+        update_value_map: Dict[str, Any] = self.update_value_map(row, policy, request)
+        body: Row = unflatten_dict(update_value_map)
```
pattisdr, line 748: I wouldn't say the `body` is of type `Row`, since it's probably a subset of the initial "row", so Dict[str, Any] instead.

pattisdr, line 748: Just since we're here, can you move both `update_value_map` and `body` down to be right above your return statement? `return "PUT", path, params, json.dumps(body)`.  Defining them near where you're using them will help separate that they are only related to body, you're using the original row for building the query and the path.

data/saas/config/mailchimp_config.yml
galvana: Removed

src/fidesops/service/connectors/query_config.py
galvana: Changed back to Dict[str, Any]

galvana: Makes sense to keep them together, done!

pattisdr (approved):
Thanks for all your hard work on this one @galvana, approved pending passing tests (formatting)

Issue #228, opened 2022-02-18
[Privacy Center UI] A basic Privacy Center UI, v0


seanpreston:
We want to explore a frictionless deployment "privacy center" solution that takes in privacy requests such as Access requests, Deletion requests, and Rectification requests.

## Requirements:

Step 1: developer user is able to pull down project without having FE deployment knowledge

- ShOPP must be built using Vercel and NextJS
- ShOPP must be trivially deployable
- ShOPP must be a static site with minimal UI configurability

Step 2: developer user is able to do a one-line deployment of the privacy center to a desired location 

- Developer user is able to build the ShOPP application with the command `create-privacy-portal`.  User is shown logs of the process and settings.
- Developer user is able to easily deploy the privacy center following the best practices of NextJS apps.

Step 3: developer user is able to minimally customize certain components on the privacy center after deployment

- User must be able to configure ShOPP application settings with easy to use/maintain file format (ie: JSON or API/web service call). This file must have config options for:
  - Page Title: Default label “Your Privacy Center” 
  - Privacy Actions: Access, Delete, Rectification 
  - Privacy Action Labels: “Download Your Data”, “Delete Your Data”, “Edit Your Data”
  - Privacy Action Icons: point to directory 
  - Identity string input: mapping to a fidesops identity type, select n-number of identity string inputs
  - Policy keys: One for access, one for erasure
  - User must be able to configure custom styling via css that the user can modify or import after deployment
  - User must be able to customize or swap default assets and text strings
  - User must be able to save changes or updates to configurations

```
{
    "title": "my privacy centre",
    "fidesops_host": "https://...",
    "actions": [
        {
            "label": "erase my data",
            "policy_key": "my_erasure_policy",
            "icon_path": "my_icon.png",
            "identity_inputs": ["email", "phone"]
        }, {
            ...
        }
    ],
}
```

Step 4: data subject user enters subject request

- User must be able to visit a publicly deployed webpage
- User must be able to enter some identifying information, such as an email address
- User must be able to select a type of privacy action
- User must be able to enter identifying information (as specified in the identity string input)
- User must be able to submit this request

adriaaaa:
completed and merged

PR #239, opened 2022-03-01
[#228] Implement Privacy Center client app

elliotbonneville:
# Purpose

Implements Next.js / React Privacy Center client app based on the [fidesui-template](https://github.com/ethyca/fidesui-template) repo. [Demo here](https://fidesops-privacy-center-9e9hiofib-elliotbonneville.vercel.app/).

Note: because there is no hosted `fidesops` server any requests made through this UI will fail.

# Changes

- Creates `clients` directory at repo root
- Implements Next.js client app
- Implements testing strategy for Next app

# Checklist

# Ticket

Described by #228


clients/privacy-center/.env.development
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_AUTH_ENDPOINT=http://0.0.0.0:8080/api/v1/oauth/token
```
seanpreston, line 1: It'd also be good to pass this URL down through `config.json`, so anyone hosting this has the option to connect it to a different server should they wish.

clients/privacy-center/README.md
```diff
@@ -0,0 +1,53 @@
...
+Additionally, if you're handy with React and are feeling a little brave you can
+customize the entire application by modifying the TypeScript source code in
+`pages/index.tsx` and `components/RequestModal.tsx`.
+
+## Development
+
+To serve this application locally, run:
+
+```bash
+npm run dev
```
seanpreston, line 27: I believe this assumes that `next` is already installed, we should add explicit instructions to `npm install next` if it's not already installed

clients/privacy-center/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+As a Next application, it will output build artifacts to the `.next` directory.
+
+## Testing
+
+To run the interactive test interface, run:
+
+```bash
+npm run test
```
seanpreston, line 45: Interesting that this is an interactive interface — what's the command here to automatically run all tests? We'll want to add this to the CI jobs too.

elliotbonneville: Great point. I'll add a command for it.

clients/privacy-center/.env.development
elliotbonneville: Yes, this is the point I mentioned on our call the other day – we need to decide where we want environment variables to live for this specific project, because it has an additional config entry point in `config.json`. Either we should specify variables in `config.json` or in a `.env` file, but not both. 

My thought is that we should just remove this file – what do you think?

clients/privacy-center/README.md
elliotbonneville: It doesn't need to be installed globally – it's [a local dependency](https://github.com/elliotbonneville/fidesops/blob/main/clients/privacy-center/package.json#L24), and NPM scripts can pick up on that.

clients/privacy-center/README.md
elliotbonneville: (It's simply `jest`, instead of `jest --watch`)

clients/privacy-center/.env.development
seanpreston: Agree, I think all the config is best off in one place

clients/privacy-center/components/RequestModal.tsx
```diff
@@ -0,0 +1,293 @@
...
+          handleError();
+        }
+      } catch (error) {
+        handleError();
+        return;
+      }
+
+      onClose();
+    },
+    validate: (values) => {
```
seanpreston, line 121: What are your thoughts on invoking this function on form submit, rather than on unfocus of an input, as this results in weird looking scenarios like:

Screenshot 2022-03-10 at 14 15 01

clients/privacy-center/README.md
seanpreston: FWIW I'm only making that comment as when I ran `npm run dev` for the first time on this fork it complained `next` couldn't be found. I'm not sure the local deps are auto-installing in that case?

clients/privacy-center/components/RequestModal.tsx
```diff
@@ -0,0 +1,293 @@
...
+        name?: string;
+        email?: string;
+        phone?: string;
+      } = {};
+
+      if (!values.email && action.identity_inputs.email === 'required') {
+        errors.email = 'Required';
+      } else if (
+        values.email &&
+        !/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values.email)
```
seanpreston, line 133: I'd like for us to get better at documenting regex's in general. This is obviously validating an email address but can we add a comment here breaking down exactly what'll match and what won't? Then if we found this anywhere, a link to the original source.

It looks as though the regex here might not cover all the cases we need too, for example we're not including the tilde `~` character in the set of valid chars.

Screenshot 2022-03-10 at 14 40 31

I understand this particular topic is hard to pin down, Wikipedia's resource looks well informed: https://en.wikipedia.org/wiki/Email_address#Local-part_normalization

I suspect it might be better here to err on the side of letting through emails that _might_ be invalid, since the worst case scenario is that it gets run through Fidesops as a privacy request and if the email is invalid then nothing will be found.

clients/privacy-center/pages/404.tsx
```diff
@@ -0,0 +1,78 @@
+import Head from 'next/head';
+import Image from 'next/image';
+import NextLink from 'next/link';
+import { Stack, Heading, Box, Text, Button } from '@fidesui/react';
+
+const Custom404 = () => (
```
seanpreston, line 6: Great attention to detail @elliotbonneville — I don't think this was in the initial spec was it?

clients/privacy-center/pages/404.tsx
```diff
@@ -0,0 +1,78 @@
...
+                src="/logo.svg"
+                alt="FidesUI logo"
+                width="124px"
+                height="38px"
+              />
+            </Box>
+          </Stack>
+        </Box>
+        <Box display={['none', 'none', 'inherit']}>
+          <Image
```
seanpreston, line 66: Should we make this an anchor link back to the homepage too?

clients/privacy-center/pages/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+              maxWidth="5xl"
+              zIndex={1}
+              position="absolute"
+            >
+              <AlertIcon />
+              <AlertDescription>{alert.description}</AlertDescription>
+              <CloseButton
+                position="absolute"
+                right="8px"
+                onClick={() => setAlert(null)}
```
seanpreston, line 55: Again, not necessary for this PR, since we're hiding UI elements (in this case the header) with this alert, we should have it fade out or auto close after a certain time

seanpreston (changes requested):
Fantastic work so far @elliotbonneville. Couple of comments to address still but we're tantalisingly close!

clients/privacy-center/components/RequestModal.tsx
daveqnet: OWASP's input validation cheatsheet has a pretty good section on email address validation which might be helpful here, @seanpreston?

Link: https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#email-address-validation

Basically, it's surprisingly very difficult to do it right and the extended validation should be offloaded to a mail server if possible, but some basic initial validation could look like this:

* The email address contains two parts, separated with an @ symbol.
* The email address does not contain dangerous characters (such as backticks, single or double quotes, or null bytes).
   * Exactly which characters are dangerous will depend on how the address is going to be used (echoed in page, inserted into database, etc).
* The domain part contains only letters, numbers, hyphens (-) and periods (.).
* The email address is a reasonable length:
   * The local part (before the @) should be no more than 63 characters.
   * The total length should be no more than 254 characters.

NevilleS: I've been baited into this conversation by @seanpreston asking for strong opinions

My strong opinion is that we _shouldn't_ try to do this kind of validation ourselves; `<input type="email">` exists and the browsers handle it...

daveqnet: Hmmmm. Front-end validation is trivial to bypass anyway.

clients/privacy-center/README.md
elliotbonneville: Yep, I've added that command to the readme.

clients/privacy-center/components/RequestModal.tsx
elliotbonneville: Form validation on field blur is a fairly standard pattern, and in fact it's the standard one that Formik (the form library we're using) recommends. I suspect this only looks strange because you're deliberately testing the input – most people will enter something before moving on to the next field. :)

If you'd prefer to validate the form on submission – and keep the "Continue" button active initially instead of having it become active when the form is valid – I can implement that. It would require reworking some of the other logic as well, but shouldn't be a problem. That's also a fairly common pattern.

clients/privacy-center/components/RequestModal.tsx
elliotbonneville: For context, the regex in this PR comes from the Formik docs (Formik is an extremely popular React form management library).

Given @neville's opinion, I'll remove email validation for now, and we can add it back in a future PR if needed.

clients/privacy-center/pages/404.tsx
elliotbonneville: Came in from the app template!

clients/privacy-center/pages/404.tsx
elliotbonneville: Sure, good call. Added.

seanpreston (approved):

seanpreston:
This branch is a fork, so cannot access the secrets, hence the unsafe checks are failing

clients/privacy-center/config/config.json
```diff
@@ -0,0 +1,31 @@
+{
+  "title": "Take control of your data",
+  "description": "When you use our services, you’re trusting us with your information. We understand this is a big responsibility and work hard to protect your information and put you in control.",
+  "fidesops_host_development": "http://localhost:8080/api/v1",
```
seanpreston, line 4: One last thing @elliotbonneville — I think we wanted to add the auth URL here?

seanpreston: Screenshot 2022-03-14 at 17 16 58

Specifically wrt this change

PR #271, opened 2022-03-08
[Admin UI Backend] Adds User Model and Create/Delete User Endpoints 

### LINKED ISSUE #240: [Admin UI] User Model

In order to support a login / logout flow for the Fidesops Admin UI, we'll need a place in the application databases to store user data.

Add a model to the Fidesops admin database:

```
class User:
    id
    username
    password
```

### Additional context

- This is to support user create / delete / login / logout flows
- `password` must be an encrypted field
- Ideally this migration could also add a `ClientDetail.user` foreign key, and `ClientDetail.fides_key`

### LINKED ISSUE #241: [Admin UI] Create + Delete User endpoints

We'll need API endpoints to create and delete users for logging into the Fidesops Admin UI.

```

POST /api/v1/user/

{
  "username": "joebloggs",
  "password": "somethingSecure"
}


DELETE /api/v1/user/<str:uuid>/

```

#### Authentication

- To create the first user, the root token is required.
- To delete a user, the root token or a token belonging to that user is required.

- The nuance of the DX here revolves around how we create the first user. We could also provide a `make` command here that allows us to create the first user for us, which could then be used to create subsequent users.

### Additional context

- Users don't own anything in the system. ie. In the initial versions all users will be able to read anything, and write anything.
- So that we can invalidate tokens at an individual level, each user's token will be issued by their own `ClientDetail`.
- When logging in a `User`, check first for the existence of that `User`'s `ClientDetail` to issue the token, if it does not exist, one must be created.
- When logging out a `User`, we must delete the `ClientDetail` so as to expire the token.

pattisdr:
❗ Contains database migration 

# Purpose

In order to support a login / logout flow for the Fidesops Admin UI, we'll need a place in the application databases to store user data.  We also need API endpoints to create and delete users for logging into the Fidesops Admin UI.  

The first user will be created by a one-time `make user` command that creates a root `FidesopsUser` and a `ClientDetail`.  This user will be able to login (endpoint coming) and get a token that they can then use to create other users.

# Changes
- Create a `fidesopsuser` table ("user" is a reserved keyword in postgres) and a FidesopsUser model
- Add API endpoints to create and delete users `POST /api/v1/user` and `DELETE /api/v1/<user_id>`
  - Users need a token with CREATE_USER perms to create
  - Users need DELETE_USER perms and either must be deleting themselves or be the root user
  - Delete user also deletes linked client if applicable
- Add optional `fides_key` column to `ClientDetail` to store which client is the admin UI root client
- Add optional `username` column to `ClientDetail` to link a client to its related user to enforce that a user can only delete themselves (coming later)
- Updates postman collection with endpoints to create/delete a user
- Adds a `make user` command that will create a root user with a given username/password combo, and create a `ClientDetail` that is the root client for the Admin UI. This user can then login and create other users.
- Updates db diagram with new tables.

# Checklist

- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)) **Database diagram added, but I'd like to add guides with user login/logout PR after more pieces have been ironed out**

# Ticket
Fixes #240 
Fixes #241


docs/fidesops/docs/postman/Fidesops.postman_collection.json
```diff
@@ -66,7 +66,7 @@
...
+							"raw": "[\n    \"client:create\",\n    \"client:update\",\n    \"client:read\",\n    \"client:delete\",\n    \"config:read\",\n    \"connection:read\",\n    \"connection:create_or_update\",\n    \"connection:delete\",\n    \"dataset:create_or_update\",\n    \"dataset:delete\",\n    \"dataset:read\",\n    \"encryption:exec\",\n    \"policy:create_or_update\",\n    \"policy:read\",\n    \"policy:delete\",\n    \"privacy-request:create\",\n    \"privacy-request:read\",\n    \"privacy-request:delete\",\n    \"rule:create_or_update\",\n    \"rule:read\",\n    \"rule:delete\",\n    \"scope:read\",\n    \"storage:create_or_update\",\n    \"storage:delete\",\n    \"storage:read\",\n    \"privacy-request:resume\",\n    \"webhook:create_or_update\",\n    \"webhook:read\",\n    \"webhook:delete\",\n    \"saas_config:create_or_update\",\n    \"saas_config:read\",\n    \"saas_config:delete\",\n    \"user:create\",\n    \"user:delete\"\n]",
```
pattisdr, line 69: Note to self, this wasn't merged correctly

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -0,0 +1,81 @@
...
+    if not (client.fides_key == ADMIN_UI_ROOT or client.username == user.username):
+        raise HTTPException(
+            status_code=HTTP_403_FORBIDDEN,
+            detail=f"Users can only remove themselves, or be the Admin UI Root User.",
+        )
+
+    user_client = ClientDetail.get_by(db, field="username", value=user.username)
+    if user_client:
+        logger.info(f"Deleting client for user '{user_id}'")
+        user_client.delete(db)
```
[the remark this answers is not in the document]
pattisdr: FK was added in edited requirements, addressing now -

src/fidesops/schemas/user.py
```diff
@@ -0,0 +1,43 @@
...
+from pydantic import validator
+
+from fidesops.schemas.base_class import BaseSchema
+
+
+class UserCreate(BaseSchema):
+    """Data required to create a FidesopsUser"""
+
+    username: str
+    password: str
```
pattisdr, line 12: This is sent in the request body, are there benefits to putting this as form data instead, like we do for the `client_id` and `client_secret` in `POST /v1/oauth/token`?

src/fidesops/models/fidesops_user.py
```diff
@@ -0,0 +1,50 @@
...
+            data["password"].encode(config.security.ENCODING),
+            salt.encode(config.security.ENCODING),
+        )
+
+        user = super().create(
+            db,
+            data={
+                "salt": salt,
+                "hashed_password": hashed_password,
+                "username": data["username"],
```
nschoenbrot, line 37: Should the username be encrypted? 

I am thinking someone might use their real name as a username. 
On the Pro Platform we use application level encryption on PII that goes into the DB. But I am not sure if we are carrying over the same security policies from Pro. Or if a username should be treated as PII.

nschoenbrot, line 37: Should the username be encrypted? 

I am thinking someone might use their real name as a username. 
On the Pro Platform we use application level encryption on PII that goes into the DB. But I am not sure if we are carrying over the same security policies from Pro. Or if a username should be treated as PII.

src/migrations/versions/5a966cd643d7_fidesops_user_and_client_detail.py
```diff
@@ -0,0 +1,59 @@
...
+        op.f("ix_fidesopsuser_username"), "fidesopsuser", ["username"], unique=True
+    )
+    op.add_column("client", sa.Column("fides_key", sa.String(), nullable=True))
+    op.add_column("client", sa.Column("user_id", sa.String(), nullable=True))
+    op.create_index(op.f("ix_client_fides_key"), "client", ["fides_key"], unique=True)
+    op.create_unique_constraint(None, "client", ["user_id"])
+    op.create_foreign_key(None, "client", "fidesopsuser", ["user_id"], ["id"])
+
+
+def downgrade():
```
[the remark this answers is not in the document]
pattisdr: Great catch, we start with an automatically generated alembic migration and edit from there.  The automatically generated version didn't specify a name for the FK constraint.  It is able to create it with a sane default, but it needs the constraint name to drop it, thanks -

src/fidesops/models/fidesops_user.py
pattisdr: I'm not sure either, I didn't treat it as PII since the requirements specified just `password` https://github.com/ethyca/fidesops/issues/240, but this may be something that was overlooked.  @seanpreston, @adriaaaa can you weigh in if admin UI usernames should be stored encrypted?

NevilleS:
Question: how would someone use `make user` without cloning the repo? The use case here is if you just use pull `fidesops` from DockerHub and run the image. We should still be able to provision user accounts in that case.

If the answer to that is to use the root oauth client and the user APIs, that'd be a reasonable approach I guess, though it's still a bit difficult. Another option would be to support env variables like `OAUTH_ROOT_USER_NAME` and `OAUTH_ROOT_USER_PASSWORD`?

pattisdr:
@NevilleS the way it was explained to me was that we didn't want hardcoded creds here, that the root user needs to be created server side, we can't inject it in the frontend because it's insecure.  So it didn't seem like we should be creating a root oauth client from the FE and using that - do your new variables have the same problem?

seanpreston:
> Question: how would someone use make user without cloning the repo? The use case here is if you just use pull fidesops from DockerHub and run the image. We should still be able to provision user accounts in that case.

This is a tricky one:
- We can't use the root creds, because that involves embedding those creds into the frontend code, which is a security hole because those creds could be used to do _anything_ else on Fidesops. This matters initially because `User`s have access to all entities, and aren't sandboxed in any way.
- We can't support unauthenticated user creation either, because we've no way of verifying emails (or identities generally) currently.

Env vars are a good idea because the person configuring Fidesops can supply them and they can be fed directly into the user creation logic on startup. Defaults here are a potential attack vector though, so we should avoid using those. Of course, if we don't use defaults we can't satisfy just running the repo straight from Dockerhub.

The nuance for security is that creating the initial user _has_ to happen server-side in order to be secure, and it also _has_ to require user input. It doesn't necessarily need to be `Make` that's executed however.

src/fidesops/models/fidesops_user.py
pattisdr: discussed in person, we'll not encrypt usernames for now, but a change made to not log them, logging user id instead

pattisdr:
@nschoenbrot comments addressed!

nschoenbrot (approved):

PR #279, opened 2022-03-10
Adds `include_identities` query param to `GET /api/v1/privacy-request`

### LINKED ISSUE #274: [Admin UI] Add option to Privacy Request status API view to return identity data

The Fidesops Admin UI is required to both return, and filter based on, the privacy request's identity data.

To facilitate this we need to make the following changes:

- [ ] Change the sane-default identity cache TTL to be days not hours
- [ ] Add `include_identity=<true / false>` to the query params for `/api/v1/privacy-request/`
- [ ] Return the identity data from the cache in the response

### Questions

- Should we add an extra scope `PRIVACY_REQUEST_IDENTITY_READ` especially for this param?

seanpreston:
# Purpose

For the Fidesops Admin UI we need to include the identities of the privacy requests displayed in the dashboard.

# Changes

- [x] Change the sane-default identity cache TTL to be days not hours
- [x] Add include_identity=<true / false> to the query params for /api/v1/privacy-request/
- [x] Return the identity data from the cache in the response

# Checklist

# Ticket

Fixes #274 
 

## Follow-up

- On implementation of this ticket I've discovered that we actually aren't encrypting the raw identity data in the cache, only the data that's returned from the traversal is encrypted by default. I've added this follow-up ticket to address that: https://github.com/ethyca/fidesops/issues/280


src/fidesops/schemas/privacy_request.py
```diff
@@ -82,6 +82,7 @@ class PrivacyRequestResponse(BaseSchema):
     finished_processing_at: Optional[datetime]
     status: PrivacyRequestStatus
     external_id: Optional[str]
+    identity: Optional[Dict[str, str]]
```
pattisdr, line 85: We could also have the type be the already defined `PrivacyRequestIdentity` schema for this too.  It makes the responses uniform, so if the identity doesn't exist, its value is null - 

```json
"identity": {
                "phone_number": null,
                "email": "customer-1@example.com"
            }
```

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -297,7 +303,14 @@ def get_request_status(
     else:
         PrivacyRequest.execution_logs_by_dataset = property(lambda self: None)
 
-    return paginate(query, params)
+    paginated = paginate(query, params)
+    if include_identities:
+        # Conditionally include the cached identity data in the response if
+        # it is explicitly requested
+        for item in paginated.items:
+            item.identity = item.get_cached_identity_data()
```
pattisdr, line 311: `paginate` also loops through every the every element in the result set before creating the page, and then we are turning around and looping through them again.


```python
def paginate(query: Query, params: Optional[AbstractParams] = None) -> AbstractPage:
    params = resolve_params(params)

    total = query.count()
    items = [_to_dict(item) for item in paginate_query(query, params)]

    return create_page(items, total, params)
```

But we are just looping through fifty elements twice, so not a big deal. Another option is having a custom paginate method here that is similar to the above that also adds the identity data, killing two birds.

pattisdr (commented):
Just minor comments for consideration, works well @seanpreston!

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
seanpreston: IMO it's best we keep these separate so it's clear we're adding this data outside of the scope of our DB and pagination functionality (given it's a very small number of iterations). Where this could prove problematic is with the CSV download functionality, where pagination isn't definitely applied.

src/fidesops/schemas/privacy_request.py
seanpreston: I've added a comment here instead. My rationale is that we'd rather not expose the potential internals of our identity data structure on data egress, given this is an API response. ie. if we use this schema then we'd end up with
```
{"email": "email@example.com", "phone_number": null}
```
for responses where we have an `email` but no `phone_number`, and if we don't have the value it's better we don't expose it at all.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: sounds good

src/fidesops/util/oauth_util.py
```diff
@@ -109,7 +109,7 @@ def is_callback_token_expired(issued_at: Optional[datetime]) -> bool:
 
     return (
         datetime.now() - issued_at
-    ).total_seconds() / 60.0 > config.redis.DEFAULT_TTL_SECONDS
+    ).total_seconds() / 60.0 > config.execution.PRIVACY_REQUEST_DELAY_TIMEOUT
```
pattisdr, line 112: Thanks for making this change! Can you update the docs at the bottom of this page:  https://ethyca.github.io/fidesops/guides/policy_webhooks/

`The reply-to-token expires when the redis cache expires (config.redis.DEFAULT_TTL_SECONDS).`

docs/fidesops/docs/guides/policy_webhooks.md
```diff
@@ -240,5 +240,5 @@ If there are no derived identities, an empty `{}` request body will suffice.
 
 The `reply-to-token` is a JWE containing the current webhook id, scopes to access the callback endpoint, 
 and the datetime the token is issued.  We unpack this and resume the privacy request execution after the 
-specified webhook. The `reply-to-token` expires when the redis cache expires (`config.redis.DEFAULT_TTL_SECONDS`).
+specified webhook. The `reply-to-token` expires when the redis cache expires (`config.execution.PRIVACY_REQUEST_DELAY_TIMEOUT`).
```
pattisdr, line 243: the former bit isn't accurate either!  "when the redis cache expires"

pattisdr (approved):

docs/fidesops/docs/guides/policy_webhooks.md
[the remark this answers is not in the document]
pattisdr: thanks for addressing

pattisdr:
Looks good!

PR #281, opened 2022-03-10
Add `configure-python` step to `run_infrastructure.py` `Make` jobs

seanpreston:
# Purpose

@adriaaaa and @conceptualshark both ran into an issue earlier when running `make quickstart` where `virtualenv` was not installed (see: https://ethyca.slack.com/archives/C029HFVSC8L/p1646935020342749). We need to be sure that the Python environment is consistent for any Make commands that will invoke `run_infrastructure.py`

# Changes

- Adds `configure-python` PHONY dep. to be called before any Make jobs that use `run_infrastructure.py` so that we can be sure the Python environment includes `virtualenv`


seanpreston:
Closing this to change tack here.

PR #283, opened 2022-03-14
272- adds support for custom http methods

### LINKED ISSUE #272: [SaaS Connector] Add config option to specify HTTP method for update requests

We need a way to specify HTTP method (PUT / PATCH / POST) for update requests.

Add a `method` field to the request.read and request.update object in the schema. Add logic to handle the diff methods. Default to most common method, e.g. GET for read requests, PUT for update requests.

### Additional context

Hubspot, for example, will require PATCH update requests- https://ethyca.atlassian.net/wiki/spaces/OSS/pages/2212135123/Hubspot+Connector+Specification

eastandwestwind:
# Purpose
Adds SaaS config option and support for custom HTTP Methods (e.g. `POST`, `PATCH`)

# Changes
- Adds logic in query config to grab HTTP method from saas config, with defaults as `GET`  for `read` and `PUT` for `update` requests, respectively
- Adds query config test for custom HTTP method
- Refactors to make `SaaSRequestParams` a schema

# Checklist

# Ticket

Fixes #272


src/fidesops/schemas/saas/saas_config.py
```diff
@@ -63,19 +63,36 @@ class Strategy(BaseModel):
...
+
+
 class SaaSRequest(BaseModel):
     """
     A single request with a static or dynamic path, and the request params needed to build the request.
     Also includes optional strategies for pre/postprocessing and pagination.
     """
 
     path: str
+    method: HTTPMethod
```
eastandwestwind, line 83: note to self-make Optional

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -63,19 +63,36 @@ class Strategy(BaseModel):
...
 
 
+class HTTPMethod(Enum):
+    """Enum to represent HTTP Methods"""
+
+    GET = "GET"
+    POST = "POST"
+    PATCH = "PATCH"
+    PUT = "PUT"
+    DELETE = "DELETE"
```
seanpreston, line 73: Should we add `OPTIONS` and `HEAD` here too? I can see those being useful for further connectors.

src/fidesops/service/connectors/query_config.py
```diff
@@ -735,7 +735,12 @@ def generate_query(
                 path = path.replace(f"<{param.name}>", input_data[param.name][0])
 
         logger.info(f"Populated request params for {current_request.path}")
-        return "GET", path, params, None
+        method: HTTPMethod = (
+            current_request.method.value
+            if current_request.method
+            else HTTPMethod.GET.value
+        )
+        return method, path, params, None
```
seanpreston, line 743: It would be useful here if we could explicitly set `body = None` earlier in the method, then return `body` here. Having the var there keeps it readable without going away to lookup what `SaaSRequestParams` is.

Can we also make `SaaSRequestParams` a schema in its own right?

tests/fixtures/saas_fixtures.py
```diff
@@ -119,6 +119,27 @@ def connection_config_saas_without_saas_config(
     connection_config.delete(db)
 
 
+@pytest.fixture(scope="function")
+def connection_config_saas_with_custom_http_method(
+        db: Session,
+) -> Generator:
+    saas_config = example_saas_configs["mailchimp"].copy()
+    saas_config["endpoints"][0]["requests"]["read"]["method"] = "POST"
```
seanpreston, line 127: Are we going to use this fixture elsewhere? If not I think we're better off making this line change in `test_generate_update_stmt_custom_http_method` below itself, that way we can set `"POST"` as a variable and assert both values are that variable? Otherwise can we use a constant var here? That way if the method ever changes we can change these two disconnected locations in one go.

galvana: The `main` branch now has the Mailchimp config separate from a newly created `saas_example_config.yml`. I planned on this example config being the one we can add these new use cases to.

src/fidesops/service/connectors/query_config.py
galvana: Hey @seanpreston, I was thinking the same thing about `SaaSRequestParams` it's getting a bit clunky in my use cases too. Can we ticket this as a separate enhancement to not hold up the custom HTTP methods?

tests/fixtures/saas_fixtures.py
eastandwestwind: I'll use `saas_example_config.yml` here too, after pulling latest main.

src/fidesops/schemas/saas/shared_schemas.py
```diff
@@ -1,5 +1,29 @@
...
-"""Custom type to represent a tuple of HTTP method, path, params, and body values for a SaaS request"""
+class HTTPMethod(Enum):
+    """Enum to represent HTTP Methods"""
+
+    GET = "GET"
+    POST = "POST"
+    PATCH = "PATCH"
+    PUT = "PUT"
+    DELETE = "DELETE"
+    OPTIONS = "OPTIONS"
```
seanpreston, line 14: Thanks!

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -80,7 +81,12 @@ def generate_query(
                 path = path.replace(f"<{param.name}>", input_data[param.name][0])
 
         logger.info(f"Populated request params for {current_request.path}")
-        return "GET", path, params, None
+        method: HTTPMethod = (
+            current_request.method if current_request.method else HTTPMethod.GET
```
seanpreston, line 85: Here and below feel like they're places too nested within the traversal to be setting defaults.

I don't think the request's method is ever really optional, so we should probably force the specification of it every time to be explicit and standard across the language.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -118,7 +124,12 @@ def generate_update_stmt(
 
         update_value_map: Dict[str, Any] = self.update_value_map(row, policy, request)
         body: Dict[str, Any] = unflatten_dict(update_value_map)
-        return "PUT", path, params, json.dumps(body)
+        method: HTTPMethod = (
+            current_request.method if current_request.method else HTTPMethod.PUT
```
seanpreston, line 128: As above

src/fidesops/schemas/saas/shared_schemas.py
```diff
@@ -1,5 +1,29 @@
...
+    HEAD = "HEAD"
+
+
+class SaaSRequestParams(BaseModel):
+    """Custom type to represent a SaaS request param"""
+
+    method: HTTPMethod
+    path: str
+    param: Dict[str, Any]
```
seanpreston, line 23: Ultra nit: can this be called `params`

(was this because your `s` key is broken? 😂)

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -73,11 +76,18 @@ class SaaSRequest(BaseModel):
     """
 
     path: str
+    method: Optional[HTTPMethod]
```
seanpreston, line 79: As mentioned, this isn't an optional field in a HTTP request and introduces a potentially confusing default setting problem.

src/fidesops/schemas/saas/shared_schemas.py
```diff
@@ -1,5 +1,29 @@
...
+    body_values: Optional[str]
```
seanpreston, line 24: Also nit: should this be `body_value` or just `body` since it's a JSON string that is one value in this context.

seanpreston (changes requested):
Just a couple more small things to address but very close! Thanks very much for making the `SaaSRequestParams` change here!

src/fidesops/schemas/saas/shared_schemas.py
eastandwestwind: it might've been 😢

src/fidesops/service/connectors/saas_query_config.py
eastandwestwind: it is optional in the sense that we have defaults if not specified: `GET`  for `read` and `PUT` for `update` requests

eastandwestwind: should we change this to be explicit across all configs instead?

eastandwestwind: this will be follow-up ticket

eastandwestwind: https://github.com/ethyca/fidesops/issues/305

seanpreston (approved):
Thanks @eastandwestwind

PR #284, opened 2022-03-14
163 yaml datasets

### LINKED ISSUE #163: Support YAML format for configuring datasets on fidesops

Today, the dataset API (see docs for [PATCH /api/v1/connection/{connection_key}/dataset](https://ethyca.github.io/fidesops/api/#operations-Datasets-patch_datasets_api_v1_connection__connection_key__dataset_patch)) only supports `application/json` as a format, which doesn't make a lot of sense as we encourage fideslang datasets to be written and stored as YAML files. It's a small bit of tech debt that I left in there when coding up the dataset API, but neglected to document it anywhere so right now it's up to the engineer to convert YAML -> JSON before uploading to fidesops.

We should simply support either JSON or YAML as a request body, and then otherwise behave exactly the same. It'd be reasonable to support this for *all* endpoints, but since I'd like to see explicit tests as much as possible, I'd probably limit it to just datasets for today (and maybe policies later...)

We can, of course, document this and continue to only allow JSON. It's a "gotcha" though and worth ironing out soon, IMO.

nschoenbrot:
# Purpose

Allow datasets to be patched using YAML in place of JSON.

# Changes

-Creates a YAML path for patching datasets. Since FastAPI does not have built in support for YAML in the way it does for JSON, the YAML endpoint will directly access the request object and deserialize the body. But the JSON and YAML paths will still hit the same underlying code. (I wanted to try and route based on content type so that the same path could be shared, from what I found content type based routing has not be built into FastAPI yet)
-Was able to get a successful response from the YAML Datasets endpoint locally.
~~-TODO Add more tests to cover edge cases such as bad content type, bad YAML file, multiple items listed under the YAML `dataset`.~~
~~-TODO Fix type checking issues found by mypy. There are typing issues around the deserialized request body.~~
~~-TODO Update postman collection.~~
~~-TODO Come up with a strategy on how to generate custom OpenAPI spec for YAML endpoints.~~

# Checklist

# Ticket

Fixes #163


nschoenbrot:
@seanpreston It looks like I will be working on this ticket again during this sprint. 

For this ticket how do we want to handle documentation for the YAML endpoints? We could just leave it out of the Open API spec, or start manually writing the Open API spec? See: https://fastapi.tiangolo.com/advanced/extending-openapi/

including @conceptualshark since there is a documentation question here.

seanpreston:
> We could just leave it out of the Open API spec

Great suggestion. Let's ship this as a v0, then follow-up with manually adding the OpenAPI spec as you've suggested as part of a subsequent ticket.

docs/fidesops/docs/postman/Fidesops.postman_collection.json
```diff
@@ -10,6 +10,17 @@
 			"item": [
 				{
 					"name": "Get Root Client Token",
+					"event": [
+						{
+							"listen": "test",
+							"script": {
+								"exec": [
+									"pm.collectionVariables.set(\"client_token\", pm.response.json()[\"access_token\"])"
+								],
```
nschoenbrot, line 19: This script will take the access token from the response and update the collection variable. That way we don't need to manually copy and paste.

src/fidesops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -166,35 +176,95 @@ def patch_datasets(
...
+                db,
+                failed,
+            )
     return BulkPutDataset(
         succeeded=created_or_updated,
         failed=failed,
     )
 
 
+def create_or_update_dataset(
```
nschoenbrot, line 236: This function was extracted from the Patch dataset function. So that both dataset PATCHing endpoints (JSON and YAML) can call the same underlying code.

src/fidesops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -166,35 +176,95 @@ def patch_datasets(
...
+    if request.headers.get("content-type") != X_YAML:
+        raise HTTPException(
+            status_code=HTTP_415_UNSUPPORTED_MEDIA_TYPE,
+            detail="Supported type: " + X_YAML,
+        )
+    body = await request.body()
+    try:
+        yaml_request_body: dict = yaml.safe_load(body)
+    except yaml.YAMLError:
+        raise HTTPException(status_code=HTTP_400_BAD_REQUEST, detail="Error in YAML")
```
galvana, line 211: Is there anything useful from `yaml.YAMLError` that we can include in the response to help fix any YAML errors?

src/fidesops/api/v1/endpoints/dataset_endpoints.py
```diff
@@ -166,35 +176,95 @@ def patch_datasets(
...
+
 
+@router.patch(
+    YAML_DATASETS,
+    dependencies=[Security(verify_oauth_client, scopes=[DATASET_CREATE_OR_UPDATE])],
+    status_code=200,
+    response_model=BulkPutDataset,
+    include_in_schema=False  # Not including this path in the schema.
+    # Since this yaml function needs to access the request, the open api spec will not be generated correctly.
+    # To include this path, extend open api: https://fastapi.tiangolo.com/advanced/extending-openapi/
```
galvana, line 195: This is fine for now but we should find a way to expose this information in our Swagger API UI if we add additional YAML endpoints. Can we document this endpoint somewhere if we haven't already?

nschoenbrot: I brought this up with @seanpreston & @conceptualshark. We can have a subsequent ticket. 

I would think our solution would depend on how many resources and actions we will support in the yaml API. For me there are open questions: 

- How much will the JSON and YAML APIs diverge? 
- Is it going to be a large API? If so we might want to find a tool to autogenerate open api spec. 
- Do we want a separate spec for the yaml api?

src/fidesops/api/v1/endpoints/dataset_endpoints.py
nschoenbrot: I pushed up a change to append the string version of the error.

galvana (approved):
Thanks for making the change to help with future troubleshooting, looks good to me!

src/fidesops/api/v1/urn_registry.py
```diff
@@ -68,6 +69,9 @@
 DATASETS = CONNECTION_BY_KEY + "/dataset"
 DATASET_BY_KEY = CONNECTION_BY_KEY + "/dataset/{fides_key}"
 
+# YAML Collection URLs
+YAML_DATASETS = YAML + DATASETS
```
seanpreston, line 73: Nit: This will result in the URL `/api/v1/yml/connection/{key}/dataset` — could we append `YAML` here instead of prepend for the result `/api/v1/connection/{key}/dataset/yml/`, to reflect the hierarchy of what's being executed.

Issue #291, opened 2022-03-16
[Saas Connector Engine][Fidesops Core] Configure "After" Param for Saas Configs and Dependent Identities


pattisdr:
❗ Blocks #256 

# What
Allow an endpoint with identity data to be specified that it needs to wait on all incoming inputs before visiting (for saas configs).  

# Why

Currently, when we build the graph of all nodes we need to visit, we create a root node, and its children are any nodes with identity data.    Other nodes wait for all of their upstream nodes to have been visited, but nodes with identity data are treated differently.  We need to be able to specify that an identity node should be sometimes be visited only after every other input is fulfilled.

## Use case

A saas connector needs a `project_id` and an `organization_id` from one endpoint, and needs to use the identity data as a query param.  We can't attempt to visit this endpoint first, because we don't yet have the `project_id` and `organization_id`.

## Alternatives considered

- There's an `after` param that can be specified on a collection, look into using that.  On a quick glance though, it seems it's a little more of an involved fix.
- We could also use a postprocessor to filter the data with the identity after the fact, but  I don't view this as a viable solution because of the enormous amount of extra queries we'd need to make.

NevilleS:
For my understanding: basically our traversal currently assumes that any node that relies on `identity` data can be visited immediately, but we need it to handle the case where the node requires both `identity` data and some other `references`?

pattisdr:
@NevilleS, correct, yes.

NevilleS:
Awesome- that makes a ton of sense and I'm following along. Thanks!

pattisdr:
Note that after some investigation, the `after` param works well as far as postponing when an identity node is featured, so we primarily need to add some handling for being able to specify this on the saas config side, which is a small lift by itself.

To get this entirely working for sentry though, this identity email is also a "grouped input" field, in that we actually need to combine this with #290.

PR #299, opened 2022-03-17
Fidesops 290 grouped inputs

pattisdr:
# Purpose

# Changes

-
# Checklist

# Ticket

Fixes #


src/fidesops/task/graph_task.py
```diff
@@ -158,22 +200,43 @@ def pre_process_input_data(self, *data: List[Row]) -> NodeInput:
...
+                    grouped_data: Dict[str, Any] = {}
+                    for (
+                        foreign_field_path,
+                        local_field_path,
+                    ) in dependent_field_mappings[collection_address]:
+                        dependent_values: List = consolidate_query_matches(
+                            row=row, target_path=foreign_field_path
+                        )
+                        grouped_data[local_field_path.string_path] = dependent_values
+                    output[FIDESOPS_GROUPED_INPUTS].append(grouped_data)
```
pattisdr, line 239: This is a change to the fidesops core, grouped fields can be separated out under a key that shouldn't overlap with a field on the user's collection/api endpoint.  This change should be added here rather than downstream because this is where we have the original rows of data from the upstream collections, and we can make sure two fields from a single incoming row are kept together.  Additionally, this opens it up to help build more extensive "AND" queries on our databases in the future, not just saas configs.

As a first step though, we're just setting up the ability to specify this on a saas config, but putting the foundation here makes it extensible.

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -93,6 +95,36 @@ def validate_request_for_pagination(cls, values: Dict[str, Any]) -> Dict[str, An
...
+                raise ValueError(
+                    "Grouped_input fields must also be declared as request_params."
+                )
+
+            referenced_collections: List[str] = []
+            for param in request_params:
+                if param.name in grouped_inputs:
+                    if not param.references:
+                        raise ValueError(
+                            "Grouped input fields must be on incoming references."
```
pattisdr, line 117: Note to self, identity data should be allowed here too, this is surfacing in follow-up https://github.com/ethyca/fidesops/issues/291

pattisdr: Addressed in https://github.com/ethyca/fidesops/pull/299/commits/fac3b48ac2463a94861e1301afceb78134dff8d8

src/fidesops/graph/config.py
```diff
@@ -394,6 +394,8 @@ class Collection(BaseModel):
     fields: List[Field]
     # an optional list of collections that this collection must run after
     after: Set[CollectionAddress] = set()
+    # An optional list of dependent fields that need to be queried together
+    grouped_inputs: Optional[List[str]] = []
```
seanpreston, line 398: Should we make this a `Set[str]` so that the inputs are de-duped?

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -82,6 +81,7 @@ class SaaSRequest(BaseModel):
     data_path: Optional[str]
     postprocessors: Optional[List[Strategy]]
     pagination: Optional[Strategy]
+    grouped_inputs: Optional[List[str]] = []
```
seanpreston, line 84: As above re: `Set[str]`

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -104,6 +104,37 @@ def validate_request_for_pagination(cls, values: Dict[str, Any]) -> Dict[str, An
...
+        """Validate that grouped_inputs must reference fields from the same collection"""
+        grouped_inputs = values.get("grouped_inputs")
+
+        if grouped_inputs:
+            request_params = values.get("request_params", [])
+            names = [param.name for param in request_params]
+
+            if not all(field in names for field in grouped_inputs):
+                raise ValueError(
+                    "Grouped_input fields must also be declared as request_params."
```
seanpreston, line 118: Ultra nit: this should read `Grouped input` or `grouped_input`

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -104,6 +104,37 @@ def validate_request_for_pagination(cls, values: Dict[str, Any]) -> Dict[str, An
...
+    @root_validator
+    def validate_grouped_inputs(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        """Validate that grouped_inputs must reference fields from the same collection"""
+        grouped_inputs = values.get("grouped_inputs")
+
+        if grouped_inputs:
+            request_params = values.get("request_params", [])
+            names = [param.name for param in request_params]
+
+            if not all(field in names for field in grouped_inputs):
```
seanpreston, line 116: If we used sets here we could use a clearer `if not names.issubset(grouped_inputs): ...`

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -104,6 +104,37 @@ def validate_request_for_pagination(cls, values: Dict[str, Any]) -> Dict[str, An
...
+            for param in request_params:
+                if param.name in grouped_inputs:
+                    if not param.references and not param.identity:
+                        raise ValueError(
+                            "Grouped input fields must be on incoming references or identity fields."
+                        )
+                    if param.references:
+                        collect = param.references[0].field.split(".")[0]
+                        referenced_collections.append(collect)
+            if not all(x == referenced_collections[0] for x in referenced_collections):
```
seanpreston, line 131: I'm not sure if this one is just preference, something like `len(set(referenced_collections)) > 1` might be more readable

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -87,6 +87,14 @@ def generate_requests(
                 request_params.append(
                     self.generate_query({string_path: [value]}, policy)
                 )
+
+        # Build SaaS requests for fields that are dependent on each other
+        grouped_inputs: List[Dict[str, Any]] = (
+            input_data.get(FIDESOPS_GROUPED_INPUTS) or []
```
seanpreston, line 93: Is there a reason not to use `input_data.get(FIDESOPS_GROUPED_INPUTS, [])` here?

src/fidesops/task/graph_task.py
```diff
@@ -98,19 +100,16 @@ def __init__(
         )
 
         # build incoming edges to the form : [dataset address: [(foreign field, local field)]
-        b: Dict[CollectionAddress, List[Edge]] = partition(
```
seanpreston, line 101: Thankyou for fixing the var naming here 🙏

src/fidesops/task/graph_task.py
```diff
@@ -158,22 +200,43 @@ def pre_process_input_data(self, *data: List[Row]) -> NodeInput:
...
 
         output: Dict[str, List[Any]] = {}
+
+        (
+            independent_field_mappings,
+            dependent_field_mappings,
+        ) = self.build_incoming_field_path_maps(group_dependent_fields)
+
+        if group_dependent_fields:
+            output[FIDESOPS_GROUPED_INPUTS] = []
```
seanpreston, line 210: It feels brittle to use a whole arg just to initialise this var. This would imply that `dependent_field_mappings[collection_address]` referenced on line 229 below will always be Falsy if the `group_dependent_fields` var is `False`. Is it safer / feasible to initialise this var unconditionally?

seanpreston (changes requested):
Thanks @pattisdr — couple of small readability nits but very well constructed and the concepts are very clearly presented nonetheless!

pattisdr:
thanks for the review @seanpreston  ⭐

src/fidesops/task/graph_task.py
pattisdr: My thinking was that I didn't want to clutter the output for the `group_dependent_fields=False` case - this method is called twice, once to process incoming data into a collection, and the other to process outputs from a collection that are stored in the cache.   Also, the implication you mention above is true, if group_dependent_fields is False, it causes the built dependent_field_mappings for all collection addresses to always be empty so this seems safe to me. 

However, I agree we can add this key in both cases. It is removed in downstream processing because it's not an actual field on a collection, so it won't persist through to the results we send to the user.

src/fidesops/graph/config.py
pattisdr: sure, i can make it a set, but I'll still have it be input as a list, and transformed to a set before persisting here.

src/fidesops/schemas/saas/saas_config.py
pattisdr: thanks!

src/fidesops/schemas/saas/saas_config.py
pattisdr: Do you not like the `all` function? length checking seems an older way to do this, but it's not a big deal to change.

src/fidesops/service/connectors/saas_query_config.py
pattisdr: no just missed in cleanup!

src/fidesops/schemas/saas/saas_config.py
seanpreston: It's not `all()` I dislike here, the comprehension took me a couple reads to grok. Then again it was late at night, so happy if you'd rather leave it in

seanpreston (approved):

src/fidesops/schemas/saas/saas_config.py
pattisdr: definitely appreciate second opinions on readability, changed it to length, thanks!

Issue #256, opened 2022-03-02
[SaaS Connector] Sentry


seanpreston:
❗ Dependent on #291 and #290 

Create a SaaS connector for Sentry's API.

[Requirements](https://ethyca.atlassian.net/wiki/spaces/OSS/pages/2212102280/Sentry+Connector+Specification)

PR #306, opened 2022-03-20
Support Dependent Identity Data and "After" Param in Saas Configs [#291]

### LINKED ISSUE #291: [Saas Connector Engine][Fidesops Core] Configure "After" Param for Saas Configs and Dependent Identities

pattisdr:
❗ Dependent on #299 

# Purpose

In Fidesops, we can specify that a collection should be forced to run after another collection.  Likewise, we need to be able to specify that a Saas Config endpoint is run "after" another collection or endpoint. This is to support the use case where an endpoint with an identity data needs to wait for a response from another endpoint before running.

Furthermore, as a follow-up to #299, where we allowed inputs into a collection to be grouped together, we need to support grouping identity data as well.  For example, a saas config endpoint needs to wait for two references from one endpoint, and also needs identity data before proceeding. These three pieces of information need to stay grouped together.

# Changes
- Support the "after" param on a Saas Config endpoint
- If an identity field is listed alongside collection references in `grouped_inputs`, combine the identity data with the other grouped_inputs in `GraphTask.pre_process_input_data` before passing that data onto the endpoint.

# Checklist

- [x] Good unit test/integration test coverage **(Note: end-to-end testing will be completed with Sentry #256 )**

# Ticket

Fixes #291


pattisdr:
Addressing test failures ^

pattisdr:
Rebasing ^

pattisdr:
Getting this up-to-date after saas config headers merge ^

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -214,6 +222,10 @@ def get_graph(self) -> Dataset:
                         name=endpoint.name,
                         fields=fields,
                         grouped_inputs=grouped_inputs,
+                        after={
+                            CollectionAddress(*s.split("."))
+                            for s in endpoint.after or []
```
seanpreston, line 227: `endpoint.after` will default to `[]` so we don't need `or []` unless you think it helps readability

src/fidesops/task/graph_task.py
```diff
@@ -128,6 +128,15 @@ def grouped_fields(self) -> Set[Optional[str]]:
...
         return self.traversal_node.node.collection.grouped_inputs or set()
 
+    @property
+    def dependent_identity_fields(self) -> bool:
+        """If the current collection needs inputs from other collections, in addition to its seed data."""
+        collection = self.traversal_node.node.collection
+        grouped_fields = [
+            collection.field(FieldPath(field)) for field in self.grouped_fields
+        ]
+        return any(field.identity for field in grouped_fields)
```
seanpreston, line 138: We could save some iterations here using a for loop:

```
for field in grouped_fields:
    if collection.field(FieldPath(field)).identity:
        return True
```

then we also wouldn't need the overlap of `grouped_fields` and `self.grouped_fields`, where one is a different format.

src/fidesops/task/graph_task.py
```diff
@@ -168,6 +177,27 @@ def can_write_data(self) -> bool:
         connection_config: ConnectionConfig = self.connector.configuration
         return connection_config.access == AccessLevel.write
 
+    def _combine_seed_data(
+        self,
+        *data: List[Row],
+        grouped_data: Dict[str, Any],
+        dependent_field_mappings: COLLECTION_FIELD_PATH_MAP,
+    ) -> Dict[str, Any]:
+        """Combine the seed data with the other dependent inputs. This is used when seed data into a collection requires
```
seanpreston, line 186: Should read `...when seeding data...`

tests/task/test_graph_task.py
```diff
@@ -249,15 +249,14 @@ def test_pre_process_input_conversation_collection(
         }
 
     def test_pre_process_input_data_group_dependent_fields(self):
-        """Test processing inputs where fields have been marked as dependent"""
+        """Test processing inputs where fields have been marked as dependent
+        Identity data has also been marked as being dependent.
```
seanpreston, line 253: Should this read `Test processing inputs where identity data has also been marked as being dependent.`?

seanpreston (approved):
Awesome work @pattisdr — just a couple of minor suggestions here.

src/fidesops/task/graph_task.py
pattisdr: hm I meant it as a noun, maybe adding `the` will help? This is used when the seed data in a collection requires inputs from another collection to generate subsequent queries.

tests/task/test_graph_task.py
pattisdr: Well it's testing both, that reference fields and an identity field have all been marked together as a group, I'll rephrase!

src/fidesops/task/graph_task.py
pattisdr: Good idea

src/fidesops/schemas/saas/saas_config.py
pattisdr: Ah I added this for mypy, I think if I remove the Optional from endpoint.after it should work.

pattisdr:
Thanks for the review @seanpreston, ready for another pass!

PR #309, opened 2022-03-21
Returning an empty list instead of None when post-processors filter a…

### LINKED ISSUE #308: Post-processors returning 'None' will cause the retrieve_data function to retry until failure
### Bug Description

If the final result of endpoint post-processing is not a list or a dict (for example None) `retrieve_data` will throw a `FidesopsException` which will cause the graph task to retry again. We have two valid cases where we don't expect data to be returned after post-processing:

* when a post-processor correctly filters out all data passing through
* when the data path for an unwrap post-processor does not exist within the data being unwrapped

It would be more accurate for post-processors to return an empty list when no data is found since an empty list can be processed by subsequent post-processors and by the overall `process_response_data` function in `saas_connectors` without any exceptions.

### Steps to Reproduce

1. Setup the Mailchimp connector.
2. Submit a privacy request with an email identity that does not contain messages in Mailchimp.
3. The graph task will reach the max number of retries before failing and returning the access request results.

### Expected behavior

The messages endpoint should only be called once since an empty result from the `filter` post-processor for the `messages` endpoint is a scenario that should not throw an exception.

galvana:
# Purpose

To fix the issue described in #308.

# Changes

Modified the filter and unwrap post-processors to return an empty list `[]` for the following scenarios:

### Filter
- Empty data passed into filter
- Identity data missing
- Field not available on the object being filtered
- No results remaining after filter (how this issue was originally discovered)

### Unwrap
- No data found at `data_path`

# Checklist

# Ticket

Fixes #308

### DIFF
--- a/src/fidesops/service/processors/post_processor_strategy/post_processor_strategy.py
+++ b/src/fidesops/service/processors/post_processor_strategy/post_processor_strategy.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Any, Dict
+from typing import Any, Dict, List, Union
 
 
 class PostProcessorStrategy(ABC):
@@ -10,5 +10,7 @@ class PostProcessorStrategy(ABC):
         """Returns strategy name"""
 
     @abstractmethod
-    def process(self, data: Any, identity_data: Dict[str, Any] = None) -> Any:
+    def process(
+        self, data: Any, identity_data: Dict[str, Any] = None
+    ) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
         """Process data from SaaS connector"""
--- a/src/fidesops/service/processors/post_processor_strategy/post_processor_strategy_unwrap.py
+++ b/src/fidesops/service/processors/post_processor_strategy/post_processor_strategy_unwrap.py
@@ -59,16 +59,20 @@ class UnwrapPostProcessorStrategy(PostProcessorStrategy):
         if isinstance(data, dict):
             unwrapped = pydash.get(data, self.data_path)
             if unwrapped is None:
-                logger.warning(f"{self.data_path} could not be found for the following "
-                    f"post processing strategy: {self.get_strategy_name()}")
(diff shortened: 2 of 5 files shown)


src/fidesops/service/processors/post_processor_strategy/post_processor_strategy_unwrap.py
```diff
@@ -48,30 +48,34 @@ def process(
...
+                if unwrapped is None:
+                    logger.warning(f"{self.data_path} could not be found for the following "
+                    f"post processing strategy: {self.get_strategy_name()}")
+                else:
+                    result.append(unwrapped)
             # flatten the list to account for the event where the output of unwrapped
             # is a list of lists
-            unwrapped = pydash.flatten(unwrapped)
-
-        if unwrapped is None:
```
galvana, line 69: Moved this inside for loop to log a warning for each occurrence.

src/fidesops/service/processors/post_processor_strategy/post_processor_strategy.py
```diff
@@ -10,5 +10,7 @@ def get_strategy_name(self) -> str:
         """Returns strategy name"""
 
     @abstractmethod
-    def process(self, data: Any, identity_data: Dict[str, Any] = None) -> Any:
+    def process(
+        self, data: Any, identity_data: Dict[str, Any] = None
+    ) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
```
galvana, line 15: Being more specific with return type since we know now to expect a dict or a list.

src/fidesops/service/processors/post_processor_strategy/post_processor_strategy_unwrap.py
eastandwestwind: thanks for restructuring, much more explicit per occurrence 💯

eastandwestwind (approved):
lgtm, nice work!

PR #311, opened 2022-03-21
Proof of concept for custom functions that override the config-derive…

galvana:
# Purpose
A proof-of-concept for custom functions that override the config-derived implementation of retrieve_data.

One of the challenges in building config-based SaaS connectors is that we need to replicate functionality that exists within the Python Requests library in YAML. The long-term goal is to provide enough "building blocks" that someone without Python experience can piece together a new integration just using the SaaS config. Even with this goal in mind, there might be some behavior that is specific to an integration and might not warrant creating a reusable strategy for.

This POC is not very robust (no error handling) but I'd like to introduce it as a conversation starter.

# Changes
- New `saas_custom_functions` directory to store custom functions
- Added `custom_function` property to `endpoints` in the `saas_config` schema
- Added to a check to `retrieve_data`  in `saas_connector` to call the custom function if available
- Added a function in `saas_connector` to dynamically load functions from the `saas_custom_functions` directory

# Checklist

### DIFF
--- a/src/fidesops/service/connectors/saas_connector.py
+++ b/src/fidesops/service/connectors/saas_connector.py
@@ -1,6 +1,7 @@
+import importlib
 from json import JSONDecodeError
 import logging
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import pydash
 from requests import Session, Request, PreparedRequest, Response
 from fidesops.common_exceptions import FidesopsException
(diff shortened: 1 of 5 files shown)


src/fidesops/service/connectors/saas_connector.py
```diff
@@ -278,5 +292,13 @@ def mask_data(
             rows_updated += 1
         return rows_updated
 
+    @staticmethod
+    def get_custom_function(function_location: str) -> Callable:
+        module, function_name = function_location.split(".")
+        module = importlib.import_module(
+            f"fidesops.service.saas_custom_functions.{module}"
```
seanpreston, line 299: Might we want each function to live in its own file? Otherwise we're forced to grow an ever expanding file here which would get messy

galvana:
This PR can be closed, it's just used a reference for https://github.com/ethyca/fidesops/issues/815

PR #326, opened 2022-03-24
[Saas Connectors] Configure Ignore Non-200 Responses [#307]

### LINKED ISSUE #307: [Saas Connector] Configure Handling for Non-200 Status Codes

Sometimes a non-200 response should just be ignored, it shouldn't cause the entire traversal to fail.  

For example, with sentry, searching a project's users for a specific email will 404 if that email doesn't exist.  We don't know if the email is going to be associated with a project ahead of time, we are just searching to see if that user's data is affiliated with the project.

I haven't thought about this much yet, but maybe a Saas Config endpoint could have an `ignore_error` flag that returns an empty list if there's an error instead of failing the entire traversal.  This may not be comprehensive enough.

### Additional context
This is a blocker for #256

pattisdr:
❗ End-to-end testing will be completed with #256 

# Purpose

Currently, a non-200 status code from a third party Saas integration causes the entire traversal to fail.  However, there are use cases where a non-200 status code is expected, and we should just ignore this response.

For example, with Sentry, searching project users with an email query param will return a 404 if that email does not exist.

# Changes
Allow setting `ignore_errors: true` on individual requests, with a default value of False.  If this is True, and the request fails, an empty Response will be returned instead. 

This can be extended in the future to have more granular handling of different response codes.

```yaml
saas_config:
  fides_key: saas_connector_example
  name: SaaS Example Config
  description: A sample schema representing a SaaS connector for Fidesops
  version: 0.0.1

...
    
  endpoints:
  - name: conversations
    requests:
      read:
        ignore_errors: true
        path: /3.0/conversations
        request_params:
          - name: count
            type: query
            default_value: 1000
          - name: offset
            type: query
            default_value: 0
          - name: placeholder
            type: query
            identity: email
        postprocessors:
          - strategy: unwrap
            configuration:
              data_path: conversations
```
-
# Checklist

# Ticket

Fixes #307


src/fidesops/schemas/saas/shared_schemas.py
```diff
@@ -22,6 +22,7 @@ class SaaSRequestParams(BaseModel):
     path: str
     params: Dict[str, Any]
     body: Optional[str]
+    ignore_errors: Optional[bool] = False
```
galvana, line 25: Just an opinion, I wonder if `ignore_errors` should be included as part of the SaaSRequestParams. I always thought of this class as being the way to track the different components of an HTTP request. Some fields like path and params get modified after going through the pagination logic, but ignore_errors just gets forwarded along. What if we pass this value in directly when we call the send function? ```self.client().send(prepared_request, saas_request.ignore_errors)```. That way we don't have to worry about it in the pagination logic, or any future code that uses SaaSRequestParams.

I approve of the current implementation and we don't have to make any changes now but I just wanted to know your thoughts.

pattisdr:
Getting this up-to-date after saas config headers merge ^

src/fidesops/schemas/saas/shared_schemas.py
pattisdr: thanks @galvana, updated to just pass this through via SaasRequest - end-to-end testing to be completed in https://github.com/ethyca/fidesops/issues/256

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -279,9 +291,13 @@ def mask_data(
             query_config.generate_update_stmt(row, policy, privacy_request)
             for row in rows
         ]
+        self.collection_name = node.address.collection
+        update_request: SaaSRequest = self.endpoints[self.collection_name].requests[
+            "update"
+        ]
```
pattisdr, line 297: I think this should work because if `update` doesn't exist, an error would have already been thrown in `generate_update_stmt` (`get_request_by_action`) -

galvana: Makes sense to me!

PR #340, opened 2022-03-29
Add Sentry Support [#256]

### LINKED ISSUE #256: [SaaS Connector] Sentry

pattisdr:
# Purpose

Adds example support for integrating with third-party infrastructure: Sentry. Note that Project Issues and Masking are not tested end-to-end here while we are waiting to get data in our test sentry instance. This will be completed in #343 

## Access request
The example Sentry Config can get Sentry user PII for Organization Users (staff), Project Issues, User Feedback, and Project Users.  Regular users will likely only have data returned from user feedback and project user endpoints. Staff could have data returned from all four endpoints.  

```
Root ---> Organizations ----> Org users (staff)
Root ---> Projects----------> Project issues (can have staff data)
                  \---------> User feedback 
                  \---------> Project users
```

## Erasure request

Very minor masking supported.  We can do an erasure to mask assignedTo to an empty string on a project only (this is a corner case for staff). 

# Changes
- Add example sentry config and sentry dataset.  Note, we currently only support looking up data by email in sentry, not username.
  - Uses new `after` and `grouped_input` params
  - Uses link pagination
- If empty row is returned, don't add to list of rows.
- Adds sentry postman collection
- Test currently only returns user feedback, project users and org users data.  No project issues in sentry instance.
- End-to-end tests for 
  - #299 
  - #306 
  - #326 

# Checklist

# Ticket

Fixes #256


data/saas/config/sentry_config.yml
```diff
@@ -0,0 +1,165 @@
...
+          param_values:
+            - name: issue_id
+              references:
+                - dataset: sentry_connector
+                  field: issues.id
+                  direction: from
+          body: "{\"assignedTo\": \"\"}"
+        read:
+          method: GET
+          path: /api/0/projects/<organization_slug>/<project_slug>/issues/
```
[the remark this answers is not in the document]
seanpreston: > I wonder if hitting any project issue counts against your allowed count

I would expect the rate limit to be set at token level vs. project level, so assuming we're using the same token for both this sounds plausible.

tests/integration_tests/test_saas_task.py
```diff
@@ -198,3 +199,198 @@ def test_saas_erasure_request_task(
         f"{dataset_name}:conversations": 0,
         f"{dataset_name}:messages": 0,
     }
+
+
+@pytest.mark.integration_saas
```
seanpreston, line 204: Can we also add `@pytest.mark.saas_connector_sentry` so we can target just Sentry should we need to

pattisdr: yes good idea

galvana (approved):
Thanks for leaving notes on the edge cases you ran into, it helped with the review. I went ahead and split out the Sentry fixtures and tests into their own files but everything else looked good!

data/saas/config/sentry_config.yml
```diff
@@ -80,11 +80,11 @@ saas_config:
                 - dataset: sentry_connector
                   field: issues.id
                   direction: from
-          body: "{\"assignedTo\": \"\"}"
+          body: '{"assignedTo": ""}'
```
pattisdr, line 83: Note the original body was intentional `body: "{\"assignedTo\": \"\"}"`, there's no test for this yet because we don't have data for this endpoint in our sentry instance. 

I believe when I tested against the sentry sandbox, specifying the body like this ` '{"assignedTo": ""}'` caused a bad request, how the body needs to be specified is pretty brittle currently, but I'll verify again in https://github.com/ethyca/fidesops/issues/343

PR #342, opened 2022-03-30
Privacy Centre [331] - Hide status bar

### LINKED ISSUE #331: [Privacy Centre] Hide the error bar on a failed privacy request after `n` seconds

When a user makes a failed privacy request, an error bar shows at the top of the page. We should automatically hide this error bar after `n` seconds.

When the follow modal displays

Screenshot 2022-03-25 at 16 28 19

it should automatically disappear after `n` seconds.

LKCSmith:
# Purpose
Hide status bar after some seconds pass.

# Changes
Setting the alert only to be hidden after 8 seconds pass.

**~~Questions:~~**
- ~~Would the Toast component from the chakra UI be more useful in this case rather than using Alerts since it has a duration prop built-in? This would require changing from Alerts to Toasts.~~
- ~~Do we only want to hide the error alerts, and no other cases?~~
- ~~Does 8 seconds seem reasonable given the length of the text?~~

# Checklist

# Ticket

Fixes #331

### DIFF
--- a/clients/privacy-center/pages/index.tsx
+++ b/clients/privacy-center/pages/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import type { NextPage } from 'next';
 import Head from 'next/head';
 import {
@@ -22,6 +22,14 @@ import config from '../config/config.json';
 const Home: NextPage = () => {
   const [alert, setAlert] = useState<AlertState | null>(null);
   const { isOpen, onClose, onOpen, openAction } = useRequestModal();
+
+  useEffect(() => {
+    if(alert?.status === 'error') {
+      const closeAlertTimer = setTimeout(() => setAlert(null), 8000);
+      return () => clearTimeout(closeAlertTimer)
+    }
+  }, [alert])
+  
   return (
     <div>
       <Head>
--- a/clients/privacy-center/pages/index.tsx
+++ b/clients/privacy-center/pages/index.tsx
@@ -24,7 +24,7 @@ const Home: NextPage = () => {
   const { isOpen, onClose, onOpen, openAction } = useRequestModal();
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


seanpreston:
> Would the Toast component from the chakra UI be more useful in this case rather than using Alerts since it has a duration prop built-in? This would require changing from Alerts to Toasts.

I'd be open to this if you think there's a compelling case.

> Do we only want to hide the error alerts, and no other cases?

In every case, apologies this could have been clearer from the spec.

> Does 8 seconds seem reasonable given the length of the text?

8 seconds is perfect.

LKCSmith:
@NevilleS @seanpreston @eastandwestwind Hey All, so I did some digging and trial and error yesterday with the toast component and it seems that it has known issues with SSR. Since we're using next, using this component here might be an issue. It runs into this error in a next environment: `Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method. ToastManager@http://localhost:3000/_next/static/chunks/pages/_app.js?ts=1627303273546:38352:5`

I've seen this issue being logged a few times in the Chakra community. I'm going to +1 an issue noted in the repo and move forward with this PR using Alerts for now. Does that sound reasonable?

NevilleS:
Oh, interesting... wow, code is hard sometimes eh?

That sounds good to me; can you leave a comment with a link to that issue so it's easy to find again later? Essentially something like "We would prefer to use a Toast component, but blocked on a bug: ..." etc. etc.

LKCSmith:
Link to associated discussion in GH: https://github.com/chakra-ui/chakra-ui/issues/4619#issuecomment-1086140024

Various related conversations also happening in Chakra UI Discord

eastandwestwind (approved):
Looks good, tested locally, and saw the status bar automatically hide after 8 secs. 

Thanks!

PR #347, opened 2022-04-02
Remove unused import

sanders41:
# Purpose
Removes the unused `Any` import in `health_endpoints.py`

# Changes
I also noticed there was not a test for the health endpoint so I added one to make sure things still worked after removing the import.

# Checklist


tests/api/v1/endpoints/test_health_endpoints.py
```diff
@@ -0,0 +1,7 @@
+from starlette.testclient import TestClient
+
+
+def test_health(api_client: TestClient) -> None:
+    url = "http://0.0.0.0:8080/health"
```
pattisdr, line 5: Rather than hardcoding this here, it might be a good opportunity to add `/health` to `urn_registry.py`  so it can be reused to both decorate the view and used as the URL in this test.

src/fidesops/api/v1/endpoints/health_endpoints.py
```diff
@@ -1,8 +1,7 @@
-from typing import Any, Dict
```
pattisdr, line 1: 🔍 good attention to detail!  looks like we're ignoring missing imports in mypy for some reason

src/fidesops/api/v1/endpoints/health_endpoints.py
```diff
@@ -1,8 +1,7 @@
...
+from typing import Dict
 
 from fastapi import APIRouter
 
-
 router = APIRouter(tags=["Public"])
```
pattisdr, line 7: if something like HEALTH= "/health" was added to urn_registry.py you could use that in the `router` decorator here as well as the test -

tests/api/v1/endpoints/test_health_endpoints.py
sanders41: Makes sense to me. I added it to the registry and updated references to the URL.

pattisdr (approved):
Thank you @sanders41!

Issue #325, opened 2022-03-23
[SaaS_Connector] Escalate update to GDPR endpoint or delete if strict masking is set to False


galvana:
### Is your feature request related to a specific problem?

Refer to #324 for a background on this topic.

Erasure requests for a SaaS connector should use the following priority list:

- Use the update request defined for the collection to perform policy-driven data masking
- If the `FIDESOPS__MASKING__STRICT` system setting is set to `False`
  - Use the GDPR endpoint
  - Use the delete request available for the collection

### Describe the solution you'd like

The `generate_update_stmt` function in `saas_query_config` should check the `FIDESOPS__MASKING__STRICT` system setting when an update endpoint cannot be found and determine if a GDPR or delete endpoint should be used instead.

**pseudocode**
```
current_request: SaaSRequest = self.get_request_by_action("update")
# log usage of update endpoint
if not current_request and not masking_strict:
  if gdpr_request
    # log usage of GDPR endpoint
  else if delete_request
    # log usage of delete endpoint
  else:
   # skip masking and log warning
else:
# skip masking and log warning
```

The `saas_config` schema should add a `data_protection_request` that exists at the root level (like `test_request`)

```
fides_key: FidesOpsKey
name: str
description: str
version: str
connector_params: List[ConnectorParam]
client_config: ClientConfig
endpoints: List[Endpoint]
test_request: SaaSRequest
data_protection_request: SaaSRequest
```

It is acceptable to call the `data_protection_request` multiple time for this iteration. Limiting calls to this endpoint is out of scope for now.

Blocked by #324

pattisdr:
@galvana, just making sure I understand, what is the `data_protection_request`?

pattisdr:
This also demonstrates a change in behavior, in that currently we fail the request if we go to mask and there's no `update` method defined, so here, if there's no update, and `MASKING_STRICT` is True, we will just warn instead of fail. Are we sure this is what we want?

PR #349, opened 2022-04-02
[WIP] Add MASKING_STRICT Config and allow GDPR Delete / DELETE Escalation [#324][#325]

### LINKED ISSUE #324: [SaaS Connector] Add system option for strict_masking

We currently fulfill policy-based erasures by masking the applicable fields of an object and submitting the updated object in the body of an update request. The problem is that updates are not always available for every API endpoint and instead we must:

- delete an entire collection (including fields that don't match the `data_category` specified in the privacy request)
- delete multiple collections (in the event of a delete cascading to other collections)
- call a dedicated GDPR endpoint

The main distinction is that the update will only mask fields that match the `data_category` of the privacy request but the listed scenarios will also delete fields that don't match the `data_category`. Therefore we must give the user the option to run strict masking (updates only) or non-strict masking (deletes or GDPR endpoint).

Create a system setting `FIDESOPS__MASKING__STRICT` which defaults to `True`

pattisdr:
❗ IN PROGRESS 

# Purpose

[Add MASKING_STRICT param. By default this is False, but if True, we can use a GDPR delete request, or a DELETE request if no UPDATE request is configured.

# Changes

-  When SaaSConnector.mask_data is called, we instantiate the `query_config` for the given Node and get the masking request from that node's config, escalating to GDPR Delete or Delete if necessary.
- Allow a data_protection_request to be configured on a SaaSConfig - this can potentially be called multiple times, limiting this is out of scope.  To constrain to just one collection, only define a primary_key on one collection for GDPR delete.

# Checklist

# Ticket

Fixes #324, #325


fidesops.toml
```diff
@@ -25,3 +25,4 @@ TASK_RETRY_COUNT=2
 TASK_RETRY_DELAY=5
 TASK_RETRY_BACKOFF=2
 REQUIRE_MANUAL_REQUEST_APPROVAL=false
+MASKING_STRICT=true
```
pattisdr, line 28: By default, `updates` only are used in our internal fidesops.toml, but you can set to `false` within an integration test to test gdpr delete/delete.

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -117,8 +118,14 @@ def __init__(self, configuration: ConnectionConfig):
...
-        """Returns the query config for a SaaS connector"""
-        return SaaSQueryConfig(node, self.endpoints, self.secrets)
+        """Returns the query config for a given node"""
+        collection_name = node.address.collection
+        configured_masking_request = self.get_masking_request_from_config(
+            collection_name
+        )
+        return SaaSQueryConfig(
+            node, self.endpoints, self.secrets, configured_masking_request
+        )
```
pattisdr, line 128: Setting the masking request here when we instantiate the `SaaSQueryConfig` so it can be used both in `SaasConnector.mask_data` as well as `SaaSQueryConfig.generate_update_stmt`

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -288,17 +317,21 @@ def mask_data(
     ) -> int:
         """Execute a masking request. Return the number of rows that have been updated"""
         query_config = self.query_config(node)
+        if not query_config.masking_request:
+            raise Exception(
+                f"No masking request configured for {node.address.collection}"
+            )
```
pattisdr, line 335: I still think we should raise an exception if nothing is configured so the user can take the appropriate steps to remedy.   They may not realize that they haven't defined an `update` request, or they may have defined a `delete` request and not set MASKING_STRICT to `false`.  If they don't want an update to be attempted at all, an easy way to remedy is to not set a primary key on the collection (this feature might not be obvious).

What are people's thoughts on this? Alternatively, we could just warn here and exit.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -160,7 +162,7 @@ def generate_update_stmt(
         This masked row is then added as the body to a dynamically generated SaaS request.
         """
 
-        current_request: SaaSRequest = self.get_request_by_action("update")
+        current_request: SaaSRequest = self.masking_request
```
pattisdr, line 165: This solution now can feed multiple types of `SaaSRequests` into `generate_update_stmt`.  Delete and GDPR deletes still share a lot of the same code as an update request - for example, Delete requests can still have request bodies, may need identity data as param values or to use connector params, etc.  

As we get more use cases I think we can add more conditional logic to `generate_update_stmt`, but I think just sharing `generate_update_stmt` among the different types of SaaSRequests is a good first step.

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -279,6 +286,40 @@ def process_response_data(
...
+        """Get the configured SaaSRequest for use in masking.
+        An update request is preferred, but we can use a gdpr delete endpoint or delete endpoint if not MASKING_STRICT.
+        """
+        requests = self.endpoints[collection_name].requests
+
+        update: Optional[SaaSRequest] = requests.get("update")
+        gdpr_delete: Optional[SaaSRequest] = None
+        delete: Optional[SaaSRequest] = None
+
+        if not config.execution.MASKING_STRICT:
```
eastandwestwind, line 301: nice, I like the way you set this up so that delete isn't even a populated option when we're in strict mode.

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -279,6 +286,40 @@ def process_response_data(
 
         return rows
 
+    def get_masking_request_from_config(
+        self, collection_name: str
+    ) -> Optional[SaaSRequest]:
+        """Get the configured SaaSRequest for use in masking.
+        An update request is preferred, but we can use a gdpr delete endpoint or delete endpoint if not MASKING_STRICT.
+        """
+        requests = self.endpoints[collection_name].requests
```
eastandwestwind, line 295: overall, can we add more return typing for vars within this method?

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -52,6 +52,7 @@ The `fidesops.toml` file should specify the following variables:
 |`TASK_RETRY_DELAY` | `FIDESOPS__EXECUTION__TASK_RETRY_DELAY` | int | 20 | 5 | The delays between retries in seconds
 |`TASK_RETRY_BACKOFF` | `FIDESOPS__EXECUTION__TASK_RETRY_BACKOFF` | int | 2 | 2 | The backoff factor for retries, to space out repeated retries.
 |`REQUIRE_MANUAL_REQUEST_APPROVAL` | `FIDESOPS__EXECUTION__REQUIRE_MANUAL_REQUEST_APPROVAL` | bool | False | False | Whether privacy requests require explicit approval to execute
+|`MASKING_STRICT` | `FIDESOPS__EXECUTION__MASKING_STRICT` | bool | True | True | Mask via updates only.  If False, can use defined deletes or GDPR endpoints, which can can delete data beyond what is specified on the Policy.
```
eastandwestwind, line 55: Would be great to clarify that if masking strict is true, you'll need an update endpoint, but if masking strict is false, you'll need at least one of the 3 endpoint types (update/gdpr/delete)

src/fidesops/service/connectors/saas_connector.py
eastandwestwind: I think we should raise an exception, but make it super clear what steps are to remedy. Maybe a clearer message would say something like "Either no masking request configured or no valid masking request for {node.address.collection}. Check that `MASKING_STRICT` env var is appropriately set."

I'm a user wouldn't want `update` at all, couldn't they simply not define an `update`, and instead define a `delete`?

eastandwestwind (commented):
This is very well organized @pattisdr ! Just have a couple comments/Qs

src/fidesops/service/connectors/saas_connector.py
pattisdr: thank you @eastandwestwind!

src/fidesops/service/connectors/saas_connector.py
pattisdr: thanks for the improved exception.  Yes, if the user doesn't want `update`, they should define a `delete`, but they also need to define `MASKING_STRICT=false`

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -52,7 +52,7 @@ The `fidesops.toml` file should specify the following variables:
...
 |`TASK_RETRY_BACKOFF` | `FIDESOPS__EXECUTION__TASK_RETRY_BACKOFF` | int | 2 | 2 | The backoff factor for retries, to space out repeated retries.
 |`REQUIRE_MANUAL_REQUEST_APPROVAL` | `FIDESOPS__EXECUTION__REQUIRE_MANUAL_REQUEST_APPROVAL` | bool | False | False | Whether privacy requests require explicit approval to execute
-|`MASKING_STRICT` | `FIDESOPS__EXECUTION__MASKING_STRICT` | bool | True | True | Mask via updates only.  If False, can use defined deletes or GDPR endpoints, which can can delete data beyond what is specified on the Policy.
+|`MASKING_STRICT` | `FIDESOPS__EXECUTION__MASKING_STRICT` | bool | True | True | If MASKING_STRICT is True, we only use "update" requests to mask data. (For third-party integrations, you should define an `update` endpoint to use.)  If MASKING_STRICT If False, you are allowing fidesops to use any defined DELETE or GDPR DELETE endpoints to remove PII. In this case, you should define `delete` or `data_protection_request` endpoints for your third-party integrations.  Note that setting MASKING_STRICT to False means that data may be deleted beyond the specific data categories that you've configured in your Policy.
```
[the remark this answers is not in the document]
pattisdr: argh thanks!

eastandwestwind (approved):
Looks great, thanks for the docs / exception message changes!

PR #356, opened 2022-04-05
Update Privacy Request Search Mechanics

seanpreston:
# Changes

- This PR changes the way in which we search for Privacy Requests by IDs
- For `PrivacyRequest.id` and `PrivacyRequest.external_id` we now look for matches that start with the value provided.
- For example `?id=12345` will return privacy requests with ids `12345` and `123456`

# Checklist

### DIFF
--- a/docs/fidesops/docs/guides/reporting.md
+++ b/docs/fidesops/docs/guides/reporting.md
@@ -51,6 +51,8 @@ If an `external_id` was provided at request creation, we can also track the priv
 
 `GET api/v1/privacy-request?external_id=<external_id>`
 
+NB. These parameters will return matching Privacy Requests based on startswith matches.
+
 ### Privacy Request Filtering Options
 
 Use the following query params to further filter your privacy requests.  Filters can be chained, for example, 
--- a/docs/fidesops/docs/guides/reporting.md
+++ b/docs/fidesops/docs/guides/reporting.md
@@ -51,7 +51,7 @@ If an `external_id` was provided at request creation, we can also track the priv
 
 `GET api/v1/privacy-request?external_id=<external_id>`
 
-NB. These parameters will return matching Privacy Requests based on startswith matches.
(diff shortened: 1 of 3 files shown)


tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -442,6 +442,49 @@ def test_get_privacy_requests_by_id(
...
+        api_client: TestClient,
+        url,
+        generate_auth_header,
+        privacy_request,
+        postgres_execution_log,
+        mongo_execution_log,
+    ):
+        auth_header = generate_auth_header(scopes=[PRIVACY_REQUEST_READ])
+        response = api_client.get(
+            url + f"?id={privacy_request.id[:5]}", headers=auth_header
```
seanpreston, line 456: This is mostly copypasta from the test above, testing that with only the first five characters passed into the search, the original privacy request is still returned.

docs/fidesops/docs/guides/reporting.md
```diff
@@ -51,6 +51,8 @@ If an `external_id` was provided at request creation, we can also track the priv
 
 `GET api/v1/privacy-request?external_id=<external_id>`
 
+NB. These parameters will return matching Privacy Requests based on startswith matches.
```
galvana, line 54: What does `NB.` mean here?

galvana (approved):
Brought up the strict search use case with @seanpreston and we decided it would be covered by the current implementation since the exact matches should be presented to the user before the wildcard matches.

docs/fidesops/docs/guides/reporting.md
seanpreston: Apologies, was under the impression this was common vernacular:

Screenshot 2022-04-05 at 13 43 55

Can change this to `Please note:` or similar

seanpreston:
Thanks for the review @galvana

PR #376, opened 2022-04-08
Validates date field ranges are sane

seanpreston:
# Purpose

This PR adds validation for date ranges so that when date field ranges are provided as query params, and subsequently combined, we return a 400 if the end date is before the start date.

I noticed that the API isn't erroring when an invalid date range is passed in via the subject request dashboard. Subsequently the UI shows no errors, so the user will be confused as to why their dashboard is still fully populated.
 
Screenshot 2022-04-07 at 22 00 11

Even if we weren't to validate this, the technically correct display would be empty as the query can return no data.


seanpreston:
Let's hold-off merging this.

As flagged by the mypy tests we'll need to be sending through a uniform type of `datetime` instead of the `Union[date, datetime]` to be sure we can actually compare the two values.

`datetime` here is necessary so that we can select all privacy requests from one day with the filters, ie. the start of `04-07-2022` to the end of `04-07-2022`.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -299,6 +299,25 @@ def get_request_status(
...
+        [created_lt, created_gt, "created"],
+        [completed_lt, completed_gt, "completed"],
+        [errored_lt, errored_gt, "errorer"],
+        [started_lt, started_gt, "started"],
+    ]:
+        if end is None or start is None:
+            continue
+        if not (isinstance(end, datetime) and isinstance(start, datetime)):
+            continue
+        else:
```
eastandwestwind, line 312: do we actually need an else here? Since we `continue` in the above conditions

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -299,6 +299,25 @@ def get_request_status(
...
+            if end < start:
+                # With date fields, if the start date is after the end date, return a 400
+                # because no records will lie within this range.
+                raise HTTPException(
```
eastandwestwind, line 316: how does front end handle 4xx for date range slider? Is this just an error banner?

eastandwestwind (approved):
minor non-blocking Qs, otherwise looks good to go

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
seanpreston: We don't, I did it to keep the diff smaller since the following lines were already at this indentation level

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
seanpreston: IIRC it currently doesn't, but the request won't work either way in this scenario currently (as no requests can be returned).

PR #380, opened 2022-04-11
Stripe Erasure

### LINKED ISSUE #365: [SaaS Connector] Stripe Erasure

galvana:
# Purpose
Updates the Stripe SaaS config with erasure endpoints.

# Changes
- Reverted `json_body` on `SaaSRequestParams` back to the more generic `body` field. JSON is now denoted by the `content-type` of `application/json`
- Added support for `application/x-www-form-urlencoded` content-type bodies
- Introduced concept of the `read-only` boolean on `fidesops_meta` to denote fields that cannot be updated or deleted. These fields are omitted in the body of the generated update requests.
- Introduced placeholder of `all_object_fields` to compliment `masked_object_fields`. This placeholder includes all of the masked fields but also row fields that were not masked. This is to support the use case where certain APIs need the entire object posted back as part of an update instead of just the modified fields.

# Checklist

# Ticket

Fixes #365


src/fidesops/schemas/dataset.py
```diff
@@ -107,6 +107,8 @@ class FidesopsMeta(BaseModel):
     """Optionally specify the allowable field length. Fidesops will not generate values that exceed this size."""
     return_all_elements: Optional[bool]
     """Optionally specify to query for the entire array if the array is an entrypoint into the node. Default is False."""
+    read_only: Optional[bool]
+    """Optionally specify if a field is read-only, meaning it can't be updated or deleted."""
```
galvana, line 111: The approach we've been taking for masking requests is to take a row from the access request and mask select fields according to the rule targets. These masked fields are then be posted via the update body. When following this approach for Stripe, it was necessary to include other required fields as part of the payload, even if they weren't being masked. However, if we took the original row, merged it with the masked fields, and posted the entire row back to the API, we would receive errors for certain fields that were not allowed in update requests.

I made a previous suggestion based on our legacy implementation to be able to flag fields with "update" or "delete" flags based on which actions were supported for that field. This would have meant having to go back and adding update/delete flags for SaaS integrations that we've already completed (which didn't seem like the right approach). The approach I settled on instead was to flag fields as "read-only" which lets us omit those fields from being included when generating the body for an update request. I am open to other suggestions if this approach does not seem flexible enough or of this flag should not be part of the more general fidesops_meta.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -48,9 +54,45 @@ def get_request_by_action(self, action: str) -> SaaSRequest:
             )
             return request
         except KeyError:
-            raise ValueError(
+            logger.info(
                 f"The '{action}' action is not defined for the '{collection_name}' endpoint in {self.node.node.dataset.connection_key}"
             )
+            return None
+
+    def get_masking_request(self) -> Optional[SaaSRequest]:
```
galvana, line 62: I moved this logic into the query config so we can update the SaaS query config's `action` field based on which request is chosen. The action is used for adding more context to logger statements and I didn't want it to be something that external code has to worry about setting.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -144,19 +180,62 @@ def map_param_values(
...
             path=path,
-            headers=headers,
+            headers=updated_headers,
             query_params=query_params,
-            json_body=json.loads(body) if body else update_values,
+            body=formatted_body,
         )
 
+    @staticmethod
+    def format_body(
```
galvana, line 195: This is my proposal for keeping the body formatting in one place. I know we've had issues with payloads being the wrong type (string vs JSON).

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -219,20 +307,53 @@ def generate_update_stmt(
...
+
+        # map of all values including those not being masked/updated
+        all_value_map: Dict[str, Any] = self.all_value_map(row)
+        complete_object: Dict[str, Any] = unflatten_dict(all_value_map)
+        deep_merge(complete_object, masked_object)
 
         # removes outer {} wrapper from body for greater flexibility in custom body config
-        param_values["masked_object_fields"] = json.dumps(update_values)[1:-1]
+        param_values["masked_object_fields"] = json.dumps(masked_object)[1:-1]
+        param_values["all_object_fields"] = json.dumps(complete_object)[1:-1]
```
galvana, line 319: New placeholder variable so user can pick if they want all the object fields or just the masked fields.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -219,20 +307,53 @@ def generate_update_stmt(
...
         saas_request_params: SaaSRequestParams = self.map_param_values(
-            current_request, param_values, update_values
+            current_request, param_values
         )
 
         logger.info(f"Populated request params for {current_request.path}")
 
         return saas_request_params
 
+    def all_value_map(self, row: Row) -> Dict[str, Any]:
```
galvana, line 330: This is the only place where use the new `fidesops_meta.read_only` field.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -191,19 +270,28 @@ def generate_update_stmt(
...
+        current_request: SaaSRequest = self.get_masking_request()
         collection_name: str = self.node.address.collection
-        collection_values: Dict[str, Row] = {collection_name: row}
+
+        # A combined dictionary of the collections that have already been retrieved
+        # along with the current row. This is to support the case where an update
+        # needs reference values from other collections, not just the values in
+        # the current row.
+        collection_values: Dict[str, Row] = {
+            **self.get_cached_results(request),
```
galvana, line 286: In this use case, the additional reference value comes from the customer collection, which only has a single row. This would not work if the reference collection has more that one row. Would it be worth it to build out a more general solution as part of Fidesops core?

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -310,64 +310,33 @@ def process_response_data(
...
     ) -> int:
-        """Execute a masking request. Return the number of rows that have been updated"""
+        """Execute a masking request. Return the number of rows that have been updated."""
+
         query_config = self.query_config(node)
-        if not query_config.masking_request:
-            raise Exception(
+        masking_request = query_config.get_masking_request()
+        if not masking_request:
+            logger.warning(
```
galvana, line 325: I changed this to a warning instead of an exception because I wanted to avoid retrying a task that will fail again.

tests/util/test_saas_util.py
```diff
@@ -253,6 +253,9 @@ def test_unflatten_dict():
     # empty dictionary
     assert unflatten_dict({}) == {}
 
+    # empty dictionary value
+    assert unflatten_dict({"A": {}}) == {"A": {}}
```
galvana, line 257: Interesting new edge case that came up!

pattisdr: good catch!

data/saas/config/stripe_config.yml
```diff
@@ -37,6 +37,28 @@ saas_config:
...
+        body: '{<all_object_fields>}'
+      delete:
+        method: DELETE
+        path: /v1/customers/<customer_id>
+        param_values:
+          - name: customer_id
+            references:
+              - dataset: stripe_connector_example
+                field: customer.id
+                direction: from
```
pattisdr, line 61: Asked elsewhere, but why are we supporting both update and delete simultaneously? Which one should someone use for masking?

src/fidesops/graph/config.py
```diff
@@ -249,6 +249,9 @@ class Field(BaseModel, ABC):
 
     is_array: bool = False
 
+    read_only: Optional[bool] = None
+    """Optionally specify if a field is read-only, meaning it can't be updated or deleted. """
+
```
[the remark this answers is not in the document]
pattisdr: We should also clarify that this is only connected for saas configs currently, this field has no effect for database operations yet.

tests/graph/test_config.py
```diff
@@ -403,6 +407,7 @@ def test_generate_object_field_with_data_categories(self):
                 is_array=False,
                 sub_fields=[apt_no_sub_field],
                 return_all_elements=None,
+                read_only=False
```
pattisdr, line 410: We could add a test where read_only=True for completeness and assert that the created field has this propagated.

src/fidesops/service/connectors/saas_query_config.py
pattisdr: This feels pretty error prone, discussed elsewhere

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -166,7 +245,7 @@ def generate_query(
         # create the source of param values to populate the various placeholders
         # in the path, headers, query_params, and body
         param_values: Dict[str, Any] = {}
-        for param_value in current_request.param_values:
```
pattisdr, line 169: Should you error here too if current_request doesn't exist?

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -144,19 +180,62 @@ def map_param_values(
...
+            (
+                value
+                for header, value in headers.items()
+                if header.lower() == "content-type"
+            ),
+            None,
+        )
+
+        if body is None:
+            return headers, None
```
pattisdr, line 214: We could return this first thing before figuring out content_type.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -144,19 +180,62 @@ def map_param_values(
...
+        if content_type is None:
+            content_type = "application/json"
+            headers["Content-Type"] = "application/json"
+
+        output: Optional[str] = None
+
+        if content_type == "application/json":
+            output = body
+        elif content_type == "application/x-www-form-urlencoded":
+            output = urlencode(json.loads(body))
```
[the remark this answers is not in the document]
galvana: Aliasing to `multidimensional_urlencode`

src/fidesops/service/connectors/saas_query_config.py
galvana: Added error check

src/fidesops/service/connectors/saas_connector.py
galvana: That's a good point, reverting back

tests/fixtures/saas/stripe_fixtures.py
```diff
@@ -87,3 +95,198 @@ def stripe_dataset_config(
...
+    )
+
+    customer = response.json()
+
+    # create dispute by adding a fraudulent card and charging it
+    response = requests.post(
+        url=f"{base_url}/v1/customers/{customer['id']}",
+        headers=headers,
+        data=urlencode({"source": "tok_createDispute"}),
+    )
```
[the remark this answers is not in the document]
galvana: Adding `assert response.ok` after each response

tests/graph/test_config.py
galvana: Test added!

tests/integration_tests/saas/test_segment_task.py
```diff
@@ -224,7 +224,7 @@ def test_segment_saas_erasure_request_task(
 
     # Create user for GDPR delete
     erasure_email = create_segment_test_data(segment_connection_config, segment_identity_email)
-    time.sleep(2)  # Pause before making access/erasure requests
+    time.sleep(20)  # Pause before making access/erasure requests
```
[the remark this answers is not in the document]
galvana: Reverting sleep to 2 seconds. Maybe we can follow a "wait until" pattern where we call the GET endpoint directly and wait until it returns a value?

tests/integration_tests/saas/test_stripe_task.py
```diff
@@ -627,3 +624,507 @@ def test_stripe_access_request_task(
...
+        f"{dataset_name}:payment_method": 2,
+        f"{dataset_name}:credit_note": 0,
+        f"{dataset_name}:bank_account": 1,
+        f"{dataset_name}:subscription": 0,
+        f"{dataset_name}:dispute": 0,
+    }
+
+    # run erasure without MASKING_STRICT to execute the delete actions
+
+    config.execution.MASKING_STRICT = False
```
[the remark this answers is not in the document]
galvana: That's correct, it was my understanding that we could have both defined, but only use update or delete based on the `config.execution.MASKING_STRICT` setting.

data/saas/config/stripe_config.yml
galvana: We can use either the update or delete, based on the value of `config.execution.MASKING_STRICT`.

src/fidesops/service/connectors/saas_query_config.py
galvana: Done!

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -144,19 +180,62 @@ def map_param_values(
...
+            None,
+        )
+
+        if body is None:
+            return headers, None
+
+        # add Content-Type: application/json if a content type is not provided
+        if content_type is None:
+            content_type = "application/json"
+            headers["Content-Type"] = "application/json"
```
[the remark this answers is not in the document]
galvana: Done!

tests/fixtures/saas/stripe_fixtures.py
```diff
@@ -87,3 +95,198 @@ def stripe_dataset_config(
     )
     yield dataset
     dataset.delete(db=db)
+
+
+@pytest.fixture(scope="function")
+def stripe_create_erasure_data(stripe_secrets) -> Generator:
```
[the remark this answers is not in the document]
galvana: Done!

src/fidesops/util/saas_util.py
```diff
@@ -121,3 +121,15 @@ def unflatten_dict(flat_dict: Dict[str, Any], separator: str = ".") -> Dict[str,
...
+def deep_merge(target: dict, source: dict) -> None:
+    """
+    Adds all leaf values in the source dictionary to the target dictionary.
+    This is a destructive operation since the target is modified in place.
+    """
+    for key, value in source.items():
+        if isinstance(value, dict):
+            deep_merge(target.setdefault(key, {}), value)
+        else:
+            target[key] = value
```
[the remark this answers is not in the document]
galvana: Decided to not reinvent the wheel here. I can just merge the two dictionaries using the existing `merge_dicts` function before unflattening to get the final result.

tests/integration_tests/saas/test_stripe_task.py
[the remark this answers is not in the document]
pattisdr: Got it, so you're providing both options here for completeness, thanks for clarifying

src/fidesops/service/connectors/saas_query_config.py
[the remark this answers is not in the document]
galvana: Added

tests/integration_tests/saas/test_stripe_task.py
```diff
@@ -627,3 +624,507 @@ def test_stripe_access_request_task(
...
+        f"{dataset_name}:invoice": 2,
+        f"{dataset_name}:card": 1,
+        f"{dataset_name}:customer_balance_transaction": 0,
+        f"{dataset_name}:payment_intent": 0,
+        f"{dataset_name}:payment_method": 2,
+        f"{dataset_name}:credit_note": 0,
+        f"{dataset_name}:bank_account": 1,
+        f"{dataset_name}:subscription": 1,
+        f"{dataset_name}:dispute": 0,
+    }
```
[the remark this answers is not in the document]
galvana: Added follow-up reads after the updates (to verify masking) and deletes (to verify deletion).

src/fidesops/service/connectors/saas_query_config.py
galvana: Removed the need for this. For Stripe, there is enough information on the row to build the update request, we don't need to access the cache results.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -144,29 +180,77 @@ def map_param_values(
...
+            ),
+            None,
+        )
+
+        # add Content-Type: application/json if a content type is not provided
+        if content_type is None:
+            content_type = "application/json"
+            headers["Content-Type"] = content_type
+
+        output: Optional[str] = None
```
pattisdr, line 221: This variable is not being used, since you're overriding it in every instance below, or raising an error.

tests/integration_tests/saas/test_segment_task.py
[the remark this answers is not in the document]
pattisdr: yeah, that's something to look into more!

src/fidesops/util/saas_util.py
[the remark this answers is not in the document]
pattisdr: nice use of existing functionality!

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -219,20 +307,53 @@ def generate_update_stmt(
...
+
+    @staticmethod
+    def get_cached_results(request: PrivacyRequest) -> Dict[str, Any]:
+        """
+        Returns a map of the first item in each collection of the current
+        dataset (if available).
+        """
+        return {
+            k.split(":")[-1]: v[0] if v else v for k, v in request.get_results().items()
+        }
```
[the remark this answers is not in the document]
pattisdr: note to self: this method has been removed

tests/integration_tests/saas/test_stripe_task.py
```diff
@@ -627,3 +625,596 @@ def test_stripe_access_request_task(
...
+    response = requests.get(url=f"{base_url}/v1/customer/{customer_id}/tax_ids")
+    tax_ids = response.json()
+    assert tax_ids["error"]["message"] == f"No such customer: '{customer_id}'"
+
+    # invoice_item
+    response = requests.get(
+        url=f"{base_url}/v1/invoiceitems", params={"customer": {customer_id}}
+    )
+    invoice_item = response.json()["data"]
+    assert len(invoice_item) == 1
```
pattisdr, line 1183: Can you add a code comment explaining why there's one invoice left?  I assume it's a certain type of invoice that you're not allowed to mask/delete

data/saas/config/stripe_config.yml
pattisdr: Can you clarify that both do not have to be specified for masking to work? I don't think this is clear in the docs, that only one needs to be defined, and that only one will be used.

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -219,20 +304,45 @@ def generate_update_stmt(
...
         # mask row values
         update_value_map: Dict[str, Any] = self.update_value_map(row, policy, request)
-        update_values: Dict[str, Any] = unflatten_dict(update_value_map)
+        masked_object: Dict[str, Any] = unflatten_dict(update_value_map)
+
+        # map of all values including those not being masked/updated
+        all_value_map: Dict[str, Any] = self.all_value_map(row)
+        # both maps use field paths for the keys so we can merge them before unflattening
+        complete_object: Dict[str, Any] = unflatten_dict(
+            merge_dicts(*[all_value_map, update_value_map])
```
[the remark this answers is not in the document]
galvana: Simplified the function call and added a comment.

tests/integration_tests/saas/test_stripe_task.py
galvana: ```Can't delete an invoice item that is attached to an invoice that is no longer editable.```

data/saas/config/stripe_config.yml
galvana: Sorry! Your original question just clicked for me. I removed the deletes from the endpoints that already have an update.

pattisdr: okay great, thanks, I think this will help clear up confusion for people looking at the stripe yaml!

src/fidesops/service/connectors/saas_query_config.py
pattisdr: makes sense to me -

pattisdr (approved):
Looks good @galvana, great work here.  Waiting to merge because of travel this week - also @conceptualshark may want to look at docs and the segment test may need to be given a little more time (less than 20 but greater than 2)

conceptualshark:
Made a few updates to the flow/organization of the docs - if that all still seems to make sense with your changes, this is a go from me 👍

Issue #383, opened 2022-04-12
[User Management] Conditionally force password reset on user login


seanpreston:
### Is your feature request related to a specific problem?

We need a secure way to force users to set their own passwords.

### Describe the solution you'd like

- A new field `FidesopsUser.last_login_at` — if that field is `null` then we know the user has never logged in before
  - Every time a user logs in this field should be updated
- A new field `FidesopsUser.password_reset_at`
  - Every time the user's password is updated we should update this field, this allows us to do things like specify a max. user password age
  - If we want to force a user to reset their password we set this to `null` or some datetime older than the max. age

PR #394, opened 2022-04-14
add date fields to user model for last login and last password reset

adamsachs:
# Purpose
User model support for resetting passwords
# Changes
-add `last_login_at` and `password_reset_at` datetime fields to the FidesopsUser model class
-corresponding db migration to add these fields
-add functions to login and reset password on the FidesopsUser model class
-invoke login function as part of login API endpoint
# Checklist

# Ticket
Fixes https://github.com/ethyca/fidesops/issues/383

### DIFF
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -100,6 +100,9 @@ def user_login(
             status_code=HTTP_403_FORBIDDEN, detail="Incorrect password."
         )
 
+    # perform what's needed on the user model object
+    user.login(db)
+
     client: ClientDetail = user.client
     if not client:
         logger.info("Creating client for login")
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -27,7 +27,6 @@ router = APIRouter(tags=["Users"], prefix=V1_URL_PREFIX)
 def perform_login(db: Session, user: FidesopsUser) -> ClientDetail:
     """Performs a login by updating the FidesopsUser instance and
     creating and returning an associated ClientDetail."""
-    user.last_login_at = datetime.utcnow()
 
     client: ClientDetail = user.client
     if not client:
@@ -36,7 +35,9 @@ def perform_login(db: Session, user: FidesopsUser) -> ClientDetail:
             db, SCOPE_REGISTRY, user_id=user.id
         )
 
+    user.last_login_at = datetime.utcnow()
     user.save(db)
+
     return client
 
 
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -5,6 +5,7 @@ from starlette.status import (
     HTTP_404_NOT_FOUND,
     HTTP_403_FORBIDDEN,
 )
+from datetime import datetime
 
 from fidesops.api import deps
 from fidesops.api.v1 import urn_registry as urls
@@ -17,14 +18,26 @@ from fidesops.schemas.user import UserCreate, UserCreateResponse, UserLogin
 from fidesops.util.oauth_util import verify_oauth_client
 from sqlalchemy.orm import Session
 
-from fidesops.api.v1.scope_registry import (
-    USER_CREATE,
-    USER_DELETE,
-    SCOPE_REGISTRY,
-)
+from fidesops.api.v1.scope_registry import USER_CREATE, USER_DELETE, SCOPE_REGISTRY
 
 logger = logging.getLogger(__name__)
 router = APIRouter(tags=["Users"], prefix=V1_URL_PREFIX)
+
+
+def perform_login(db: Session, user: FidesopsUser) -> ClientDetail:
+    """Performs a login by updating the FidesopsUser instance and
+    creating and returning an associated ClientDetail."""
+    user.last_login_at = datetime.utcnow()
+
+    client: ClientDetail = user.client
+    if not client:
+        logger.info("Creating client for login")
+        client, _ = ClientDetail.create_client_and_secret(
+            db, SCOPE_REGISTRY, user_id=user.id
+        )
+
+    user.save(db)
+    return client
 
 
 @router.post(
@@ -90,7 +103,9 @@ def user_login(
     *, db: Session = Depends(deps.get_db), user_data: UserLogin
 ) -> AccessToken:
     """Login the user by creating a client if it doesn't exist, and have that client generate a token"""
-    user = FidesopsUser.get_by(db, field="username", value=user_data.username)
+    user: FidesopsUser = FidesopsUser.get_by(
+        db, field="username", value=user_data.username
+    )
 
     if not user:
         raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail="No user found.")
@@ -100,15 +115,7 @@ def user_login(
             status_code=HTTP_403_FORBIDDEN, detail="Incorrect password."
         )
 
-    # perform what's needed on the user model object
-    user.login(db)
-
-    client: ClientDetail = user.client
-    if not client:
-        logger.info("Creating client for login")
-        client, _ = ClientDetail.create_client_and_secret(
-            db, SCOPE_REGISTRY, user_id=user.id
-        )
+    client: ClientDetail = perform_login(db, user)
 
     logger.info("Creating login access token")
     access_code = client.create_access_code_jwe()
--- a/src/fidesops/models/fidesops_user.py
+++ b/src/fidesops/models/fidesops_user.py
@@ -1,6 +1,6 @@
 from typing import Dict, Any
 
-from sqlalchemy import Column, String
+from sqlalchemy import Column, String, DateTime
 from sqlalchemy.orm import Session, relationship
 
 from fidesops.core.config import config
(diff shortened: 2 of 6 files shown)


src/fidesops/models/fidesops_user.py
```diff
@@ -14,6 +14,8 @@ class FidesopsUser(Base):
     username = Column(String, unique=True, index=True)
     hashed_password = Column(String, nullable=False)
     salt = Column(String, nullable=False)
+    last_login_at = Column(DateTime(timezone=True), nullable=True)
```
seanpreston, line 17: Great start! Now we just need to have this field update every time the user logs in (https://github.com/ethyca/fidesops/blob/main/src/fidesops/api/v1/endpoints/user_endpoints.py#L89)

adamsachs:
@seanpreston if you could look over [my latest commit](https://github.com/ethyca/fidesops/pull/394/commits/52451f3c3f8c0df61110fd6e9887c2a2b973a148), that'd be great. from my naive POV, i thought that only the basic password_reset functionality belonged on the model object -- and felt that any logic around validating existing credentials, looking up the user, etc. belonged in the API layer, and probably deserves a new endpoint?

if you agree with that, i'm happy to try and create that new endpoint, i just wanted to go about this piecemeal, especially given my lack of familiarity with the codebase. happy to discuss more offline.

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -100,6 +100,9 @@ def user_login(
             status_code=HTTP_403_FORBIDDEN, detail="Incorrect password."
         )
 
+    # perform what's needed on the user model object
+    user.login(db)
+
     client: ClientDetail = user.client
```
pattisdr, line 106: It feels strange to me to have this login method that only sets the `last_login_at` field but doesn't actually do any of the "logging in", which happens by creating this separate ClientDetail resource below.  There are a few ways to organize this (because the logging in involves updating two models), but my thought is that having all the "logging in" logic in one method is more understandable.

src/fidesops/models/fidesops_user.py
```diff
@@ -48,3 +57,18 @@ def credentials_valid(self, password: str) -> bool:
...
+
+    def update_password(self, db: Session, new_password: str) -> None:
+        """Updates the user's password to the specified value.
+        No validations are performed on the old/existing password within this function."""
+
+        hashed_password, salt = FidesopsUser.hash_password(new_password)
+        self.hashed_password = hashed_password
+        self.salt = salt
+        self.password_reset_at = datetime.utcnow()
+        self.save(db)
```
pattisdr, line 74: It'd be nice to have a model-level test on this method - nice job moving the hash_password logic into its own class method to be shared between create/update.

src/fidesops/api/v1/endpoints/user_endpoints.py
adamsachs: yeah, that makes a lot of sense. i should have taken a more comprehensive look at the whole workflow here. your suggestion seems good to me - let me get a more holistic understanding of the user/client framework, and i'll adjust accordingly.

src/fidesops/models/fidesops_user.py
adamsachs: agreed - will do.

adamsachs: added in https://github.com/ethyca/fidesops/pull/394/commits/0c6766548f54a58e1a27f5c41bd49a9b8eacdc62

src/fidesops/api/v1/endpoints/user_endpoints.py
adamsachs: addressed in https://github.com/ethyca/fidesops/pull/394/commits/308a53035a3edffbab869a5649a1c12a4ea5ebfd 

note that our our first choice had been to move the `ClientDetail` instantiation into the `login` method on the `FidesopsUser` class, but doing so introduced a circular dependency between `ClientDetail` and `FidesopsUser` -- the fk specification on `ClientDetail` refers to the `FidesopsUser.id_field_path` class field. so @pattisdr and i decided on the solution in the above commit.

src/migrations/versions/c9759675b5d3_add_user_login_and_password_reset_dates.py
```diff
@@ -0,0 +1,31 @@
...
+
+def upgrade():
+    op.add_column(
+        "fidesopsuser",
+        sa.Column("last_login_at", sa.Date(), nullable=True),
+    )
+    op.add_column(
+        "fidesopsuser", sa.Column("password_reset_at", sa.Date(), nullable=True)
+    )
+
```
pattisdr, line 27: Your models have these variables as DateTime type, but the migration adds them as Dates.  I'd change this migration to DateTime type, so we record more detail.

src/fidesops/models/fidesops_user.py
```diff
@@ -48,3 +57,13 @@ def credentials_valid(self, password: str) -> bool:
...
```
pattisdr, line 69: Is there another ticket that adds an API endpoint to update the password/validate that the existing password is correct before doing so? If not, can you create one?

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -17,16 +18,28 @@
...
+    user.last_login_at = datetime.utcnow()
+
+    client: ClientDetail = user.client
+    if not client:
+        logger.info("Creating client for login")
+        client, _ = ClientDetail.create_client_and_secret(
+            db, SCOPE_REGISTRY, user_id=user.id
+        )
+
+    user.save(db)
```
pattisdr, line 39: minor, only leaving this comment since we need the migration update too - I'd group "like" functionality together, so saving the user above on L31 instead of on L39.

pattisdr (commented):
Looks good @adamsachs, just a couple of minor things and this is good to go.  Thanks for all the good docstrings and PR comments -

src/fidesops/models/fidesops_user.py
adamsachs: i took a shot with https://github.com/ethyca/fidesops/issues/404 - happy to discuss/adjust as needed. thanks!

pattisdr: looks good @adamsachs thanks

pattisdr (approved):
looks good to me 👍

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -17,16 +18,29 @@
...
 
 logger = logging.getLogger(__name__)
 router = APIRouter(tags=["Users"], prefix=V1_URL_PREFIX)
 
 
+def perform_login(db: Session, user: FidesopsUser) -> ClientDetail:
+    """Performs a login by updating the FidesopsUser instance and
+    creating and returning an associated ClientDetail."""
+
+    client: ClientDetail = user.client
```
seanpreston, line 31: Hey @adamsachs: are you 100% sure that a logout purges this relation alongside deleting the record in the `ClientDetail` table? It'd be good to have certainty on that part, otherwise this check may return false negatives if say, `user.client` is derived from `user.client_id`, and that's pointing to a deleted `ClientDetail` record.

src/fidesops/models/fidesops_user.py
```diff
@@ -48,3 +57,13 @@ def credentials_valid(self, password: str) -> bool:
         )
 
         return provided_password_hash == self.hashed_password
+
+    def update_password(self, db: Session, new_password: str) -> None:
```
seanpreston, line 61: It would be good to have some validation in this flow such that the old password is verified before the password is updated to the new password. Another take on this could be that so long as the endpoint is protected with the right scopes, and that we're validating that the token's user and user who's password is being updated matches, then that's enough.

pattisdr: There's also discussion over adding this validation in follow-up https://github.com/ethyca/fidesops/issues/404

src/fidesops/api/v1/endpoints/user_endpoints.py
adamsachs: good question, i was curious about the lifecycle here too. to be clear - i don't think my changes should have had any impact on this behavior (though i could be wrong about that). either way, i can see why it's worth confirming the behavior now that our attention is on it.

seanpreston: Agree, none of these changes would affect the behaviour, though the behaviour will affect whether this conditional works as intended I believe?

adamsachs: right, i guess i'm just saying that the conditional was already in place, and i'd tried not to touch more than the scope of the ticket - but i can see that it's worth validating the behavior here.

a little testing locally seems to confirm your suspicion (i'm getting a failed assertion when i check whether the user's `client` is `None` after the logout is performed in a unit test).

i see a few ways of potentially addressing:
-have the logout endpoint remove the user -> client reference
-upon login (i.e. in this spot), perform a lookup on the referenced client to check if its stale/deleted. if so, create a new client
-just create a new client/session any time the `/login` endpoint is invoked, whether or not the user has logged out. i'm not exactly sure on the intended/expected semantics of `/login`, i.e. if it's necessary (or desirable) that it reuses an existing and valid session

pattisdr: Are you sure @adamsachs? (Can you specify what your failed assertion is)  If in testing, **you reload the user object**, I'm fairly sure that the `client` won't exist.  This is only a relationship in the ORM side.

adamsachs: @pattisdr helped me out and made me realize i was misinterpreting my local test results. in https://github.com/ethyca/fidesops/pull/394/commits/42740520d8a59a3b815518be664f4d48a208a379 and https://github.com/ethyca/fidesops/pull/394/commits/011d810c6e7b1383a39fbf00bf96a7d6e8b97627 i've updated the test cases to confirm that the user -> client pointer is getting cleaned up, and that requests with outdated client credentials are properly rejected with a 403.

PR #405, opened 2022-04-21
GET for users in an org

### LINKED ISSUE #381: [User Management] GET for users in an org

The admin UI needs to display all users in an organisation.

`GET /api/v1/users/`: returns a paginated list of all users
`GET /api/v1/user/:user_id/`: returns just the user requested

TheAndrewJackson:
# Purpose
The admin UI needs to display all users in an organization.

# Changes
Added 2 new routes
- GET `/api/v1/users/`: returns a paginated list of all users
- GET `/api/v1/user/:user_id/`: returns just the user requested

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)). If docs have been changed, tag in @conceptualshark for review.

# Ticket
Fixes #381


src/fidesops/api/v1/scope_registry.py
```diff
@@ -45,6 +45,7 @@
 SAAS_CONFIG_DELETE = "saas_config:delete"
 
 USER_CREATE = "user:create"
+USER_READ = "user:read"
```
seanpreston, line 48: One subtle change we need to make here is to not only have a scope for reading the current, logged-in, user. We should also have another scope for reading all users in the system. Could we name this scope accordingly? Something like `ALL_USERS_READ` would be clearer, that way we can have another for `CURRENT_USER_READ`. We'd need to follow this same convention then for delete and update too, that can be addressed subsequently.

The alternative here is that we don't have special scopes for users updating their own info, and that on those endpoints we verify that the token making the request has either the user update scope, or belongs to the user being updated.

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -49,6 +59,34 @@ def create_user(
...
+    urls.USERS,
+    dependencies=[Security(verify_oauth_client, scopes=[USER_READ])],
+    response_model=Page[UserResponse],
+)
+def get_users(
+    *, db: Session = Depends(deps.get_db), params: Params = Depends()
+) -> AbstractPage[FidesopsUser]:
+    """Returns a paginated list of all users"""
+    logger.info(f"Returned a paginated list of all users.")
+    return paginate(FidesopsUser.query(db), params=params)
```
seanpreston, line 72: Paginated queries need a default order passed into the query too, so that each page can be selected consistently. See: https://github.com/ethyca/fidesops/blob/main/src/fidesops/api/v1/endpoints/privacy_request_endpoints.py#L354

seanpreston: I've added a bug for the remaining parts of the system where this is an issue here: https://github.com/ethyca/fidesops/issues/411

seanpreston (changes requested):
Great start @TheAndrewJackson! Just a couple small tweaks

src/fidesops/api/v1/scope_registry.py
TheAndrewJackson: Is it okay that users with the `CURRENT_USER_READ` scope are able to read the data for any user as along as they have the `user_id`? Currently the `get_user` route doesn't validate that it's returning the currently logged in users data.

`CURRENT_USER_READ` makes me thing this scope is limited to reading the currently logged in users data.

seanpreston: > Is it okay that users with the CURRENT_USER_READ scope are able to read the data for any user as along as they have the user_id?

`CURRENT_USER_READ`: this should allow the token to read _only_ the logged in user, or nothing at all if the client which issued the token does not have an associated user.

TheAndrewJackson: Got it. #381 says `GET /api/v1/user/:user_id/: returns just the user requested`. Will the `get_users` function need to have conditional behavior based on the current users scopes?

It could work like this:

- `CURRENT_USER_READ`: Only return the user associated with the current token
- `USER_READ`: Return any user

src/fidesops/api/v1/scope_registry.py
```diff
@@ -45,6 +45,8 @@
...
+ALL_USERS_READ = "user:read-all"
```
seanpreston, line 49: From our chat earlier I was under the impression that:
- `USER_READ` would be the scope to signify that the token can read all users in the system
- `CURRENT_USER_READ` will be a scope we can add later for the endpoint returning only the current logged in user's data

Is that about right?

TheAndrewJackson: Yes, that is correct. That is how the api's are implemented right now.

- `USER_READ` is currently doing what is describe in your comment 
- `CURRENT_USER_READ` will be added along with a new GET route in the future per our conversation

seanpreston (approved):
This is good to merge once the conflicts are fixed, thanks @TheAndrewJackson

eastandwestwind (approved):

PR #425, opened 2022-04-27
#391 track user privileges

### LINKED ISSUE #391: [User Management] Track user privileges in the DB

We need a way of assigning users with privileges on a permanent basis.

We should create a new model that can be linked to a `FidesopsUser`, which keeps track of the privileges that are assigned to that user. This new model will need its own endpoints with a new scope attached to manipulate user privileges.

- We need to keep track of this outside of the context of a user's `ClientDetail` because those clients will be deleted every time a user logs out
- What should the default permissions be for a user?
  - `privacy-request:read`

TheAndrewJackson:
# Purpose
Add a way to assign users with privileges on a permanent basis.

# Changes
- `FidesopsUserPermissions` model that keeps track of user privileges across logins
- Accompanying routes for `FidesopsUserPermissions` to retrieve and modify permissions for a user
- Automatically apply the permissions stored in `FidesopsUserPermissions` to `ClientDetail` when a user logs in
- Removed some unused imports

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md)). If docs have been changed, tag in @conceptualshark for review.

# Ticket

Fixes #391


src/fidesops/api/v1/endpoints/user_permission_endpoints.py
```diff
@@ -0,0 +1,91 @@
...
+    db: Session = Depends(deps.get_db),
+    user_id: str,
+    permissions: UserPermissionsEdit,
+) -> FidesopsUserPermissions:
+    user = validate_user_id(db, user_id)
+    logger.info("Updated FidesopsUserPermission record")
+    return FidesopsUserPermissions.create_or_update(
+        db=db,
+        data={"id": user.permissions.id, "user_id": user_id, **permissions.dict()},
+    )
```
seanpreston, line 78: We'll also need to update that User's `ClientDetail` with the new scopes if it exists, so that the new scopes can be issued.

src/fidesops/schemas/user_permission.py
```diff
@@ -0,0 +1,35 @@
...
+
+class UserPermissionsCreate(BaseSchema):
+    """Data required to create a FidesopsUserPermissions record"""
+
+    scopes: List[str]
+
+    @validator("scopes")
+    def validate_scopes(cls, scopes: List[str]) -> List[str]:
+        """Validates that all incoming scopes are valid"""
+        if not all([scope in SCOPE_REGISTRY for scope in scopes]):
```
seanpreston, line 17: I'd preference using set arithmetic here. Something like:
```
diff = set(SCOPE_REGISTRY).difference(set(scopes))
```
that way we can include the invalid scopes (the elements of `diff`) as part of the error message.

src/fidesops/api/v1/endpoints/user_permission_endpoints.py
TheAndrewJackson: Good catch!

src/fidesops/schemas/user_permission.py
TheAndrewJackson: Good idea. I'll add some tests to make sure the error message contains the invalid scopes.

src/fidesops/api/v1/urn_registry.py
```diff
@@ -77,6 +77,9 @@
 USERS = "/user"
 USER_DETAIL = "/user/{user_id}"
 
+# User Permission URLs
+USER_PERMISSIONS = f"{USER_DETAIL}/permission"
```
seanpreston, line 81: Nit: Could you actually show this as the whole URL `/user/{user_id}/permission` since that's how the others are stored

seanpreston (approved):
Thanks @TheAndrewJackson, just one tiny nit left to change then let's merge.

PR #443, opened 2022-04-29
update masking API to take multiple input values

### LINKED ISSUE #441: Update masking API to support multiple input values per invocation
The masking engine project requires some updates to the core fidesops masking API. Specifically, per the [requirements doc](https://ethyca.atlassian.net/wiki/spaces/OSS/pages/2267873292/Fidesops-Plus+Masking+Engine+Requirements), the masking API should be updated to be able to take multiple input values (i.e., an array of values) to be masked. The masked values should also be returned as an array.

* The `/api/v1/masking/mask` endpoint should be updated such that its `value` query parameter support multiple values to be masked.
    * Fast API provides built-in support for multi-value query params, documented [here](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#query-parameter-list-multiple-values). The client provides multiple values by simply repeating the query param, e.g. `/api/v1/masking/mask?value=value1&value=value2`.
     * Single value inputs to be masked should obviously still be accepted.
* The internal `MaskingStrategy.mask` abstract method should be updated accordingly to support a `List[str]` input as its `value` argument, and it should return a `List[str]`. Currently both are typed as just single-value `str`.
    * Importantly, this means that it is up to the particular `MaskingStrategy` implementation to define any sort of parallelization of work across the multiple input values. See alternatives considered below for more information.
* The response body to our masking call should be updated to pass back arrays of `plain` and `masked_value` values, rather than single-value fields.
    * This should be considered a breaking public API change, in that we are changing the public-facing API here. We are not aware of anyone using this the `/mask` API directly at this point, so it should have minimal public-facing impact.
* The internal use of the masking functionality by other `fidesops` operations will need to be updated, since the `MaskingStrategy.mask` abstract method's signature will be changed. These changes should be minimal - passing in singleton lists and handling singleton lists that are returned.
* The existing `MaskingStrategy` subclasses will need to be updated to support the `mask` method updates. At this point, all existing `MaskingStrategy` subclasses should just simply loop through their input values serially. This should mimic existing functionality.
    * As a note, the NLP `MaskingStrategy` implementation, which this ticket is aimed at supporting eventually, but has not yet been implemented in fidesops, should implement more complex handling of multi-value input. That work is not in scope for this specific ticket.

I've considered whether we should avoid changes to the internal API (i.e. the `MaskingStrategy.mask` abstract method), keeping that as a single-value operation, and instead only update the thin public API layer to take multiple input values. The solution described above pushes multi-value support further down the stack, which I think increases flexibility in how individual masking strategies handle multi-value inputs. Rather than imposing any strategy on the "abstract" level, we give each masking strategy the leeway to implement any parallelization or special-handling of multiple input values how it wishes. Looking specifically toward the thrust of this ticket -- the NLP masking support -- this means we can implement a handling multiple free-text input values to mask in a way that is optimized for the specific NLP masking processing, without having to worry about whether this handling is generally optimal for other types of masking strategies.

Given that our API calls still only support one masking strategy per call (i.e., multiple values but not multiple masking strategies), this implementation also doesn't limit us in any way. If, at some point, there were compelling reasons to support multiple values and multiple masking strategies per API call, we may look to revisit the implementation here and support parallelization at the abstract level, i.e. across the different masking strategies provided.

adamsachs:
# Purpose
Allow masking API calls to take multiple input values. This is part of masking engine enhancements outlined in #441 

# Changes
-update public-facing masking API to take multiple input values and return multiple output (masked) values
-move input values into a multi-value request body field rather than query param and update request model accordingly. 
-update internal masking strategy API, implementations and references to support multi-value inputs and ouptuts.
-update existing tests accordingly to accommodate API changes
-add tests for multi-value support

# Checklist

  - [] documentation issue created (tag docs-team to complete issue separately)

# Ticket

Fixes #441


TheAndrewJackson (approved):
Looks great!

TheAndrewJackson:
@adamsachs After thinking about it a little more I realized that this PR contains a breaking change. Any consumer of the previous query param based endpoints will have their code not function as expected. I think this is okay for now. It's just something to keep in mind when it's time to release the next version. We need to document this in the release notes and possibly increase to a 2.0.0 version since we're following [Semantic Versioning](https://semver.org/)

adamsachs:
yup @TheAndrewJackson, thanks for drawing attention to that! a few comments:

* i'd actually initially spec'd the enhancement in a way that maintained backward-compatibility with the public-facing API. based on @NevilleS 's comments on https://github.com/ethyca/fidesops/issues/441, and some private discussion, we thought it'd make more sense to make the breaking-change updates to the public API (the comments on the issue proivde some reasoning why). that being said, it'd probably be good to get some wider buy-in on this decision from the team (you and others)
* totally agree that we need to document in release notes, regardless on the implementation we go with. i know there was some talk at the team healthcheck on formalizing the release process, and making sure changes are well documented in release notes. i'm not sure if we've really done anything to that end yet. i'd labeled the PR as [Needs doc review](https://github.com/ethyca/fidesops/labels/Needs%20doc%20review), which i think may be a starting point? but i'd be happy to do whatever else we think is needed to make sure this is caught in release notes.
* agreed on the semantic versioning implications. i'm not sure how closely we're following. it would seem like this is not a big enough feature to bump up to 2.0.0, but i also have no idea on the current release plans of the team... probably need @seanpreston to weigh in here.

adriaaaa:
@eastandwestwind and @galvana and @adriaaaa to review

adamsachs:
summarizing the outcomes of the review so we have it documented here:

- we discussed potentially responding with an error/warning response with a clear message to requests that attempt to hit the new API with the old API's request model. doing so would not be straightforward, technically, as the old requests will not make it through the basic fast API request validation framework. the fast API support for route-specific exception overriding seems to currently be a bit [lackluster]( https://github.com/tiangolo/fastapi/issues/1667), and so it would likely require creating and injecting a [custom route class](https://fastapi.tiangolo.com/advanced/custom-request-and-route/#custom-apiroute-class-in-a-router). given those considerations, we decided against this special error handling
- @pattisdr also pointed out that we've had breaking changes before, and because there is not heavy usage yet, there hasn't been much fallout. that being said, we should look to at least make sure we're doing a good job documenting those breaking changes in release notes. @TheAndrewJackson raised the idea of incorporating a `Changelog.md` into fidesops, similar to what's being done now in [fidesctl](https://github.com/ethyca/fides/blob/main/CHANGELOG.md). he'll be following up on this effort
- for now, tagging @ethyca/docs-authors here just to make extra sure that the breaking changes made in this PR are properly documented in 1.5.0 release notes, and to review the documentation updates made as part of this PR

PR #448, opened 2022-05-02
Adding missing import

galvana:
# Purpose
To fix linter issue on `main` where `fidesops.core.config` is missing from `saas_connector.py`

# Changes
- adding missing import

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [ ] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

PR #453, opened 2022-05-03
414- adds DRP action to Policy

### LINKED ISSUE #414: [DRP] Add DRP action name to Fidesops Policy Model
### Requirements

1. MUST have the ability to specify a “DRP action” when creating or editing a policy
  - Right now `action` will be equivalent to the `right` column from the spec. here: https://github.com/consumer-reports-digital-lab/data-rights-protocol#301-supported-rights-actions and we should only focus on `access` and `delete`.
2. MUST require that a given DRP action is only used once (this ensures a 1:1 match from DRP actions to Fidesops policies)
3. MUST validate the action against the list of valid actions defined in the DRP [specification](https://github.com/consumer-reports-digital-lab/data-rights-protocol#301-supported-rights-actions)
4. SHOULD NOT require that DRP actions are defined; it is valid to have multiple Fidesops policies defined but none mapped to a DRP action
5. SHOULD NOT ascribe any specific behaviour to this DRP action, it is purely to translate between standards-based DRP actions and the user-defined policies in Fidesops

### Implementation Notes

- includes ingesting - the whole flow to associating an action with a policy
- SQLAlchemy will allow us to specify that the DRP action field as unique.
- validating the action can be db-level validation
- see full requirements [here](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2259943434/Fidesops+Data+Rights+Protocol+DRP+Implementation+Requirements)

eastandwestwind:
# Purpose
Adds ability to add DRP Action to any given Policy.

# Changes
- DB migration to Policy table
- Adds DRP action to Policy model
- Adds validation such that DRP action is valid, and that any given DRP action is only used once across Policies
- Adds docs
- Adds unit tests for Policy model/endpoints

# Checklist

# Ticket

Fixes #414

### DIFF
--- a/src/fidesops/models/policy.py
+++ b/src/fidesops/models/policy.py
@@ -323,9 +323,6 @@ class Rule(Base):
     @classmethod
     def create(cls, db: Session, *, data: Dict[str, Any]) -> FidesopsBase:
         """Validate this object's data before deferring to the superclass on update"""
-        associated_policy: Optional[Policy] = (
-            db.query(Policy).filter_by(id=data["policy_id"]).first()
-        )
         _validate_rule(
             action_type=data.get("action_type"),
             storage_destination_id=data.get("storage_destination_id"),
--- a/src/fidesops/models/policy.py
+++ b/src/fidesops/models/policy.py
@@ -85,16 +85,10 @@ when project migrations are consolidated.
 """
 
 
-def _validate_drp_action(
-    drp_action: Optional[str], drp_action_exists: bool, class_name: str
-) -> None:
+def _validate_drp_action(drp_action: Optional[str]) -> None:
     """Check that DRP action is supported"""
     if not drp_action:
         return
-    if drp_action_exists:
-        raise common_exceptions.DrpActionValidationError(
-            f"DRP Action {drp_action} already exists in {class_name}."
-        )
     if drp_action in [
         DrpAction.sale_opt_in.value,
         DrpAction.sale_opt_out.value,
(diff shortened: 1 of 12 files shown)


eastandwestwind:
Lint failing due to known issue on `main`. Will be fixed here- https://github.com/ethyca/fidesops/pull/448

src/fidesops/models/policy.py
```diff
@@ -115,6 +150,42 @@ class Policy(Base):
...
+    @classmethod
+    def create_or_update(cls, db: Session, *, data: Dict[str, Any]) -> FidesopsBase:
+        """Overrides base create or update to add custom error for drp action already exists"""
+        db_obj = None
+        if data.get("id") is not None:
+            # If `id` has been included in `data`, preference that
+            db_obj = cls.get(db=db, id=data["id"])
+        elif data.get("key") is not None:
+            # Otherwise, try with `key`
+            db_obj = cls.get_by(db=db, field="key", value=data["key"])
```
seanpreston, line 162: It's better to start this override by calling the inherited `create_or_update` method (https://github.com/ethyca/fidesops/blob/main/src/fidesops/db/base_class.py) to retrieve the DB class.

seanpreston (changes requested):
Mostly there, a couple of minor changes around how the validation works required. Thanks @eastandwestwind

src/fidesops/models/policy.py
```diff
@@ -70,6 +85,27 @@ class is referenced in multiple database migrations. This class is to be removed
 """
 
 
+def _validate_drp_action(
+    drp_action: Optional[str], drp_action_exists: bool, class_name: str
+) -> None:
+    """Check that DRP action is supported"""
+    if not drp_action:
+        return
+    if drp_action_exists:
```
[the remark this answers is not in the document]
eastandwestwind: mostly just wanted to map that err case to the `DrpActionValidationError` class so that it's handled the same way as other invalid DRP actions. I can remove and just use the base DB exception, if you think that's better?

src/fidesops/models/policy.py
eastandwestwind: wait, confused what you mean by call the base `create_or_update`, since that actually creates/updates in the db, and I need to add pre-logic to validate DRP action

eastandwestwind:
Thanks for the review @seanpreston! Just need some additional input from you, above

src/fidesops/models/policy.py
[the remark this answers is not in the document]
seanpreston: I think that's better in this case as it makes the code more readable, and saves us two database interactions.

src/fidesops/models/policy.py
seanpreston: When you override `create_or_update` on `FidesopsUser`, you can still override `FidesopsBase.create_or_update` e.g.

```
class FidesopsUser:
    ...

        @classmethod
    def create_or_update(cls, db: Session, *, data: Dict[str, Any]) -> FidesopsBase:
        """Validate this object's data before deferring to the superclass on update"""
        instance = super().create(db=db, data=data)
        # mutate instance further
        return instance
```

eastandwestwind: So I won't actually call `super().create_or_update()` from `Policy` but rather `FidesopsBase.get_by_key_or_id()`?

conceptualshark:
Spoke with @adriaaaa about the docs changes - I ended up doing some cleanup and will break out some additional documentation in a new ticket, so this is good from me. 👍

src/fidesops/models/policy.py
```diff
@@ -115,6 +144,19 @@ class Policy(Base):
         backref="policies",
     )  # Which client created the Policy
 
+    @classmethod
+    def create_or_update(cls, db: Session, *, data: Dict[str, Any]) -> FidesopsBase:
+        """Overrides base create or update to add custom error for drp action already exists"""
+        db_obj = cls.get_by_key_or_id(db=db, data=data)
```
seanpreston, line 150: Thanks for making this change!

eastandwestwind:
Ok, ready to go again @seanpreston!

seanpreston (approved):

PR #464, opened 2022-05-05
updates to saas config guide doc to clarify <masked_object_fields> behavior

### LINKED ISSUE #459: <masked_object_fields> docs example should include trailing comma or be at end of request body
### Docs Update Description

A minor "bug" in saas config documents that tripped me up in some of my own testing.

Our `<masked_object_fields>` placeholder is replaced by a JSON serialization of fields that does include a trailing comma (`,`). This means the example provided leads users to a config that produces invalid JSON in their request at runtime.

The example should be updated to either include a `,` after the `<masked_object_fields>` placeholder, or the placeholder should be at the end of the request body (I prefer the former, as it gives more clarity on the behavior). Additionally, it may be good to provide a brief warning/info note about the fact that trailing commas are not included in the placeholder substitution, and therefore need to be included in the template.

old example:
```
body: '{
    "properties": {
      <masked_object_fields>
      "user_ref_id": <user_ref_id>            
    }
  }'
```
proposed new example:
```
body: '{
    "properties": {
      <masked_object_fields>,
      "user_ref_id": <user_ref_id>            
    }
  }'
```
(note the `,` following `<masked_object_fields>`)
### Additional context

One other potential "solution" I thought of was to update the placeholder substitution to include a trailing `,` after the fields, but I think that ultimately decreases flexibility or requires a bunch of string introspection that's probably not necessary. I think our current implementation is effective, so long as we've got this nuance clearly documented.

CC @ethyca/docs-authors

adamsachs:
@ethyca/docs-authors  - thought i'd just give a shot at the update for #459 because it was quick and i was already thinking about it, but please feel free to suggest different wording and/or close the PR to make a different change if you think that'd be better

# Ticket
Fixes #459


docs/fidesops/docs/guides/saas_config.md
```diff
@@ -493,7 +493,7 @@ update:
           direction: from
 ```
 
-Fidesops will replace the `<masked_object_fields>` placeholder with the result of the policy-driven masking service, for example `{'company': None, 'email': None}`.
+Fidesops will replace the `<masked_object_fields>` placeholder with the result of the policy-driven masking service, for example `'company': None, 'email': None`. Note that neither enclosing curly brackets (`{` `}`) nor a trailing comma (`,`) are included as part of the replacement string generated by Fidesops,
```
conceptualshark, line 496: Just a copy edit to replace the comma at the end of this line with a full stop `.` - otherwise this is good to go!

conceptualshark (changes requested):

docs/fidesops/docs/guides/saas_config.md
adamsachs: whoops, good catch - thanks :)

conceptualshark (approved):

PR #466, opened 2022-05-05
use black to format tests

seanpreston:
# Purpose

Running `black` against `tests/` will stop us seeing arbitrary diffs in test change PRs. As identified here: https://github.com/ethyca/fidesops/pull/465#discussion_r866027941.


Makefile
```diff
@@ -140,8 +140,8 @@ pytest-saas: compose-build
 .PHONY: black
 black: compose-build
 	@echo "Running black formatting against the src/ directory..."
-	@docker-compose run $(IMAGE_NAME) \
-	black src/
+	@docker-compose run $(IMAGE_NAME) black src/
```
TheAndrewJackson, line 143: Would it be possible to run just `black .` instead since the projects  `pyproject.toml` has a list of files and directories to exclude? It might be slightly faster than spinning up the project twice to run `black`

seanpreston: The nuance here is that we also only want to run it on `src/` and `tests/` to avoid formatting things like virtualenvs.

PR #470, opened 2022-05-05
Extracts privacy request endpoint logic into separate service for DRP

### LINKED ISSUE #415: [DRP] Privacy Request logic to support DRP endpoints
### Requirements
Refactor privacy request creation and dispatch logic so it can be reused by new APIs. 

Details: Logic in `create_privacy_request` in `privacy_request_endpoints.py` can be refactored out into a shared service, so that it can be re-used by the new DRP endpoint API. Let's try to refactor out as much as we can that can be shared, e.g. identity caching, erasure secrets generation/caching.

Notable differences for DRP privacy request creation to keep in mind when refactoring- 
1. only will execute 1 privacy request, not bulk
2. will get policy by DRP action instead of policy key
3. won't have an option to require manual approval
4. must cache entire request body (for troubleshooting purposes), in addition to just the identity (as is the current function). Request body looks like this: 
```
{
  "meta": {
    "version": "0.5"
  },
  "regime": "ccpa",
  "exercise": [
    "sale:opt-out"
  ],
  "relationships": ["customer", "marketing"],
  "identity": <jwt>... ,
  "status_callback": "https://dsr-agent.example.com/update_status"
}
```

- see full requirements [here](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2259943434/Fidesops+Data+Rights+Protocol+DRP+Implementation+Requirements)

eastandwestwind:
# Purpose
Refactor privacy request logic for future DRP work

# Changes
- Extracts privacy request endpoint logic into separate service for DRP
- Adds schema/enums for DRP request body
- Adds caching utils for DRP request body data

# Checklist

# Ticket

Fixes #415


src/fidesops/service/privacy_request/request_service.py
```diff
@@ -0,0 +1,73 @@
...
+) -> Dict[str, Any]:
+    return {
+        "requested_at": requested_at,
+        "policy_id": policy_id,
+        "status": PrivacyRequestStatus.pending.value,
+    }
+
+
+def create_privacy_request(db: Session, kwargs: Dict[str, Any]):
+    return PrivacyRequest.create(db=db, data=kwargs)
```
eastandwestwind, line 38: ideally let service handle calls to db. Better separation of concerns imo

src/fidesops/service/privacy_request/request_service.py
```diff
@@ -0,0 +1,73 @@
...
+def cache_data(
+    privacy_request: PrivacyRequest,
+    policy: Policy,
+    identity: PrivacyRequestIdentity,
+    encryption_key: Optional[str],
+    drp_request_body: Optional[DrpPrivacyRequestCreate],
+) -> None:
+    # Store identity and encryption key in the cache
+    logger.info(f"Caching identity for privacy request {privacy_request.id}")
+    privacy_request.cache_identity(identity)
```
eastandwestwind, line 50: DRP identity will come in as encrypted JWT. We will decrypt before calling this method so that the format is consistent with current functionality

src/fidesops/service/privacy_request/request_service.py
```diff
@@ -0,0 +1,73 @@
...
+        unique_masking_strategies_by_name.add(strategy_name)
+        masking_strategy = get_strategy(strategy_name, configuration)
+        if masking_strategy.secrets_required():
+            masking_secrets: List[
+                MaskingSecretCache
+            ] = masking_strategy.generate_secrets_for_cache()
+            for masking_secret in masking_secrets:
+                privacy_request.cache_masking_secret(masking_secret)
+    if drp_request_body:
+        privacy_request.cache_drp_request_body(drp_request_body)
```
eastandwestwind, line 73: not sure why GH is showing weird indents here
Screen Shot 2022-05-05 at 2 27 03 PM

seanpreston: Looks the same to me, might be viewport width?

TheAndrewJackson (commented):
It looks good! I think you'll want a second opinion for merging though

src/fidesops/service/privacy_request/request_service.py
eastandwestwind: yeah probably

src/fidesops/service/privacy_request/request_service.py
seanpreston: I'm not sure these are adding any value beyond being housed here? What's the justification for this being better than using the ORM util methods?

seanpreston (commented):
Thanks @eastandwestwind — this looks mostly good, just one question around those DB abstractions.

src/fidesops/service/privacy_request/request_service.py
eastandwestwind: I went ahead and removed the db abstraction for now. Imo, endpoint/controller logic shouldn't be mixed with db calls, maybe a topic of discussion for later. It can reduce chances of duplicate code / circular dependencies in the future. Say, for example, we wish to set a log status whenever we create a privacy request. If we directly called db from the endpoint, we'd have to update every instance of it, vs updating once in the service

seanpreston: Thanks! You're right, I think that's useful. My only concern is that we create these methods now and those conditions never manifest, or they do and somebody decides to go a different way.

seanpreston (approved):

Issue #382, opened 2022-04-12
[User Management] Update individual user


seanpreston:
### Is your feature request related to a specific problem?

The Admin UI needs the ability to update user details:
- first_name
- last_name
- permissions

Separately we also need a password reset endpoint that validates the user's current password, and updates it to a new value.

### Describe the solution you'd like

`PATCH /api/v1/user/:user_id/`
```
{
  "first_name": "...",
  "last_name": "...",
}
```

Aha! Link: https://ethyca.aha.io/features/OPS-85

PR #471, opened 2022-05-05
[#382] Update User API Endpoints

seanpreston:
# Purpose

This PR adds an API endpoints to:
- update fields on the user model
- reset users passwords

# Changes

- `/api/v1/user/{user_id}` — updates the currently logged-in user's first and last names.

```
{
  "first_name": "...",
  "last_name": "..."
}
```

- `/api/v1/user/{user_id}/reset-password` — updates the currently logged-in user's password if `old_password` matches the user's current password at the time of the update.

```
{
  "old_password": "...",
  "new_password": "..."
}
```

# Checklist

# Ticket

Fixes #

### DIFF
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -48,6 +48,7 @@ from fidesops.api.v1.scope_registry import (
     PRIVACY_REQUEST_READ,
     USER_READ,
     USER_DELETE,
+    USER_PASSWORD_RESET,
 )
 
 logger = logging.getLogger(__name__)
@@ -120,7 +121,6 @@ def _validate_current_user(user_id: str, user_from_token: FidesopsUser) -> bool:
 def update_user(
     *,
     db: Session = Depends(deps.get_db),
-    current_user: FidesopsUser = Depends(get_current_user),
     user_id: str,
     data: UserUpdate,
 ) -> FidesopsUser:
@@ -128,16 +128,20 @@ def update_user(
     Update a user given a `user_id`. By default this is limited to users
     updating their own data.
     """
-    _validate_current_user(user_id, current_user)
+    user = FidesopsUser.get(db=db, id=user_id)
+    if not user:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND, detail=f"User with id {user_id} not found."
+        )
 
-    current_user.update(db=db, data=data.dict())
-    logger.info(f"Updated user with id: '{current_user.id}'.")
-    return current_user
+    user.update(db=db, data=data.dict())
+    logger.info(f"Updated user with id: '{user.id}'.")
+    return user
 
 
 @router.post(
     urls.USER_PASSWORD_RESET,
-    dependencies=[Security(verify_oauth_client, scopes=[USER_UPDATE])],
+    dependencies=[Security(verify_oauth_client, scopes=[USER_PASSWORD_RESET])],
     status_code=HTTP_200_OK,
     response_model=UserResponse,
 )
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -156,7 +156,7 @@ def update_user_password(
 
     if not current_user.credentials_valid(data.old_password):
         raise HTTPException(
-            status_code=HTTP_403_FORBIDDEN, detail="Incorrect password."
+            status_code=HTTP_401_UNAUTHORIZED, detail="Incorrect password."
         )
 
     current_user.update_password(db=db, new_password=data.new_password)
--- a/src/fidesops/api/v1/endpoints/user_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/user_endpoints.py
@@ -16,6 +16,7 @@ from starlette.status import (
     HTTP_201_CREATED,
     HTTP_204_NO_CONTENT,
     HTTP_400_BAD_REQUEST,
+    HTTP_401_UNAUTHORIZED,
     HTTP_403_FORBIDDEN,
     HTTP_404_NOT_FOUND,
 )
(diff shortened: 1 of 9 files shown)


src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -89,6 +96,75 @@ def create_user(
...
+    if not user_from_token:
+        raise HTTPException(
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"User with ID {user_id} does not exist.",
+        )
+
+    if user_id != user_from_token.id:
+        raise HTTPException(
+            status_code=HTTP_401_UNAUTHORIZED,
+            detail=f"You are only authorised to update your own user data.",
```
LKCSmith, line 109: Just checking here (sorry if I'm reading this wrong!), but I think the idea is that if the user has those admin privileges, then they can update a user's data as well, not just that user themselves. @adriaaaa can you confirm?

src/fidesops/util/oauth_util.py
```diff
@@ -76,6 +76,20 @@ async def verify_oauth_client(
...
+    authorization: str = Security(oauth2_scheme),
+    db: Session = Depends(deps.get_db),
+) -> FidesopsUser:
+    """A wrapper around verify_oauth_client that returns that client's user if one exsits."""
+    client = await verify_oauth_client(
+        security_scopes=security_scopes,
+        authorization=authorization,
+        db=db,
+    )
+    return client.user
```
seanpreston, line 90: FYI @TheAndrewJackson this is what I ended up implementing to restrict editing data for only one user, as opposed to using a new scope as I suggested before.

src/fidesops/api/v1/endpoints/user_endpoints.py
adriaaaa: Which admin privs specifically @LKCSmith ? If it's the usercreate and userupdate then yes that user should have the ability to create/update other users.

src/fidesops/api/v1/endpoints/user_endpoints.py
```diff
@@ -89,6 +96,75 @@ def create_user(
     return user
 
 
+def _validate_current_user(user_id: str, user_from_token: FidesopsUser) -> bool:
```
TheAndrewJackson, line 99: This is a minor nit. It looks like the 2 functions that call `_validate_current_user` don't use the returned `True` value.  The `return True` statement could safely be removed. I could see future code using the return boolean in the future though. 
```suggestion
def _validate_current_user(user_id: str, user_from_token: FidesopsUser) -> None:
```

src/fidesops/db/base_class.py
```diff
@@ -234,6 +234,10 @@ def delete_all(cls, db: Session) -> int:  # pylint: disable=W0622
         deleted_count = db.query(cls).delete()
         return deleted_count
 
+    def refresh_from_db(self, db: Session) -> FidesopsBase:
```
TheAndrewJackson, line 237: I like this! I've recently queried for records that I just updated

src/fidesops/api/v1/endpoints/user_endpoints.py
LKCSmith: @adriaaaa From the mocks, it looks an admin can update a user's name, password, and privileges?

src/fidesops/util/oauth_util.py
TheAndrewJackson: This is interesting. When `Depends(deps.get_db)` is called within this function does it reuse the same database connection from the routes that call `get_current_user`?

For example `update_user_password` has both `db: Session = Depends(deps.get_db)` and `current_user: FidesopsUser = Depends(get_current_user)` which both call `deps.get_db`.

TheAndrewJackson: This is probably minor but the route decorator is also calling `verify_oauth_client` with `Security(verify_oauth_client, scopes=[USER_UPDATE])`. It looks like this work is being duplicated a little. It's probably negligible on performance since these routes won't be called that much

TheAndrewJackson (commented):
It looks good! I just have a few questions before merging

src/fidesops/api/v1/endpoints/user_endpoints.py
seanpreston: Change made 👍🏻

src/fidesops/util/oauth_util.py
seanpreston: > When Depends(deps.get_db) is called within this function does it reuse the same database connection from the routes that call get_current_user?

Dependency injection will pull a new connection from the session returned by the `get_db_session` function, so they aren't guaranteed to be the same connection, just from the same connection pool.

> It looks like this work is being duplicated a little.

It is! I didn't want to omit that security dependency entirely for readability.

src/fidesops/db/base_class.py
seanpreston: There's an intricacy here in the way DB sessions work that I'll explain.

The data in a DB session isn't guaranteed to be the same as in the database. In the event that another session mutates that same data that's currently held in another session, the data in that session becomes out of date. This manifests in our API tests because we have one session scoped to the test, and another that exists on the server within the API endpoint itself. What's happening is the test-scoped session creates data that the endpoint scoped session then mutates, and the test-scoped session no longer has accurate data. Without an effective refresh our assertions past this time would fail. This is why we also call `db.expunge(...)` before we call this function, to manually purge this object from the session so it's refreshed from the database.

There's a rabbit hole to go down here, that if we call `db.expunge(self)` within `refresh_from_db(...)` the following error is raised:

```
sqlalchemy.exc.InvalidRequestError: Instance <FidesopsUser at 0x7f5d4c70a640> is not present in this Session
```

which would imply the instance isn't part of the session at all, when if you call this same method within a `pdb.set_trace()`, the method works without issue.

src/fidesops/api/v1/scope_registry.py
```diff
@@ -1,3 +1,6 @@
+from fidesops.api.v1.urn_registry import USER_PASSWORD_RESET
```
TheAndrewJackson, line 1: Is there a particular reason for importing the `USER_PASSWORD_RESET` urn at the top? It looks line it gets reassigned on line 54 to the scope

src/fidesops/db/base_class.py
TheAndrewJackson: That's pretty interesting! So if `db.expunge(self)` isn't called `refresh_from_db(...)` won't get the most update to date data from the DB? I guess I assumed that the new query would pull in data directly from the DB instead of the session.

src/fidesops/api/v1/scope_registry.py
seanpreston: That was my IDE getting up to no good. FIxed.

TheAndrewJackson (approved):

PR #486, opened 2022-05-10
[Misc] Add first_name and last_name to createsuperuser script

seanpreston:
# Purpose

- This PR adds `first_name` and `last_name` to the `createsuperuser` script

### To test:

- `make user` and verify that it asks for `first_name` and `last_name`:

```
Enter your username: seanprestonblah
Enter your first name: sean
Enter your last name: preston
Enter your password: 
Enter your password again: 
Superuser 'seanprestonblah' created successfully!
```

- Lookup the user in the database and verify the fields have been added:

```
In [1]: from fidesops.models.fidesops_user import FidesopsUser
In [2]: from fidesops.models.client import ClientDetail
In [3]: from fidesops.db.session import get_db_session
In [4]: db = get_db_session()()
In [5]: user = FidesopsUser.get_by(db=db, field="username", value="seanprestonblah")
In [6]: user.first_name
Out[6]: 'sean'
In [7]: user.last_name
Out[7]: 'preston'
```

### DIFF
--- a/create_superuser.py
+++ b/create_superuser.py
@@ -29,12 +29,23 @@ def get_password(prompt: str) -> str:
     return password
 
 
+def get_input(prompt: str) -> str:
+    """
+    Prompt the user for generic input.
+
+    NB. This method is important to preserve so that
+    our tests can effectively mock the `input` function
+    and data that it returns.
+    """
+    return input(prompt)
+
+
 def collect_username_and_password(db: Session) -> UserCreate:
     """Collect username and password information and validate"""
     username = get_username("Enter your username: ")
     password = get_password("Enter your password: ")
-    first_name = input("Enter your first name: ")
-    last_name = input("Enter your last name: ")
+    first_name = get_input("Enter your first name: ")
+    last_name = get_input("Enter your last name: ")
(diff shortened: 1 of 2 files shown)


create_superuser.py
```diff
@@ -29,16 +29,34 @@ def get_password(prompt: str) -> str:
...
+    return input(prompt)
+
+
 def collect_username_and_password(db: Session) -> UserCreate:
     """Collect username and password information and validate"""
     username = get_username("Enter your username: ")
     password = get_password("Enter your password: ")
+    first_name = get_input("Enter your first name: ")
+    last_name = get_input("Enter your last name: ")
     verify_pass = get_password("Enter your password again: ")
```
pattisdr, line 49: I'd just move asking for `first_name` and `last_name` after `verify_pass` and not before so the password prompts are kept together -

pattisdr (commented):
Just one small thing!

create_superuser.py
seanpreston: Good suggestion! Apologies this was a mistake

pattisdr (approved):

PR #364, opened 2022-04-05
#297 Admin UI v0

elliotbonneville:
# Purpose

Implements v0 of an Admin UI to moderate privacy requests.

image

## Notes

This PR is not ready to be merged yet, but we would like to open up the application and code for review ahead of time. Tests are being written now and final polishing is still happening, so don't spend too much time nitpicking minor details (although any feedback is, of course, appreciated).

- Tests have not been written yet, please ignore test harness during review (for now)
- PR includes documentation changes to the privacy center README, should probably be split out
- PR includes DB seed migration which should be probably be split out

# Changes

- Adds `admin-ui` to clients folder containing Next application

# Checklist

- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md))
- [ ] Good unit test/integration test coverage
- [x] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #297.

PR #487, opened 2022-05-10
[Misc] Adds a script to seed initial privacy requests

seanpreston:
# Purpose

This PR adds a new `make test-data` command to seed our application database with some initial privacy requests, based on the data migration we originally tested the Fidesops Admin UI with (https://github.com/ethyca/fidesops/pull/364/commits/fc4ac38e43f8db1b4d02c7fc954eaa487cb842d0)

### DIFF
--- a/create_test_data.py
+++ b/create_test_data.py
@@ -11,6 +11,12 @@ from fidesops.models.client import ClientDetail
 from fidesops.models.fidesops_user import FidesopsUser
 from fidesops.models.policy import ActionType, Policy, Rule, RuleTarget
 from fidesops.models.privacy_request import PrivacyRequest, PrivacyRequestStatus
+from fidesops.models.storage import StorageConfig, ResponseFormat
+from fidesops.schemas.storage.storage import (
+    FileNaming,
+    StorageDetails,
+    StorageType,
+)
 from fidesops.util.data_category import DataCategory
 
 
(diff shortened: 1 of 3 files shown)


create_test_data.py
```diff
@@ -0,0 +1,97 @@
...
+def _create_policy(
+    db: orm.Session,
+    action_type: str,
+    client_id: str,
+) -> Policy:
+    rand = string.ascii_lowercase[:5]
+    policy = Policy.create(
+        db=db,
+        data={
+            "name": f"example {action_type} policy {rand}",
```
TheAndrewJackson, line 29: Could we also add a `download` and `delete` policy to this function? The privacy center uses these policies by default and they need to be recreated each time the database is reset. I found myself doing that a few times while developing frontend.

seanpreston: By `download` do you mean `access`?

TheAndrewJackson: No, I do mean `download` and `delete`. It looks like the privacy center has a config file set up to use those as the policy keys. Here is the mentioned [file](https://github.com/ethyca/fidesops/blob/main/clients/privacy-center/config/config.json#L7-L30).

```json
  "actions": [
    {
      "policy_key": "download",
      "icon_path": "/download.svg",
      "title": "Download your data",
      "description": "We will email you a report of the data related to your account.",
      "identity_inputs": {
        "name": "optional",
        "email": "required",
        "phone": "optional"
      }
    },
    {
      "policy_key": "delete",
      "icon_path": "/delete.svg",
      "title": "Delete your data",
      "description": "We will delete all of your account data. This action cannot be undone.",
      "identity_inputs": {
        "name": "optional",
        "email": "required",
        "phone": "optional"
      }
    }
  ]
```

TheAndrewJackson: Are `access` and `erasure` policy keys seeded already?

seanpreston: > Are access and erasure policy keys seeded already?

They aren't

TheAndrewJackson (approved):

PR #491, opened 2022-05-10
Publish fidesops to PyPI

### LINKED ISSUE #481: Publish fidesops to PyPI

- Fidesops is not easily installed via `pip`
- When installing fidesops in another project directly via git, https://github.com/ethyca/fidesops/tree/main/src/fidesops/api is not included
  - It's missing an `__init__.py` file

1. Fidesops should be published to PyPI in it's entirety
2. An `__init__.py` file should be added to the https://github.com/ethyca/fidesops/tree/main/src/fidesops/api directory

Fidesops can be installed directly via git:
```sh
pip install git+https://github.com/ethyca/fidesops.git@1.4.1
```

### Additional context

The enterprise team is beginning work on extending fidesops in https://github.com/ethyca/fidesops-plus. For this to be done gracefully, that repository should be able to install a pinned version of fidesops that is easily `import`able in source code. Ex:
```sh
# requirements.py

fidesops==1.4.1
...
```
```python
# src/main.py

from fidesops.main import app as fidesops
...
```

PSalant726:
# Purpose

Fidesops is not currently published to PyPI. This makes it difficult to install and use fidesops as a package in other Python projects. As work proceeds on https://github.com/ethyca/fidesops-plus, installing fidesops via git becomes cumbersome, as the enterprise team would like to maintain feature parity by pinning the imported version of fidesops.

# Changes

- Add missing `__init__.py` files (12 total)
  - And fix all ensuing static analysis errors
- Add a GitHub Actions workflow to publish fidesops to PyPI whenever a new tag is created (releases)
  - [x] The `PYPI_TOKEN` secret needs to be configured on this repository
- Alphabetize `requirements.txt`
- Add the [`versioneer`](https://pypi.org/project/versioneer/) dependency
- Install `versioneer`
  - And use it in `setup.py`

# Checklist

# Ticket

Fixes #481


.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,23 @@
+name: Publish to PyPI
+
+on:
+  push:
+    tags:
+      - "*"
```
PSalant726, line 6: There may be a better trigger for this workflow based on the preferred release process.

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Install Twine
+        run: pip install twine
+
+      - name: Twine Upload
+        run: |
+          python setup.py sdist
+          twine upload dist/*
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
```
PSalant726, line 23: This secret needs to be added to this repository; I don't have the necessary privileges.

requirements.txt
```diff
@@ -1,34 +1,35 @@
...
-multidimensional_urlencode==0.0.4
\ No newline at end of file
+sqlalchemy-redshift==0.8.8
+sqlalchemy-stubs==0.4
+SQLAlchemy-Utils==0.37.8
+sqlalchemy==1.4.14
+starlette~=0.14.2
+Unidecode==1.2.0
+uvicorn~=0.13.4
+versioneer==0.19
```
PSalant726, line 35: This line is the only addition in this file. The large diff is due to the alphabetization change, which will pay for itself in saved time (after about 200 years).

PSalant726 (commented):
The large number of added files is due mostly to the new `__init__.py` files that were previously missing, and the files added by `versioneer`.

.github/workflows/publish_to_pypi.yml
ThomasLaPiana: @PSalant726 check fidesctl we do it in a smarter way for the docs -> https://github.com/ethyca/fides/blob/main/.github/workflows/publish_docs.yaml#L7

PSalant726: Updated. @seanpreston does this now align with the current release process?

ThomasLaPiana:
I highly highly strongly recommend adopting the `fidesctl` docker setup so that in the production image the package is installed as if it was coming directly from pypi, it will catch so many potential packaging errors (such as no init files)

.github/workflows/publish_to_pypi.yml
seanpreston: Secret added 👍

pyproject.toml
```diff
@@ -39,22 +39,28 @@ exclude = '''
...
+    "no-self-use",
     "raise-missing-from",
+    "too-few-public-methods",
     "too-many-ancestors",
-    "fixme"]
+    "too-many-arguments",
+    "unnecessary-comprehension",
+    "unsubscriptable-object", # Otherwise throws errors on certain Type annotations
+    "unused-argument",
+]
```
seanpreston, line 63: Any reason you've added these extra lint rules? We usually talk change management in our retrospectives.

ThomasLaPiana:
in fidesctl we use multi-stage builds for our dockerfile. This means that when we run our tests we use a production image with `fidesctl` installed from a clean build, instead of symlinked. This makes sure that our tests run against a package that has been installed as if it were a `pip install fidesctl`.

This is definitely for another issue, but is a best practice that should be done in a follow-up to this one

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -365,7 +368,7 @@ def get_request_status(
     *,
     db: Session = Depends(deps.get_db),
     params: Params = Depends(),
-    id: Optional[str] = None,
+    request_id: Optional[str] = None,
```
seanpreston, line 371: I think this might be a breaking change as it requires us to the param name in requests sent to the server?

seanpreston:
Hey @PSalant726 — thanks for being thorough, I worry that this PR is a little bloated as you seem to have changed a lot more than is required just to publish the project to PyPi?

pyproject.toml
PSalant726: 1. `broad-except`
	To eliminate linter errors thrown by code like:
	```python
	try:
		do_something()
	except Exception: # thrown here
		handle_any_exception()
	``` 
	Originally seen in several places throughout the `api` and `service` packages.
1. `missing-class-docstring`
	To eliminate linter errors thrown by...missing class docstrings. I don't have the context to author many of these.
1. `missing-function-docstring`
	Same as above.
1. `too-many-arguments`
	IMO, it's a silly guideline rule that happens to be enabled by default. It was throwing for a couple of functions.
1. `unnecessary-comprehension`
	The alternative would be to significantly change logic where this error was thrown, which I was not comfortable doing (at least not in this PR).

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
PSalant726: None of the tests were failing as a result. If you can point me to the coverage, or lack thereof, I can investigate.

seanpreston: Here's an example of a false positive: https://github.com/ethyca/fidesops/blob/main/tests/api/v1/endpoints/test_privacy_request_endpoints.py#L510-L532

Since this is a query param, a lack thereof is not actually filtering the requests and simply returning a page.

Another here that explicitly tests the param: https://github.com/ethyca/fidesops/blob/main/tests/api/v1/endpoints/test_privacy_request_endpoints.py#L557-L600

I think this is only still passing because there's exactly one PrivacyRequest in the system at the time the test runs.

src/fidesops/service/pagination/pagination_strategy_factory.py
```diff
@@ -39,7 +46,7 @@ def get_strategy(
     Returns the strategy given the name and configuration.
     Raises NoSuchStrategyException if the strategy does not exist
     """
-    if strategy_name not in SupportedPaginationStrategies.__members__:
+    if not SupportedPaginationStrategies.__contains__(strategy_name):
```
seanpreston, line 49: What's the rationale for moving away from `x in y`?

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -601,6 +600,34 @@ def test_filter_privacy_requests_by_status(
...
+        privacy_request.id = new_request_id
+        privacy_request.save(db)
+
+        response = api_client.get(
+            url + f"?request_id={new_request_id}", headers=auth_header
+        )
+        assert response.status_code == status.HTTP_200_OK
+        resp = response.json()
+        assert len(resp["items"]) == 1
+        assert resp["items"][0]["id"] == privacy_request.id
```
seanpreston, line 629: Thankyou for adding this 🙏

versioneer.py
```diff
@@ -0,0 +1,1856 @@
...
+    return errors
+
+
+if __name__ == "__main__":
+    cmd = sys.argv[1]
+    if cmd == "setup":
+        errors = do_setup()
+        errors += scan_setup_py()
+        if errors:
+            sys.exit(1)
```
seanpreston, line 1856: Does Versioneer need to be included like this vs. being a pip import itself?

PSalant726: `versioneer` was added as an import: https://github.com/ethyca/fidesops/blob/f9cf2f06f2f41ff0737876b6997cfb41db8a8587/requirements.txt#L36

This script is the result of installing it with `versioneer install`. See https://github.com/python-versioneer/python-versioneer#theory-of-operation for more info.

seanpreston (approved):
Thanks @PSalant726

PR #496, opened 2022-05-11
416 Adds DRP exercise endpoint

### LINKED ISSUE #416: [DRP] Add support for POST /exercise endpoint
### Requirements

1. MUST expose a POST /api/v1/drp/exercise endpoint (similar to POST /api/v1/privacy-request)
3. MUST support all required request body fields as defined by the [specification](https://github.com/consumer-reports-digital-lab/data-rights-protocol#202-post-exercise-data-rights-exercise-endpoint) (meta, exercise, identity); note that regime, relationships, and status_callback are covered by other requirements below
5. MUST lookup the Fidesops policy to use for the request based on the “DRP action” (instead of based on a Fidesops policy_key)
7. SHOULD throw an error if multiple “DRP actions” are specified in the exercise field; Fidesops requests currently only support a single policy per request
9. MUST decode the DRP identity JWT token and validate the included identity fields based on the [specification](https://github.com/consumer-reports-digital-lab/data-rights-protocol#304-schema-identity-encapsulation) (name, email, phone_number, address)
11. MUST map all valid Fidesops request status to DRP “Request Statuses” as defined in the [specification](https://github.com/consumer-reports-digital-lab/data-rights-protocol#303-schema-status-of-a-data-subject-exercise-request) (e.g. open, in_progress, fulfilled)
13. MUST create the privacy request otherwise normally, using the given DRP identity and DRP action 
15. MUST return a JSON response matching the “Exercise Status” schema defined in the DRP [specification](https://github.com/consumer-reports-digital-lab/data-rights-protocol#303-schema-status-of-a-data-subject-exercise-request) (request_id, status, received_at, etc.)
17. SHOULD support the status_callback 
19. SHOULD cache the contents of the request body in the identity cache (for troubleshooting)

eastandwestwind:
# Purpose
Adds DRP exercise endpoint

# Changes
- Adds endpoint for DRP exercise
- Refactors DRP status endpoint into DRP-specific endpoint file
- Adds unit tests for DRP exercise
- Adds DRP folder to postman collection 
- Additionally, removes ignore_missing_imports flag for mypy

# Checklist

# Ticket

Fixes #416


src/fidesops/models/privacy_request.py
```diff
@@ -186,10 +186,17 @@ def cache_drp_request_body(self, drp_request_body: DrpPrivacyRequestCreate) -> N
         drp_request_body_dict: Dict[str, Any] = dict(drp_request_body)
         for key, value in drp_request_body_dict.items():
             if value is not None:
-                cache.set_with_autoexpire(
-                    get_drp_request_body_cache_key(self.id, key),
-                    value,
-                )
+                # handle nested dict/objects
+                if not isinstance(value, (bytes, str, int, float)):
```
eastandwestwind, line 190: cache doesn't handle vals as bytes/str/int/float. Perhaps we should instead cache the whole req body as a str (`json.loads()`) instead of separating the top level props as cache keys. But this works as far as caching for debugging purposes only.

src/fidesops/schemas/drp_privacy_request.py
```diff
@@ -22,7 +24,7 @@ class DrpPrivacyRequestCreate(BaseSchema):
 
     meta: DrpMeta
     regime: Optional[DrpRegime]
-    exercise: DrpAction
+    exercise: List[DrpAction]
```
eastandwestwind, line 27: DRP spec dictates that this is a list. So we'll take in a list even though we don't support > 1 item

eastandwestwind:
Verified both endpoints manually in postman:

Screen Shot 2022-05-11 at 2 29 41 PM

Screen Shot 2022-05-11 at 2 27 03 PM

src/fidesops/core/config.py
```diff
@@ -120,6 +120,7 @@ class SecuritySettings(FidesSettings):
     AES_ENCRYPTION_KEY_LENGTH: int = 16
     AES_GCM_NONCE_LENGTH: int = 12
     APP_ENCRYPTION_KEY: str
+    DRP_JWT_SECRET: Optional[str]
```
eastandwestwind, line 123: should I also add this to `fidesops.toml` as an empty str?

src/fidesops/api/v1/endpoints/drp_endpoints.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Given a drp privacy request body, create and execute
+    a corresponding Fidesops PrivacyRequest
+    """
+
+    jwt_key: str = config.security.DRP_JWT_SECRET
+    if jwt_key is None:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail="JWT key must be provided",
```
seanpreston, line 60: This isn't really a `400` since if this variable isn't in the config it represents a server misconfiguration. `501_NOT_IMPLEMENTED` or `500_SERVER_ERROR` might be best.

src/fidesops/api/v1/endpoints/drp_endpoints.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+    except common_exceptions.RedisConnectionError as exc:
+        logger.error("RedisConnectionError: %s", exc)
+        # Thrown when cache.ping() fails on cache connection retrieval
+        raise HTTPException(
+            status_code=HTTP_424_FAILED_DEPENDENCY,
+            detail=exc.args[0],
+        )
+    except Exception as exc:
```
seanpreston, line 115: These `except` blocks should surround only the minimal code, and we should avoid using broad excepts unless there's a specific reason?

src/fidesops/api/v1/endpoints/drp_endpoints.py
```diff
@@ -0,0 +1,154 @@
...
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"Privacy request with ID {request_id} does not exist, or is not associated with a data rights protocol action.",
+        )
+
+    logger.info(f"Privacy request with ID: {request_id} found for DRP status.")
+    return PrivacyRequestDRPStatusResponse(
+        request_id=request.id,
+        received_at=request.requested_at,
+        status=DrpFidesopsMapper.map_status(request.status),
+    )
```
seanpreston, line 154: Thanks for moving this into this file!

src/fidesops/api/v1/urn_registry.py
```diff
@@ -88,3 +87,8 @@
 
 # Health URL
 HEALTH = "/health"
+
+# DRP
+DRP = "/drp"
+DRP_EXERCISE = DRP + "/exercise"
+DRP_STATUS = DRP + "/status"
```
seanpreston, line 94: NIT: can we store URLs as absolute strings so that they're easily `cmd + f`-able?

src/fidesops/core/config.py
seanpreston: Yes please!

src/fidesops/service/drp/drp_fidesops_mapper.py
```diff
@@ -0,0 +1,45 @@
+from typing import Dict
+
+from fidesops.models.privacy_request import PrivacyRequestStatus
+from fidesops.schemas.drp_privacy_request import DrpIdentity
+from fidesops.schemas.privacy_request import PrivacyRequestDRPStatus
+from fidesops.schemas.redis_cache import PrivacyRequestIdentity
+
+
+class DrpFidesopsMapper:
```
seanpreston, line 9: Nice implementation 🙌

src/fidesops/service/drp/drp_fidesops_mapper.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Currently, both email and phone_number identity props map 1:1 to the corresponding
+        Fidesops identity props in PrivacyRequestIdentity. This may not always be the case.
+        This class also allows us to implement custom logic to handle "verified" id props.
+        """
+        identity_kwargs = {
+            "email": drp_identity.email,
+            "phone_number": drp_identity.phone_number,
+        }
+        return PrivacyRequestIdentity(**identity_kwargs)
```
seanpreston, line 25: Do we want to discard any extra attributes on `DrpIdentity` silently?

tests/api/v1/endpoints/test_drp_endpoints.py
```diff
@@ -0,0 +1,288 @@
...
+        self,
+        run_access_request_mock,
+        url,
+        db,
+        api_client: TestClient,
+        policy_drp_action,
+        cache,
+    ):
+
+        config.security.DRP_JWT_SECRET = "secret"
```
seanpreston, line 47: It's probably worth setting a default for this value that's overridden on deployment [in the same way we do with the root client](https://github.com/ethyca/fidesops/blob/main/fidesops.toml#L20-L21). That way we won't need to override it here?

src/fidesops/api/v1/endpoints/drp_endpoints.py
eastandwestwind: ah right, good point

src/fidesops/api/v1/endpoints/drp_endpoints.py
eastandwestwind: this was to be consistent with what we have for POST `"/privacy-request"`. But I can remove it, just lemme know your preference

src/fidesops/api/v1/urn_registry.py
eastandwestwind: should we eventually update the rest of the `urn_registry` to be `cmd + f`-able in the same way?

src/fidesops/models/privacy_request.py
```diff
@@ -186,10 +186,17 @@ def cache_drp_request_body(self, drp_request_body: DrpPrivacyRequestCreate) -> N
...
+                    cache.set_with_autoexpire(
+                        get_drp_request_body_cache_key(self.id, key),
+                        repr(value),
```
seanpreston, line 193: I know this probably feels hacky but good choice in effort/reward trade-off if this is only for debug purposes as you say.

Looks like `value` can only ever be a key from [this schema](https://github.com/ethyca/fidesops/blob/579f028faa620b04c34e58187c1bb212f62bf535/src/fidesops/schemas/drp_privacy_request.py#L22), which should be just fine.

If we want to do more with this data though let's make sure we're breaking it down and storing it by key. We could write a util method to handle that automatically for Pydantic schemas for instance.

seanpreston (approved):
LGTM — would be nice if we could address the URL nit and JWT config var before merging!

src/fidesops/service/drp/drp_fidesops_mapper.py
eastandwestwind: I'll surface a warning log message using a validator or a check within this mapper

eastandwestwind:
Thanks for the review @seanpreston! This is good to go.

I've added a warning message in case we don't support the specific identity props passed in: https://github.com/ethyca/fidesops/pull/496/files#diff-e2fc0ccbfb0913f3f97fd1f425377e29beb158ea9581cc5e171e044b8927c4a4R30

Issue #490, opened 2022-05-10
[SaaS Connector] OAuth2 Authentication Strategy


galvana:
### Is your feature request related to a specific problem?

Create the schema, endpoints, and logic to support OAuth2 authentication for SaaS connectors
https://ethyca.atlassian.net/wiki/spaces/OSS/pages/2279538853/SaaS+Connector+OAuth2+Authentication

### Describe the solution you'd like

- New API endpoints
  - /api/v1/connection/{connection_key}/authorize
  - /api/v1/oauth/callback
- Create schema for OAuth2 configuration
- OAuth2 authentication strategy
  - Create logic inside of new `oauth2_authentication_strategy` to create the various OAuth2 endpoints from the new schema

### Additional context
Use ngrok, Localtunnel, or any other reverse-proxy solution to forward the callback response back to your local Fidesops instance for development. The callback server portion of this workflow will be covered in (https://github.com/ethyca/fidesops/issues/489)

Depends on: https://github.com/ethyca/fidesops/issues/488

PR #504, opened 2022-05-12
Refactoring SaaS authentication in preparation for OAuth2

### LINKED ISSUE #488: [SaaS Connector] Use strategy factory pattern for authentication strategies

We need to support multiple types of authentication for our SaaS connectors. We currently have basic and bearer authentication but would like to add OAuth2 and API key authentication (https://github.com/ethyca/fidesops/issues/334). To have a better foundation for these additional authentication methods we need to move to a code structure that will allow for the new authentication logic to be properly encapsulated in its own implementation separate from the other strategies. We have the currently supported authentication strategies all inside the `add_authentication` function of the `AuthenticatedClient` class. While this was acceptable for the first phase of SaaS development, we need to clean this up now.

Create the following classes (closely following the strategy factory pattern we're using for post-processors and pagination)
- `authentication_strategy_factory`
- `authentication_strategy` base class
- `authentication_configuration` base class for configuration validation
- specific implementations of the `authentication_strategy` base class (`basic_authentication_strategy`, `bearer_authentication_strategy`)

galvana:
# Purpose
Refactoring the SaaS authentication logic in preparation for OAuth2 (https://github.com/ethyca/fidesops/issues/490)

# Changes
- Created `authentication_strategy_factory` and separate authentication strategies for basic, bearer, and "query_param" authentication
- Simplified and standardized authorization configuration (by using placeholders instead of `ConnectorParamRefs`)
- Simplified auth strategy names from `basic_authentication` and `bearer_authentication` to just `basic` and `bearer`, and to make more consistent with strategy naming conventions for pagination, filtering, masking, etc.
- Updated SaaS configs for existing connectors and updated the associated Postman collections
- Fixed minor typos in the SaaS documentation

# Checklist

# Ticket

Fixes #488


data/saas/config/hubspot_config.yml
```diff
@@ -10,13 +10,12 @@ saas_config:
 
   client_config:
```
galvana, line 11: In order to make the `client_config` consistent with the notation we use for `SaaSRequest`, I updated `client_config` to also use value placeholders (`<domain>`,`<api_key>`). The placeholder values for authentication can only come from `connector_params` so there is no need to define that here.

data/saas/config/mailchimp_config.yml
```diff
@@ -11,15 +11,12 @@ saas_config:
 
   client_config:
     protocol: https
-    host:
-      connector_param: domain
+    host: <domain>
     authentication:
-      strategy: basic_authentication
```
galvana, line 17: I thought it was redundant to specify "authentication" in the strategy name so I made the following changes:

- `basic_authentication` -> `basic`
- `bearer_authentication` -> `bearer`

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -1,34 +1,39 @@
-from json import JSONDecodeError
```
galvana, line 1: Running `isort` on all the files I'm modifying

tests/fixtures/saas/segment_fixtures.py
```diff
@@ -203,7 +203,7 @@ def segment_erasure_data(
                 "The user endpoint did not return the required data for testing during the time limit"
             )
         retries -= 1
-        time.sleep(5)
+        time.sleep(10)
```
galvana, line 206: Needed to give Segment more time for the data to be available

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -81,7 +65,11 @@ def get_authenticated_request(
             params=request_params.query_params,
             data=request_params.body,
         ).prepare()
-        return self.add_authentication(req, self.client_config.authentication)
+
+        authentication = self.client_config.authentication
+        return get_authentication_strategy(
+            authentication.strategy, authentication.configuration
+        ).add_authentication(req, self.secrets)
```
PSalant726, line 72: It's a nit, but I think this reads more clearly as:

```suggestion
        auth_strategy = get_authentication_strategy(
            self.client_config.authentication.strategy,
            self.client_config.authentication.configuration,
        )
        
        return auth_strategy.add_authentication(req, self.secrets)
```

Overall though, I don't love that `add_authentication` is a method on `AuthenticationStrategy`, since it should really be on `PreparedRequest`. It just seems like strange usage, because you're adding authentication to the request, not the auth strategy. Obviously we can't make changes to `PreparedRequest` though, and I'm not seeing a better way to improve the usage 😕

src/fidesops/util/saas_util.py
```diff
@@ -165,3 +167,21 @@ def format_body(
...
+
+
+def assign_placeholders(value: str, param_values: Dict[str, Any]) -> Optional[str]:
+    """
+    Finds all the placeholders (indicated by <>) in the passed in value
+    and replaces them with the actual param values
+
+    Returns None if any of the placeholders cannot be found in the param_values
+    """
+    if value and isinstance(value, str):
```
PSalant726, line 179: `value` will always be a string, since it's type is asserted in the function signature. We only have to account for `""`.
```suggestion
    if value:
```

src/fidesops/util/saas_util.py
```diff
@@ -165,3 +167,21 @@ def format_body(
...
+        placeholders = re.findall("<(.+?)>", value)
+        for placeholder in placeholders:
+            placeholder_value = param_values.get(placeholder)
+            if placeholder_value:
+                value = value.replace(f"<{placeholder}>", str(placeholder_value))
+            else:
+                return None
```
PSalant726, line 186: This function will still have mutated `value` if the second (or later) `placeholder` has no `placeholder_value` in `param_values`. Is this intended behavior? If not, we should duplicate `value`, mutate our duplicate, and return either that new string or `None`. We should also add test coverage for that case.

Separately, does every instance where `assign_placeholders` is called currently handle the possibility of `None` being returned? It might be more intuitive to instead `raise` a new custom exception.

src/fidesops/util/url_util.py
```diff
@@ -1,4 +1,4 @@
-from urllib.parse import urlsplit, parse_qs, urlencode, urlunsplit
+from urllib.parse import parse_qs, urlencode, urlsplit, urlunsplit
```
PSalant726, line 1: Here's another instance where the only change to the file is a reordering of imports. Does this need to be included in this PR?

tests/fixtures/saas/segment_fixtures.py
PSalant726: Classic Segment 🙄

PSalant726 (changes requested):
Great refactor! Much cleaner than before. My comments are pretty low level.

src/fidesops/service/authentication/authentication_strategy_basic.py
```diff
@@ -0,0 +1,44 @@
...
+            auth=(
+                assign_placeholders(self.username, secrets),
+                assign_placeholders(self.password, secrets),
+            )
+        )
+        return request
+
+    @staticmethod
+    def get_configuration_model() -> StrategyConfiguration:
+        return BasicAuthenticationConfiguration
```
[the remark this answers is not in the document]
PSalant726: Thinking more about this, we can abstract it away to the `AuthenticationStrategy` class, and have it become automatically available on anything that inherits from it 😎

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -11,7 +11,6 @@
     FieldAddress,
     ScalarField,
     CollectionAddress,
-    Field,
```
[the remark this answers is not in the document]
galvana: Reverting change to avoid future confusion

src/fidesops/schemas/saas/strategy_configuration.py
```diff
@@ -91,3 +92,29 @@ class CursorPaginationConfiguration(StrategyConfiguration):
...
     field: str
+
+
+class BasicAuthenticationConfiguration(StrategyConfiguration):
+    """
+    Username and password to add basic authentication to HTTP requests
+    """
+
+    username: str
+    password: Optional[str]
```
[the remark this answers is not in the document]
galvana: Some APIs expect the access_token to be passed in as basic auth without a password 🤷‍♂️

src/fidesops/service/authentication/authentication_strategy.py
```diff
@@ -0,0 +1,25 @@
...
+from requests import PreparedRequest
+
+from fidesops.schemas.saas.strategy_configuration import StrategyConfiguration
+
+
+class AuthenticationStrategy(ABC):
+    """Abstract base class for SaaS authentication strategies"""
+
+    @abstractmethod
+    def get_strategy_name(self) -> str:
```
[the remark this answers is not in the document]
galvana: Makes sense!

src/fidesops/service/authentication/authentication_strategy_basic.py
```diff
@@ -0,0 +1,44 @@
...
+from fidesops.schemas.saas.strategy_configuration import (
+    BasicAuthenticationConfiguration,
+    StrategyConfiguration,
+)
+from fidesops.service.authentication.authentication_strategy import (
+    AuthenticationStrategy,
+)
+from fidesops.util.saas_util import assign_placeholders
+
+STRATEGY_NAME = "basic"
```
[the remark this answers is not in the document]
galvana: Changed to class-level static variable

src/fidesops/service/authentication/authentication_strategy_basic.py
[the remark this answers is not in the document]
galvana: So clean! 😎

src/fidesops/service/authentication/authentication_strategy_factory.py
```diff
@@ -0,0 +1,72 @@
...
+            **configuration
+        )
+        return strategy(configuration=strategy_config)
+    except ValidationError as e:
+        raise FidesopsValidationError(message=str(e))
+
+
+def get_strategies() -> List[AuthenticationStrategy]:
+    """Returns all supported authentication strategies"""
+    return [e.value for e in SupportedAuthenticationStrategies]
```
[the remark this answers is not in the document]
galvana: Changed to `get_strategy_names` and updated the usage on L54

src/fidesops/service/connectors/saas_connector.py
galvana: Changed this part so it's easier to read

src/fidesops/util/url_util.py
galvana: Reverting change

src/fidesops/service/authentication/authentication_strategy_factory.py
```diff
@@ -0,0 +1,70 @@
...
+    Raises NoSuchStrategyException if the strategy does not exist
+    """
+    if not SupportedAuthenticationStrategies.__contains__(strategy_name):
+        valid_strategies = ", ".join(get_strategy_names())
+        raise NoSuchStrategyException(
+            f"Strategy '{strategy_name}' does not exist. Valid strategies are [{valid_strategies}]"
+        )
+    strategy = SupportedAuthenticationStrategies[strategy_name].value
+    try:
+        strategy_config: StrategyConfiguration = strategy.configuration_model()(
```
galvana, line 60: I started implementing your proposed `configuration_model` changes but ran into an issue here. We need configuration_model to be a static method since we won't have an instance of it here.

src/fidesops/service/authentication/authentication_strategy.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @abstractmethod
+    def add_authentication(
+        self, request: PreparedRequest, secrets: Dict[str, Any]
+    ) -> PreparedRequest:
+        """Add authentication to the request"""
+
+    @staticmethod
+    @abstractmethod
+    def get_configuration_model() -> StrategyConfiguration:
```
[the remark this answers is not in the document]
galvana: Keeping the get for now, until we're able to make the change to the other strategies, I didn't want this one to be different from the other ones.

src/fidesops/util/saas_util.py
PSalant726: I see you added `int` as a possible type for `value`, but this line means the function only handles string `value`s. There's no handling for `int`s that I can see. Please add comprehensive test coverage for this function.

src/fidesops/util/saas_util.py
PSalant726: Ok, but the first issue still stands - if multiple `placeholder`s exist, but only a single `placeholder_value`, then `value` itself will still have been mutated when the function returns `None`. It's important that we add test coverage for this helper, so we can assert that it behaves as expected.

src/fidesops/service/authentication/authentication_strategy_basic.py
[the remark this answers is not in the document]
PSalant726: I tried some things locally to abstract away `get_configuration_model` from each subclass of `AuthenticationStrategy`, but I can see that it won't work because of [this hacky code](https://github.com/ethyca/fidesops/blob/5a5a87a030201163f53925edb89e2e0165400b73/src/fidesops/service/authentication/authentication_strategy_factory.py#L58-L63). The other strategy factories do things this way, so it's fine to leave it for now. A larger refactor for another PR 😕

PSalant726 (changes requested):
Looking good - all that's left is some work on the `assign_placeholders` helper.

PR #541, opened 2022-05-19
353- initial infra for fideslog integration

eastandwestwind:
# Purpose
Adds fideslog integration to fidesops.

# Changes
- Adds new Analytics class and associated schema, so we have ability to send event data
- Adds new env var in config for global opt in/out with default as opt out = True (we can adjust as needed in future as we implement more granular user controls
- Updates docs for config

# Checklist
- [ ] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file in appropriate sections

# Ticket

Fixes #


eastandwestwind:
Got some feedback from @PSalant726 on this via screenshare code review. Will work on this Monday

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -86,6 +89,9 @@ TASK_RETRY_DELAY=20
 TASK_RETRY_BACKOFF=2
 REQUIRE_MANUAL_REQUEST_APPROVAL=True
 MASKING_STRICT=True
+
+[user]
+ANALYTICS_OPT_OUT=True
```
PSalant726, line 94: You probably don't want to encourage opt-outs by including `True` as the value in your example config...

fidesops.toml
```diff
@@ -27,3 +27,6 @@ TASK_RETRY_DELAY=1
 TASK_RETRY_BACKOFF=1
 REQUIRE_MANUAL_REQUEST_APPROVAL=false
 MASKING_STRICT=true
+
+[user]
+ANALYTICS_OPT_OUT=true
```
PSalant726, line 32: Same here.

requirements.txt
```diff
@@ -34,3 +34,4 @@ starlette~=0.14.2
 Unidecode==1.2.0
 uvicorn~=0.13.4
 versioneer==0.19
+fideslog==1.1.5
```
PSalant726, line 37: A very talented and good-looking engineer recently alphabetized this file in https://github.com/ethyca/fidesops/pull/491/commits/ecfb63d333daa24d613be73e6689fee2d8adc873 (#491). Please keep it nice 😄

src/fidesops/analytics.py
```diff
@@ -0,0 +1,74 @@
...
+    """`True` if the command was submitted within a Docker container. Default: `False`."""
+    return True
+
+
+def running_on_local_host() -> bool:
+    """For events submitted as a result of making API server requests, `True` if the API server is running on the user's local host. Default: `False`."""
+    return False
+
+
+class Analytics:
```
PSalant726, line 38: I'll reiterate that I don't think this class is necessary, and that you're creating a lot of undue complexity with it. The fideslog SDK is designed to be small, strict, and intuitive. Rather than adding this class + the `FideslogAnalyticsEvent` schema, I think you'll be fine with just:
```python
from fideslog.sdk.python.client import AnalyticsClient
from fideslog.sdk.python.utils import FIDESOPS, generate_client_id

analytics_client = AnalyticsClient(
    client_id=config.user.analytics_id or generate_client_id(FIDESOPS),
    ...
)
```
Then, throughout the fidesops codebase, you'll be able to do:
```python
from fideslog.sdk.python.event import AnalyticsEvent
from fideslog.sdk.python.exceptions import AnalyticsError
from fidesops.analytics import analytics_client

if not config.user.analytics_opt_out:
    try:
        analytics_client.send(AnalyticsEvent(...))
    except AnalyticsError as err:
        logger.warning(f"Failed to send analytics event: {err}")
```

src/fidesops/core/config.py
```diff
@@ -170,13 +170,20 @@ class Config:
         env_prefix = "FIDESOPS__SECURITY__"
 
 
+class UserSettings(FidesSettings):
+    """Configuration settings for Analytics variables."""
+
+    ANALYTICS_OPT_OUT: bool = True
```
PSalant726, line 176: This should default to the "unset" consent state, and other code should ensure that the user is prompted for consent as soon as possible.
```suggestion
    ANALYTICS_OPT_OUT: Optional[bool]
```

src/fidesops/schemas/analytics.py
```diff
@@ -0,0 +1,25 @@
+from typing import Optional, List, Dict
+
+from fidesops.schemas.base_class import BaseSchema
+
+
+class FideslogAnalyticsEvent(BaseSchema):
```
PSalant726, line 6: If you check out [the fideslog `AnalyticsEvent` code](https://github.com/ethyca/fideslog/blob/main/fideslog/sdk/python/event.py), you'll see that it's already enforcing event validity for events created using the `AnalyticsEvent` constructor. This class shouldn't be needed unless you have plans to add new features on top of fideslog. If that's the case, maybe it's worth discussing if those features should be included in the SDK itself?

PSalant726 (commented):
Notes from today's screen share so I don't forget 😅

requirements.txt
eastandwestwind: would be nice to have this automated via linter or some auto formatter. Thoughts?

src/fidesops/schemas/analytics.py
eastandwestwind: `FideslogAnalyticsEvent` makes it easier to pass around necessary event-level data within fidesops with less duplication of code. Since `event_created_at `, `docker`, and `local_host` properties don't require event-specific info, we don't need to pass it in to `analytics_client.send()`. So, these props can be calculated only when we construct the `AnalyticsEvent`

src/fidesops/analytics.py
eastandwestwind: I'd prefer to handle checking for `config.user.analytics_opt_out` and exceptions at the analytics class level, so that we don't duplicate code. We may also need to pass in user id in the future for retrieving analytics permissions which I'd also like to keep scoped to the analytics class.

eastandwestwind: My biggest unknown, though, is  `generate_ client_id`. I know you said this is supposed to be unique per client, but you mentioned in our screenshare that I should instead make it unique per user + instance. I guess I just need more clarity around this, if fidesops requires sending user-level event data.

eastandwestwind:
@PSalant726 this is ready for another pass!

Makefile
```diff
@@ -28,7 +28,7 @@ help:
 
 init-db: compose-build
 	@echo "Check for new migrations to run..."
-	@docker-compose run --rm $(IMAGE_NAME) \
+	@docker-compose run --rm -e ANALYTICS_OPT_OUT \ $(IMAGE_NAME) \
```
PSalant726, line 31: No `\` needed here, since you're not adding a newline.
```suggestion
	@docker-compose run --rm -e ANALYTICS_OPT_OUT $(IMAGE_NAME) \
```

run_infrastructure.py
```diff
@@ -36,6 +36,7 @@ def run_infrastructure(
     run_tests: bool = False,  # Should we run the tests after creating the infra?
     run_create_superuser: bool = False,  # Should we run the create_superuser command?
     run_create_test_data: bool = False,  # Should we run the create_test_data command?
+    analytics_opt_out: bool = False, # Should we opt out of analytics?
```
PSalant726, line 39: My formatter is adding another space here? 🤷‍♂️ 
```suggestion
    analytics_opt_out: bool = False,  # Should we opt out of analytics?
```

run_infrastructure.py
```diff
@@ -91,6 +92,7 @@ def run_infrastructure(
             datastores,
             docker_compose_path=path,
             pytest_path=pytest_path,
+            analytics_opt_out=analytics_opt_out
```
PSalant726, line 95: And a trailing `,` here... 🤷‍♂️ 
```suggestion
            analytics_opt_out=analytics_opt_out,
```

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+from fidesops.core.config import config, update_config_file
+
+from ._version import get_versions
+
+logger = logging.getLogger(__name__)
+
+
+def get_version() -> str:
+    """Version of Fidesops"""
+    return get_versions()["version"]
```
PSalant726, line 20: Rather than reimplement this, you can do:
```python
from fidesops import __version__ as fidesops_version
```

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+analytics_client = AnalyticsClient(
+    # env var (internal mode) supersedes config file
+    client_id=os.getenv("FIDESOPS__ROOT_USER__ANALYTICS_ID")
+    or config.root_user.ANALYTICS_ID
+    or generate_and_store_client_id(),
+    developer_mode=in_developer_mode(),
+    extra_data=None,
+    os=system(),
+    product_name="fidesops",
+    production_version=get_version(),
```
PSalant726, line 56: In conjunction with the above, this can just become:
```suggestion
    production_version=fidesops_version,
```

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def get_version() -> str:
+    """Version of Fidesops"""
+    return get_versions()["version"]
+
+
+def in_developer_mode() -> bool:
+    """True if running in dev mode"""
+    return config.dev_mode
```
PSalant726, line 25: Is there a need for this function if it always returns this constant (per instance) value?

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def in_docker_container() -> bool:
+    """`True` if the command was submitted within a Docker container. Default: `False`."""
+    return True
```
PSalant726, line 30: I wouldn't assume this will always be `True`. In fidesctl, we [create an ENV variable when running in docker](https://github.com/ethyca/fides/blob/a1e08d7b2565e01a82f70ad7625179620db54c74/Dockerfile#L90-L91). Then, we [read the variable at run time](https://github.com/ethyca/fides/blob/4e6b6591406a842ca5376109bf10b862d69f147c/src/fidesctl/cli/utils.py#L143) to make this determination. I'd recommend doing the same here.

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def running_on_local_host() -> bool:
+    """For events submitted as a result of making API server requests, `True` if the API server is running on the user's local host. Default: `False`."""
+    return False
```
PSalant726, line 35: Here we need a way to detect whether the server is running on `localhost` or equivalent (`0.0.0.0`, `127.0.0.1`, etc). Passing `False` for all events would skew the analytics.

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def generate_and_store_client_id() -> str:
+    update_obj: Dict[str, Dict] = {}
+    client_id: str = generate_client_id(FIDESOPS)  # get fideslog bytestring
```
PSalant726, line 40: This comment threw me. My interpretation was "get the bytestring representing fideslog", not "get the fidesops bytestring from the fideslog SDK". Since the latter is clearly what you meant, and since `FIDESOPS` is being imported from the fideslog SDK above, I think this comment should just be removed.
```suggestion
    client_id: str = generate_client_id(FIDESOPS)
```

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+    logger.info("analytics client id generated")
```
PSalant726, line 41: I'd be hesitant to bloat the logs with a line like this, given that it doesn't provide any specific information. If you'd prefer to keep the line, maybe downgrade it to the `DEBUG` level?

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+    update_obj.update(root_user={"ANALYTICS_ID": client_id})
+    update_config_file(update_obj)
+    return client_id
+
+
+analytics_client = AnalyticsClient(
+    # env var (internal mode) supersedes config file
+    client_id=os.getenv("FIDESOPS__ROOT_USER__ANALYTICS_ID")
+    or config.root_user.ANALYTICS_ID
+    or generate_and_store_client_id(),
```
PSalant726, line 51: It would be safer to offload this logic to the `config.py` file. That way, you guarantee that there is always a root user analytics ID prior to the server starting up.

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+    developer_mode=in_developer_mode(),
+    extra_data=None,
+    os=system(),
+    product_name="fidesops",
+    production_version=get_version(),
+)
+
+
+def send_event(event: AnalyticsEvent) -> None:
+    if config.root_user.ANALYTICS_OPT_OUT or os.environ.get("ANALYTICS_OPT_OUT"):
```
PSalant726, line 61: If you add:
```python
class Config:
    env_prefix = "FIDESOPS__ROOT_USER__"
```
to the `RootUserSettings` class in `config.py`, you'll be able to check for a single ENV variable everywhere, and it would override anything in a config file. Then, this check simplifies to:
```suggestion
    if config.root_user.ANALYTICS_OPT_OUT:
```

src/fidesops/core/config.py
```diff
@@ -191,6 +191,14 @@ class Config:
         env_prefix = "FIDESOPS__SECURITY__"
 
 
+class RootUserSettings(FidesSettings):
+    """Configuration settings for Analytics variables."""
+
+    ANALYTICS_OPT_OUT: Optional[bool]
+    # analytics id should be set to "internal" when in development / testing
```
PSalant726, line 198: This should only be documented internally. As an open-source repo, we risk having random non-Ethyca employee users set their `ANALYTICS_ID` to `"internal"`, and skew our analytics.

Unfortunately I don't know the best place to document this. Previously I just reminded people to set the variable in their local ENV's a few times via Slack 🤷‍♂️

src/fidesops/core/config.py
```diff
@@ -327,6 +336,33 @@ def get_censored_config(the_config: FidesopsConfig) -> Dict[str, Any]:
     return filtered
 
 
+def update_config_file(updates: Dict[str, Dict[str, Any]]) -> None:
```
PSalant726, line 339: Where have I seen this before? 🤔 😎

src/fidesops/main.py
```diff
@@ -47,6 +51,21 @@ def start_webserver() -> None:
     if config.database.ENABLED:
         logger.info("Starting scheduled request intake...")
         initiate_scheduled_request_intake()
+    import os
+
+    if os.getenv("FIDESOPS__ROOT_USER__ANALYTICS_ID"):
```
PSalant726, line 56: Oh look you're using `FIDESOPS__ROOT_USER__ANALYTICS_ID`! I think you do need to explicitly define the prefix in `config.py` though, no?

src/fidesops/main.py
```diff
@@ -47,6 +51,21 @@ def start_webserver() -> None:
...
+        logger.info(
+            f"Analytics ID override active for internal use. Will send events with client id = {os.getenv('FIDESOPS__ROOT_USER__ANALYTICS_ID')}"
```
PSalant726, line 58: Theoretically anyone could set this ENV variable, so it won't necessarily be "for internal use". Regardless, if you offload the `client_id` population to `config.py` as suggested above, you won't need this log line at all.

src/fidesops/schemas/analytics.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class EVENT(Enum):
```
PSalant726, line 4: Maintaining the event names like this is a fantastic idea!! This is definitely something that should be done in fidesctl too.

Just one suggestion: since all of the event names will likely be strings:
```suggestion
class EVENT(str, Enum):
```
This will enable some nice shortcuts if we ever need to manipulate or compare the names  outside of this `Enum` class.

src/fidesops/analytics.py
```diff
@@ -0,0 +1,70 @@
...
+    or generate_and_store_client_id(),
+    developer_mode=in_developer_mode(),
+    extra_data=None,
+    os=system(),
+    product_name="fidesops",
+    production_version=get_version(),
+)
+
+
+def send_event(event: AnalyticsEvent) -> None:
```
PSalant726, line 60: It's a nit, but if you rename this to `send_analytics_event`, it would enable some better readability when this function is [seemingly] randomly called in other files.
```suggestion
def send_analytics_event(event: AnalyticsEvent) -> None:
```

PSalant726 (changes requested):

src/fidesops/analytics.py
eastandwestwind: ahh good point

src/fidesops/analytics.py
eastandwestwind: Nope, can inline for now.

src/fidesops/analytics.py
eastandwestwind: good call, I missed this.

src/fidesops/analytics.py
eastandwestwind: will do, and it should only be generated once per instance, so worst case won't spam logs. DEBUG if fine with me

src/fidesops/analytics.py
eastandwestwind: good idea!

src/fidesops/core/config.py
eastandwestwind: Good point, removed!

src/fidesops/analytics.py
eastandwestwind: So technically we always run "locally" as in local to the docker instance. But perhaps in the future we'll have a hosted version. - https://github.com/ethyca/fidesops/blob/main/src/fidesops/main.py#L62

src/fidesops/analytics.py
eastandwestwind: nice, thanks!

Makefile
```diff
@@ -105,32 +107,39 @@ isort-ci:
 
 pylint: compose-build
 	@echo "Running pylint checks..."
-	@docker-compose run $(IMAGE_NAME) \
+	@docker-compose run \
+	  	-e ANALYTICS_OPT_OUT \
```
PSalant726, line 111: Nit: a few extra spaces here
```suggestion
	-e ANALYTICS_OPT_OUT \
```

data/config/fidesops.toml
```diff
@@ -1,34 +1,37 @@
...
\ No newline at end of file
+TASK_RETRY_COUNT = 0
+TASK_RETRY_DELAY = 1
+TASK_RETRY_BACKOFF = 1
+REQUIRE_MANUAL_REQUEST_APPROVAL = false
+MASKING_STRICT = true
+
+[root_user]
+ANALYTICS_OPT_OUT = false
+ANALYTICS_ID = "450d2654ecb9ed27ae2327b4509cbe3d"
```
PSalant726, line 37: I think this file is intended just for tests? So, two things:
1. Do you want to opt out of analytics for test runs? (I'd recommend it)
2. Whatever the purpose of this `ANALYTICS_ID`, you don't want to commit a "valid" one to VCS. This should either be removed completely, or updated to be `"internal"`.

```suggestion
[root_user]
ANALYTICS_OPT_OUT = true
ANALYTICS_ID = "internal"
```
Right now this is the only blocker 😄

fidesops.toml
```diff
@@ -1,34 +1,37 @@
...
 
 [security]
-APP_ENCRYPTION_KEY="OLMkv91j8DHiDAULnK5Lxx3kSCov30b3"
-CORS_ORIGINS=["http://localhost", "http://localhost:8080", "http://localhost:3000", "http://localhost:3001"]
-ENCODING="UTF-8"
-OAUTH_ROOT_CLIENT_ID="fidesopsadmin"
-OAUTH_ROOT_CLIENT_SECRET="fidesopsadminsecret"
-DRP_JWT_SECRET="secret"
+APP_ENCRYPTION_KEY = "OLMkv91j8DHiDAULnK5Lxx3kSCov30b3"
+CORS_ORIGINS = [ "http://localhost", "http://localhost:8080", "http://localhost:3000", "http://localhost:3001",]
```
PSalant726, line 22: This change feels arbitrary, and would just mess with the `git blame`. Please revert 😄

run_infrastructure.py
PSalant726: Looks like the `,` is still missing. I'm asking you to add it in because auto-formatting tools will do so the next time this file is touched, and it's likely that the rest of the changes would be unrelated at that point.

run_infrastructure.py
```diff
@@ -233,6 +236,9 @@ def _run_tests(
             for env_var in EXTERNAL_DATASTORE_CONFIG[datastore]:
                 environment_variables += f"-e {env_var} "
 
+    if analytics_opt_out:
+        environment_variables += f"-e ANALYTICS_OPT_OUT"
```
PSalant726, line 240: No need for an f-string here.
```suggestion
        environment_variables += "-e ANALYTICS_OPT_OUT"
```

src/fidesops/analytics.py
PSalant726: I see... So this isn't straightforward, but maybe it's best to always return `True` if we're always running on `0.0.0.0` ? It's your call.

PSalant726 (changes requested):
Only one of these comments is blocking this right now - the rest are just nits. This is really close 🙏

fidesops.toml
eastandwestwind: These spaces will be added anyway, when the fidesops.toml gets overridden when `client_id` is generated. Probably something with `toml.dump`- https://github.com/ethyca/fidesops/pull/541/files#diff-9aca0346a2a86990290c4e236109ceab5ff39557687d38c0aefd74da1659f3edR416. Should I still update?

eastandwestwind:
Thanks again for the review @PSalant726 ! Back over to you, with some need for clarification

data/config/fidesops.toml
```diff
@@ -1,34 +1,37 @@
...
\ No newline at end of file
+TASK_RETRY_COUNT = 0
+TASK_RETRY_DELAY = 1
+TASK_RETRY_BACKOFF = 1
+REQUIRE_MANUAL_REQUEST_APPROVAL = false
+MASKING_STRICT = true
+
+[root_user]
+ANALYTICS_OPT_OUT = true
+ANALYTICS_ID = "internal"
```
PSalant726, line 37: This value shouldn't be needed at all if `ANALYTICS_OPT_OUT = true`, since no analytics should ever be sent. If it's being added automatically by `config.py`, that's worth addressing.

PSalant726 (approved):
All of my comments at this point are just nice-to-haves, nothing blocking. Thanks for sticking with this - it's looking great now 🙌

Makefile
eastandwestwind: Confused.. still not seeing the extra spaces
Screen Shot 2022-06-07 at 2 15 57 PM

eastandwestwind: confused... still not seeing the extra spaces on my end:
Screen Shot 2022-06-07 at 2 17 28 PM

data/config/fidesops.toml
eastandwestwind: yeah it shouldn't be needed

fidesops.toml
```diff
@@ -1,34 +1,37 @@
...
-REQUIRE_MANUAL_REQUEST_APPROVAL=false
-MASKING_STRICT=true
+TASK_RETRY_COUNT = 0
+TASK_RETRY_DELAY = 1
+TASK_RETRY_BACKOFF = 1
+REQUIRE_MANUAL_REQUEST_APPROVAL = false
+MASKING_STRICT = true
+
+[root_user]
+ANALYTICS_OPT_OUT = false
```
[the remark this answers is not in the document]
eastandwestwind: Checked in with Adria and Kelly- confirmed we're OK defaulting to `false`

eastandwestwind:
Ok we're ready again @PSalant726 !

PSalant726 (approved):

Issue #545, opened 2022-05-20
environment var to disable fidesops db integration for certain use cases


adamsachs:
### Is your feature request related to a specific problem?

Certain targeted uses of Fidesops (e.g. [using fidesops as a masking service](https://ethyca.github.io/fidesops/guides/masking_strategies/#using-fidesops-as-a-masking-service)) make use of APIs that do not require the core application db integration. In use cases that want to leverage only these APIs, it may be desirable to allow admins to deploy the application in a mode that does not require db integration, enabling lighter-weight deployments.

### Describe the solution you'd like

An environment variable (e.g. `FIDESOPS__DATABASE__ENABLED` ?), defaulted to `True`, that when set to `False` bypasses the `init_db()` call in `start_webserver()` in our `main.py`. If the flag is set to `False`, any endpoints that require application db integration (most) will fail; preferably they will fail with a clear error message reported to the user. Any endpoints that do **not** require application db integration (e.g. the standalone masking endpoint) will continue to operate normally.

Some initial investigation points me to [get_db()](https://github.com/ethyca/fidesops/blob/main/src/fidesops/api/deps.py#L7) as potentially a good place to check for the config setting and throw a clearer error message to the user, rather than letting the app attempt to connect to the db and fail ungracefully.

PR #550, opened 2022-05-23
environment variables to disable app db and cache. some related cleanup

adamsachs:
# Purpose
Provide config options to allow fidesops to start cleanly without db and cache integration. Used for narrow use cases like standalone masking engine.

# Changes
- Added `FIDESOPS__DATABASE__ENABLED` and `FIDESOPS__REDIS__ENABLED` configuration variables to allow `fidesops` to run cleanly in a "stateless" mode without any database or redis cache integration
- added exception handlers and associated support to provide clean error handling when endpoints that require db or cache integration are invoked when they're configured to be disabled
- minor update in `main.py`'s uvicorn invocation to consistently start app even if `init_db` isn't called
- documentation updates for above config options; cleanup some typos in existing documentation ( @ethyca/docs-authors  review please!)

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file in appropriate sections

# Ticket

Fixes [#545](https://github.com/ethyca/fidesops/issues/545)


Makefile
```diff
@@ -43,6 +43,9 @@ reset-db:
 server: compose-build
 	@docker-compose up
 
+server-no-db: compose-build
```
adamsachs, line 46: useful for local testing

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -30,16 +30,18 @@ The `fidesops.toml` file should specify the following variables:
 | TOML Variable | ENV Variable | Type | Example | Default | Description |
 |---|---|---|---|---|---|
 | `SERVER` | `FIDESOPS__DATABASE__SERVER` | string | postgres.internal | N/A | The networking address for the Fideops Postgres database server |
-| `USER` | `FIDESOPS__DATABASE_USER` | string | postgres | N/A | The database user with which to login to the Fidesops application database |
-| `PASSWORD` | `FIDESOPS__DATABASE_PASSWORD` | string | apassword | N/A | The password with which to login to the Fidesops application database |
+| `USER` | `FIDESOPS__DATABASE__USER` | string | postgres | N/A | The database user with which to login to the Fidesops application database |
```
adamsachs, line 33: these were doc typos as far as i could tell -- missing the double underscore

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -84,8 +88,8 @@ OAUTH_ROOT_CLIENT_SECRET="fidesopsadminsecret"
 TASK_RETRY_COUNT=3
 TASK_RETRY_DELAY=20
 TASK_RETRY_BACKOFF=2
-REQUIRE_MANUAL_REQUEST_APPROVAL=True
-MASKING_STRICT=True
+REQUIRE_MANUAL_REQUEST_APPROVAL=true
+MASKING_STRICT=true
```
adamsachs, line 92: updated doc reference because camel-cased booleans in my `.toml` weren't working -- pretty sure lowercase is needed

src/fidesops/api/v1/exception_handlers.py
```diff
@@ -0,0 +1,20 @@
...
+class ExceptionHandlers:
+    def functionality_not_configured_handler(
+        request: Request, exc: FunctionalityNotConfigured
+    ) -> JSONResponse:
+        return JSONResponse(
+            status_code=HTTP_500_INTERNAL_SERVER_ERROR, content={"message": str(exc)}
+        )
+
+    @classmethod
+    def get_handlers(cls) -> List[Callable[[Request, Exception], Response]]:
```
adamsachs, line 19: wanted to provide a starting point if more exception handlers would be needed over time...perhaps there's a more elegant/dynamic way of doing this, having to list out the methods seems a bit cumbersome

conceptualshark (commented):
👍  docs changes look great to me, thanks for catching them!

src/fidesops/main.py
```diff
@@ -29,21 +31,29 @@
...
-    initiate_scheduled_request_intake()
+    if config.database.ENABLED:
+        # don't schedule request intake if database is disabled
+        logger.info("Starting scheduled request intake...")
+        initiate_scheduled_request_intake()
 
     logger.info("Starting web server...")
+
     uvicorn.run(
-        "src.fidesops.main:app",
```
adamsachs, line 46: when `init_db` wasn't run, uvicorn wasn't finding the `fidesops` module anymore when this was the run command. i think i was able to pin it down to the fact that within `init_db()`, we run `src/migrations/env.py`, which happens to add `..` to the `sys.path` -- and only because of that the `src.fidesops` path is picked up correctly. to me this seems like an unrecognized and unintentional side-effect -- `init_db()` should not be needed to get the base app running, i don't think -- if it is intended, then we need to do it in a much less obfuscated way.

as far as i could tell, making this switch here doesn't impact existing behavior, and it also allows the app to run if `init_db` is not called. i don't know of any negative consequences (though i could certainly be missing something!)

docker-compose.no-db.yml
```diff
@@ -0,0 +1,38 @@
...
+        source: ./
+        target: /fidesops
+        read_only: False
+      - /fidesops/src/fidesops.egg-info
+    environment:
+      - FIDESOPS__DEV_MODE=${FIDESOPS__DEV_MODE}
+      - FIDESOPS__LOG_PII=${FIDESOPS__LOG_PII}
+      - FIDESOPS__HOT_RELOAD=${FIDESOPS__HOT_RELOAD}
+      - FIDESOPS__DATABASE__ENABLED=false
+      - FIDESOPS__REDIS__ENABLED=false
```
NevilleS, line 28: Commenting here with some `docker compose` trivia... you can avoid this separate `no-db.yml` by exposing this ENV variable on the main file like so:
```
- FIDESOPS__DATABASE__ENABLED =${FIDESOPS__DATABASE__ENABLED:-true}
```
That will default to true, or take the value of the ENV var provided to the docker compose invocation.

Then you just need to set the value of the ENV variable in the Makefile to adjust that behaviour, like:
```
server-no-db: export FIDESOPS__DATABASE__ENABLED=false
server-no-db: ... <continue existing target here>
```

I do this exact pattern in fidesdemo, see:
* https://github.com/ethyca/fidesdemo/blob/main/docker-compose.yml#L68
* https://github.com/ethyca/fidesdemo/blob/main/Makefile#L189

adamsachs: certainly more concise for the ENV var piece of it, but the `no-db.yml` compose also doesn't bootstrap the `db` and `redis` (and `docs`) services, to more closely simulate the "standalone" deployment. as far as i know (and based on our conversation on friday), i don't think there's a way to have branching logic in `docker-compose` based on env var? but i can dig on that if you think it's possible.

NevilleS: Ah yes, that's true... it's Hard™️ to make docker compose files dynamic and support different run configurations.

NevilleS: TIL this exists: https://docs.docker.com/compose/profiles/

adamsachs: oh - that looks pretty nice. i can take a shot at that, seems a lot cleaner. thanks for pointing me to that!

adamsachs: hmm. because we've got dependencies between `fidesops` and `db` and `redis`, i don't think this is going to work for us, without some significant refactor. as far as i can tell, there's no way to express conditional dependencies, and i feel like the no-db use case is enough of an edge case for now that i am reluctant to remove those dependencies...

src/fidesops/api/deps.py
```diff
@@ -1,11 +1,17 @@
...
+from fidesops.common_exceptions import FunctionalityNotConfigured
+from fidesops.core.config import config
 from fidesops.db.session import get_db_session
 from fidesops.util.cache import get_cache as get_redis_connection
 
 
 def get_db() -> Generator:
     """Return our database session"""
+    if not config.database.ENABLED:
+        raise FunctionalityNotConfigured(
```
sanders41, line 12: Let's add a tests for this and the `get_cache`. I'm thinking add a new file at `tests/api/test_deps.py` unless you think there is a better place to put it.

```py
import pytest

from fidesops.api.deps import get_cache, get_db
from fidesops.common_exceptions import FunctionalityNotConfigured
from fidesops.core import config


@pytest.fixture
def mock_config():
    db_enabled = config.config.database.ENABLED
    redis_enabled = config.config.redis.ENABLED
    config.config.database.ENABLED = False
    config.config.redis.ENABLED = False
    yield
    config.config.database.ENABLED = db_enabled
    config.config.redis.ENABLED = redis_enabled


@pytest.mark.usefixtures("mock_config")
def test_get_cache_not_enabled():
    with pytest.raises(FunctionalityNotConfigured):
        next(get_cache())


@pytest.mark.usefixtures("mock_config")
def test_get_db_not_enabled():
    with pytest.raises(FunctionalityNotConfigured):
        next(get_db())
```

sanders41 (changes requested):

src/fidesops/main.py
sanders41: You explanation here makes sense to me and everything seems to work fine with your change.

src/fidesops/api/deps.py
adamsachs: @sanders41 thank you for writing the test up! i ran a bit further with this and also added in a test on the endpoint layer to make sure the exception handlers work as expected - let me know what you think.

src/fidesops/api/v1/exception_handlers.py
```diff
@@ -0,0 +1,21 @@
...
+from fidesops.common_exceptions import FunctionalityNotConfigured
+
+
+class ExceptionHandlers:
+    @staticmethod
+    def functionality_not_configured_handler(
+        request: Request, exc: FunctionalityNotConfigured
+    ) -> JSONResponse:
+        return JSONResponse(
+            status_code=HTTP_500_INTERNAL_SERVER_ERROR, content={"message": str(exc)}
```
adamsachs, line 16: @sanders41  while you're here - any thoughts on `500` as the error code here? i considered other options but anything else i could think of seemed like it could be misinterpreted, whereas i feel like a generic `500` with a clear error message probably has the best chance of being well understood. but i'd like a second opinion on this!

sanders41: `500` make sense to me here. The other option I could see making sense is `503`, but `503` I think would imply things are down temporarily and will come back up which won't be the case so that could be confusing.

PR #554, opened 2022-05-23
Pause Access Request and Resume with Manual Input [#521]

### LINKED ISSUE #521: Support for pausing an execution (Access Requests)
###Description
Our current means of executing DAGs need to be enhanced to support scalability, reliability, and additional use cases when it comes to Subject Request processing (like pausing a request, cancelling a request, etc). The scope of this ticket is focused on adding a node that can pause an ACCESS request, take in manual input for example, and resume the request with control from the user.

We may need to split out the resume endpoint into another ticket or PR, but leaving it in scope for now.

pattisdr:
❗ Contains a DB migration 

# Purpose

Sometimes there is no automated way to retrieve data.  A service may have no external API or user data may be buried in a physical filing cabinet.  During an access request, allow the graph to pause and wait for the user to manually add the data to resume execution.  (If there's no manual data to add, resume with an empty list).

# Changes

Big picture, a new connector type: `ManualConnector` has been added.  When we go to retrieve data from a manual source, rather than making an API request or a database query as we do with other connector types, we look to see if there's any data in the cache for that collection.  If no data has been added, we pause the PrivacyRequest.  **This stops the execution of all other collections in the graph as well.**

The user should add the manual rows for the paused collection via a POST to `{{host}}/privacy-request/{{privacy_request_id}}/manual_input`.  This will add the manual data to the cache for the paused collection and resume executing the privacy request.    Resuming a paused graph actually builds a modified graph behind the scenes.  A new graph will be built that doesn't run queries for collections we've already visited. Instead, those collections are marked as not having any upstream dependencies, and we just return the data we've already retrieved. When we reach the paused collection on resume, the node will find the manually added data in the cache, and proceed, passing that data onward to downstream collections.

A big change to the core, is that we now **execute the tasks for a graph sequentially**.   Pausing throws an exception and stops execution of all other tasks.  Running sequentially helps make sure the session is closed at the right time, and in-flight tasks aren't trying to use that session.  It also ensures the creation of ExecutionLogs (which are written for each collection at different points in the workflow) are not interrupted. A side effect of this is that one node is paused at a time, but I'd imagine that most "manual" collections could be defined as a single collection.  

## Detailed Changes

- Adds a manual connector type.  You can now define manual datasets that describe what data the manual data depends on, and what manual fields the user is expected to add.
  - Adds a demonstration manual dataset with two manual collections
- Adds a paused state for ExecutionLogs
- Adds methods to the `PrivacyRequest` to cache the paused collection and what step we're on (access or erasure). This is needed because an erasure runs both an access graph and an erasure graph. The paused collection tells us where to store the manual data the user gives us and can also be something we surface to them.
- Adds logic when resuming a privacy request to skip straight to building the access request instead of re-running the Pre-webhooks.
- Separates out `upload_access_results` from PrivacyRequestRunner.run because the method is getting too long.
- Adds `update_mapping_from_cache` which uses data from the cache instead of revisiting a node if it exists when building the mapping we pass to Dask for execution.
- Updates the postman collection with examples that can run an access request with a manual node and resume execution.
- Adds an API endpoint for resuming a paused privacy request with manually input data. .  Fields should match the fields declared in the dataset yml. If no data is found, submit an empty list.

POST `{{host}}/privacy-request/{{privacy_request_id}}/manual_input`
```json
[{
    "id": 2,
    "authorized_user": "Jiminy Customer",
    "customer_id": 1,
    "payment_card_id": "pay_bbb-bbb"
}]
```

# The New Graph on Resume

In this example, we have a graph of postgres collections, as well as a collection requiring manual input:  `manual_example:filing_cabinet`, that needs information from the customer table to be able to locate the manual data.  When we run the access request, we query the `postgres_example:customer` and `postgres_example:orders` collections, before it hits the `manual_example:filing_cabinet` node and pauses. 

before

After the user has retrieved and entered the data from the filing cabinet, a new graph is formed. Our previous run-through got data from `postgres_example:customer` and `postgres_example:orders`, so we start with those, we don't need to re-query those tables.  We also don't need anything from the the root node because its dependent collections were already visited. When we reach the `manual_example:filing_cabinet` node this time, we pull the manually added data from the cache, and pass that to the `postgres_example:payment_card` node to use in further querying.

after

# Worth noting

- Make sure the data is cached for an appropriate amount of time in Redis.  Redis automatically expires the data in your cache per your configured settings.   If your expiration time is shorter than than the amount of time you need to manually gather data to fulfill requests, you should lengthen this time.
- This implementation assumes that the user hasn't updated their collections since the privacy request has paused, which won't always hold.  An alternative is caching the dictionary we feed to dask with each node and its dependencies. and commit to fulfilling that exact graph, rather than building a new one.    There are pros and cons to both approaches, and reasons why we wouldn't be able to fulfill the new graph or the old graph if the user made collection edits.
- This doesn't yet expose to the user which node on which node in the graph the privacy request is paused, but it's something we could add to the privacy request status endpoint, or whever else it might be useful.

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file in appropriate sections

# Ticket

Fixes #521


tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
                 "name": privacy_request.policy.name,
             },
         }
+
+
+class TestResumeWithManualInput:
+    @pytest.fixture(scope="function")
+    def url(self, db, privacy_request):
```
sanders41, line 1620: The `db` parameter is never used, can it be removed?

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+            ActionType.access, CollectionAddress("manual_input", "filing_cabinet")
+        )
+        response = api_client.post(url, headers=auth_header, json=[{"mock": "row"}])
+        assert response.status_code == 422
+        assert (
+            response.json()["detail"]
+            == "Cannot save manual rows. No 'mock' field defined on the 'manual_input:filing_cabinet' collection."
+        )
+
+    def test_resume_with_manual_input(
```
sanders41, line 1703:
```diff
- def test_resume_with_manual_input(
-    self,
-    db,
-    api_client,
-    url,
-    generate_auth_header,
-    privacy_request,
-    postgres_example_test_dataset_config,
-    manual_dataset_config,
- )
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config", "manual_dataset_config")
+ def test_resume_with_manual_input(
+    self,
+    db,
+    api_client,
+    url,
+    generate_auth_header,
+    privacy_request,
+ )

File: tests/api/v1/endpoints/test_privacy_request_endpoints.py [hunk 3]
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+            ActionType.access, CollectionAddress("manual_example", "filing_cabinet")
+        )
+        response = api_client.post(url, headers=auth_header, json=[{"mock": "row"}])
+        assert response.status_code == 422
+        assert (
+            response.json()["detail"]
+            == "Cannot save manual rows. No collection in graph with name: 'manual_example:filing_cabinet'."
+        )
+
+    def test_resume_with_manual_input_invalid_data(
```
sanders41, line 1678: Since `postgres_example_test_dataset_config` and `manual_dataset_config` are fixtures that aren't directly accessed:

```diff
- def test_resume_with_manual_input_invalid_data(
-    self,
-    db,
-    api_client,
-    url,
-    generate_auth_header,
-    privacy_request,
-    postgres_example_test_dataset_config,
-    manual_dataset_config,
- )
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config", "manual_dataset_config")
+ def test_resume_with_manual_input_invalid_data(
+    self,
+    db,
+    api_client,
+    url,
+    generate_auth_header,
+    privacy_request,
+ )

File: tests/api/v1/endpoints/test_privacy_request_endpoints.py [hunk 4]
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+        assert response.status_code == 200
+
+        db.refresh(privacy_request)
+        assert privacy_request.status == PrivacyRequestStatus.in_processing
+
+
+class TestValidateManualInput:
+    """Verify pytest cell-var-from-loop warning is a false positive"""
+
+    def test_all_fields_match(self, db, postgres_example_test_dataset_config):
```
sanders41, line 1741:
```diff
- def test_all_fields_match(self, db, postgres_example_test_dataset_config):
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config")
+ def test_all_fields_match(self, db):
```

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+        paused_location = CollectionAddress("postgres_example_test_dataset", "address")
+
+        manual_rows = [{"city": "Nashville", "state": "TN"}]
+        validate_manual_input(manual_rows, paused_location, db)
+
+    def test_one_field_does_not_match(self, db, postgres_example_test_dataset_config):
```
sanders41, line 1747:
```diff
- def test_one_field_does_not_match(self, db, postgres_example_test_dataset_config):
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config")
+ def test_one_field_does_not_match(self, db):
```

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+
+        manual_rows = [{"city": "Nashville", "state": "TN", "ccn": "aaa-aaa"}]
+        with pytest.raises(HTTPException) as exc:
+            validate_manual_input(manual_rows, paused_location, db)
+        assert (
+            exc.value.detail
+            == "Cannot save manual rows. No 'ccn' field defined on the 'postgres_example_test_dataset:address' collection."
+        )
+
+    def test_field_on_second_row_does_not_match(
```
sanders41, line 1758:
```diff
- def test_field_on_second_row_does_not_match(self, db, postgres_example_test_dataset_config):
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config")
+ def test_field_on_second_row_does_not_match(self, db):
```

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,173 @@ def test_resume_privacy_request(
...
+            {"city": "Austin", "misspelled_state": "TX"},
+        ]
+        with pytest.raises(HTTPException) as exc:
+            validate_manual_input(manual_rows, paused_location, db)
+        assert (
+            exc.value.detail
+            == "Cannot save manual rows. No 'misspelled_state' field defined on the 'postgres_example_test_dataset:address' collection."
+        )
+
+    def test_collection_does_not_exist(self, db, postgres_example_test_dataset_config):
```
sanders41, line 1774:
```diff
- def test_collection_does_not_exist(self, db, postgres_example_test_dataset_config):
+ @pytest.mark.usefixtures("postgres_example_test_dataset_config")
+ def test_collection_does_not_exist(self, db):
```

tests/integration_tests/test_manual_task.py
```diff
@@ -0,0 +1,272 @@
...
+)
+from fidesops.task import graph_task
+
+from ..graph.graph_test_util import assert_rows_match
+from ..task.traversal_data import postgres_and_manual_nodes
+
+
+@pytest.mark.integration_postgres
+@pytest.mark.integration
+def test_postgres_with_manual_input_access_request_task(
```
sanders41, line 22:
```diff
- def test_postgres_with_manual_input_access_request_task(
-    db,
-    policy,
-    integration_postgres_config,
-    integration_manual_config,
-    postgres_integration_db,
-    cache,
- ) -> None:
+ @pytest.mark.usefixtures("postgres_integration_db", "cache")
+ def test_postgres_with_manual_input_access_request_task(
+    db,
+    policy,
+    integration_postgres_config,
+    integration_manual_config,
+) -> None:

**Review by conceptualshark — COMMENTED** [review]
Will make a few quick changes in a separate commit - just one question, and this summary looks great!

File: tests/integration_tests/test_manual_task.py [hunk 9]
```diff
@@ -0,0 +1,272 @@
...
+    )
+    assert [log.status for log in address_logs] == [
+        ExecutionLogStatus.in_processing,
+        ExecutionLogStatus.complete,
+    ]
+
+
+@pytest.mark.integration_postgres
+@pytest.mark.integration
+def test_no_manual_input_found(
```
sanders41, line 192: The `db` fixture doesn't look to be used and can be removed here?

```diff
- def test_no_manual_input_found(
-    db,
-    policy,
-    integration_postgres_config,
-    integration_manual_config,
-    postgres_integration_db,
-    cache,
- ) -> None:
+ @pytest.mark.usefixtures("postgres_integration_db", "cache")
+ def test_no_manual_input_found(
+    policy,
+    integration_postgres_config,
+    integration_manual_config,
+) -> None:

File: src/fidesops/api/v1/endpoints/privacy_request_endpoints.py [hunk 10]
```diff
@@ -569,6 +576,93 @@ def resume_privacy_request(
...
+) -> None:
+    """Validate manually-added data for a collection.
+
+    The specified collection must exist and all fields must be previously defined.
+    """
+    datasets = DatasetConfig.all(db=db)
+    dataset_graphs = [dataset_config.get_graph() for dataset_config in datasets]
+    dataset_graph = DatasetGraph(*dataset_graphs)
+
+    node: Node = dataset_graph.nodes.get(collection)
```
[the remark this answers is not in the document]
pattisdr: yep good catch

src/fidesops/models/privacy_request.py
```diff
@@ -226,6 +228,77 @@ def get_results(self) -> Dict[str, Any]:
...
+        portion of the privacy request flow, and the collection tells us where we should cache manual input data for later use,
+        In other words, this manual data belongs to this collection.
+        """
+        cache: FidesopsRedis = get_cache()
+        cached_paused_node = cache.get_encoded_objects_by_prefix(
+            f"PAUSED_LOCATION__{self.id}"
+        )
+        node_addr = (
+            list(cached_paused_node.values())[0]
+            if len(cached_paused_node.keys()) == 1
```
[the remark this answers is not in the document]
pattisdr: Yeah this is confusing the way I have it.  To answer your question, I only ever expect one key at most here, and if one exists, I wanted to get its matching value without knowing its entire key.

Possible `cached_paused_nodes`:
- Privacy request is paused: `{'EN_PAUSED_LOCATION__test_postgres_access_request_task_dc0b5372-17a0-4852-980d-f8c35824a4d4': 'access__manual_example:storage_unit'}`
- Privacy request has been unpaused: `{'EN_PAUSED_LOCATION__test_postgres_access_request_task_dc0b5372-17a0-4852-980d-f8c35824a4d4': None}`
- Privacy request has never been paused: `{}`


Looks like there's an existing method I could use that would be less confusing here:  `cache.get_encoded_by_key(f"EN_PAUSED_LOCATION__{self.id}")`

src/fidesops/models/privacy_request.py
```diff
@@ -226,6 +228,77 @@ def get_results(self) -> Dict[str, Any]:
...
+            else None
+        )
+
+        if node_addr:
+            split_addr = node_addr.split("__")
+            return ActionType(split_addr[0]), CollectionAddress.from_string(
+                split_addr[1]
```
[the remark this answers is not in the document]
pattisdr: Shouldn't be, `get_paused_step_and_collection` is retrieving the data from the method above: `cache_paused_step_and_collection`, which stores two values separated by a "__".  Maybe I should make the separator even more unique in case the CollectionAddress happened to have two underscores. I'll dry up the separator and share between the two methods too.

src/fidesops/models/privacy_request.py
```diff
@@ -226,6 +228,77 @@ def get_results(self) -> Dict[str, Any]:
...
+
+    def get_manual_input(
+        self, collection: CollectionAddress
+    ) -> Dict[str, Optional[List[Row]]]:
+        """Retrieve manually added rows from the cache for the given CollectionAddress.
+        Returns the manual key mapped to the manual data.
+        """
+        cache: FidesopsRedis = get_cache()
+        prefix = f"MANUAL_INPUT__{self.id}__{collection.value}"
+        value_dict = cache.get_encoded_objects_by_prefix(prefix)
```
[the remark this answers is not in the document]
pattisdr: good cleanup, leftover from earlier version

conceptualshark:
Made some quick updates to the docs. Feel free to edit/clarify if I misinterpreted anything. Thank you for the easy to understand overview!

src/fidesops/service/connectors/manual_connector.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        node: TraversalNode,
+        policy: Policy,
+        privacy_request: PrivacyRequest,
+        input_data: Dict[str, List[Any]],
+    ) -> Optional[List[Row]]:
+        """
+        Returns manually added data for the given collection if it exists, otherwise pauses the Privacy Request.
+        """
+        cached_results: Dict[
```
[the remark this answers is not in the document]
pattisdr: Thanks Paul, I wasn't being careful with my types, I think I want `Optional[Dict[str, Optional[List[Row]]]]` here.

src/fidesops/service/connectors/manual_connector.py
```diff
@@ -0,0 +1,62 @@
...
+            Optional[str], Optional[List[Row]]
+        ] = privacy_request.get_manual_input(node.address)
+
+        if cached_results:
+            privacy_request.cache_paused_step_and_collection()  # Caches paused location as None
+            return list(cached_results.values())[0]
```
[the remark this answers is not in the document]
pattisdr: Well the current design only allows one node to be paused at any given time now that we're running the nodes sequentially.  

Separately, I think I still want the [0] because I'm trying to get the corresponding value out of the list of dictionary values, I want to remove the outer wrapper:

```python
cached_results
{'EN_MANUAL_INPUT__test_postgres_access_request_task_586cddf4-2694-4437-8192-fdadbf87430f__manual_example:storage_unit': [{'box_id': 5, 'email': 'customer-1@example.com'}]}

cached_results.values()
dict_values([[{'box_id': 5, 'email': 'customer-1@example.com'}]])

list(cached_results.values())[0]
[{'box_id': 5, 'email': 'customer-1@example.com'}]
```

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -225,10 +210,67 @@ def run(
...
+                logging.error(
+                    f"Error uploading subject access data for rule {rule.key} on policy {policy.key} and privacy request {privacy_request.id} : {exc}"
+                )
+                privacy_request.status = PrivacyRequestStatus.error
+
     def dry_run(self, privacy_request: PrivacyRequest) -> None:
         """Pretend to dispatch privacy_request into the execution layer, return the query plan"""
 
 
+def _log_exception(exc: BaseException) -> None:
```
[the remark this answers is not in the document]
pattisdr: I can do that, I might have to make dev mode an argument for circular import reasons.

src/fidesops/task/graph_task.py
```diff
@@ -72,6 +73,13 @@ def result(*args: Any, **kwargs: Any) -> List[Optional[Row]]:
                         self.log_start(action_type)
                     # Run access or erasure request
                     return func(*args, **kwargs)
+                except PrivacyRequestPaused as ex:
+                    logger.warning(
+                        f"Privacy request {method_name} paused {self.traversal_node.address}"
+                    )
+                    self.log_paused(action_type, ex)
+                    # Re-raise to stop privacy request execution.
+                    raise PrivacyRequestPaused(ex)
```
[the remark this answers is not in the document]
pattisdr: I didn't know that, thank you!

src/fidesops/task/graph_task.py
```diff
@@ -486,6 +500,26 @@ def collect_queries_fn(
...
+    start_fn: Callable,
+) -> None:
+    """When resuming a privacy request from a paused or failed state, update the `dsk` dictionary with results we've
+    already obtained from a previous run. Remove upstream dependencies for these nodes, and just return the data we've
+    already retrieved, rather than visiting them again.
+
+    If there's no cached data, the dsk dictionary won't change.
+    """
+
+    cached_results: Dict[str, List[Row]] = resources.get_all_cached_objects()
```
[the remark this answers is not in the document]
pattisdr: hm, I think the actual type of `get_all_cached_objects` is Dict[str, Optional[List[Row]]], I'll change in those two places.

tests/api/v1/endpoints/test_privacy_request_endpoints.py
pattisdr: Thank you, I wasn't aware of `@pytest.mark.usefixtures`

docs/fidesops/docs/guides/manual_data.md
```diff
@@ -7,17 +7,16 @@ In this section we'll cover:
...
 
 ## Describing a manual dataset
 
-In this example, we have a manual dataset with one `storage_unit` collection.  `email` is 
-an identity, so this is the field needed to locate the  `box_id` in the storage unit.
+In the following example, the Manual Dataset contains one `storage_unit` collection.  `email` is 
+defined as the unit's [identity](datasets.md#field-members), which will then be used to retrieve the `box_id` in the storage unit.
 
-You would need to add both a Manual [ConnectionConfig](database_connectors.md) and this Manual Dataset to that ConnectionConfig.
+To add a Manual Dataset, first create a [Manual ConnectionConfig](database_connectors.md#example-6-manual-connectionconfig). The following Manual Dataset can then be added to the new ConnectionConfig:
```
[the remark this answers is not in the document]
pattisdr: Hm, I didn't go into detail here because this part isn't new behavior: every ConnectionConfig can have multiple DatasetConfigs attached to them, I've just added a new type of ConnectionConfig.

Looking further, I thought this was covered in the guides, but I don't see it.  Is this detail you could add @conceptualshark?  I would think it would be an extension to `datasets.md`.

Generally you create a ConnectionConfig which describes how to connect to postgres, mongodb, etc.  and then you create a DatasetConfig which describes the data contained there.  This is not new behavior added here, this is something you'd do for every connector type.  We describe what should be in the datasets here: https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/guides/datasets.md, but I don't see further explanation of how to actually add that to fidesops with an API request.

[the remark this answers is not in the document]
conceptualshark: Thanks for the clarification! I checked through the datasets doc and wasn't sure if I was missing something - a whole step makes sense. I can take a look and try to write something up for that piece. 👍

src/fidesops/task/graph_task.py
```diff
@@ -526,11 +561,14 @@ def termination_fn(*dependent_values: List[Row]) -> Dict[str, List[Row]]:
         env: Dict[CollectionAddress, Any] = {}
         end_nodes = traversal.traverse(env, collect_tasks_fn)
 
-        dsk = {k: (t.access_request, *t.input_keys) for k, t in env.items()}
-        dsk[ROOT_COLLECTION_ADDRESS] = (start_function(traversal.seed_data),)
+        dsk: Dict[CollectionAddress, Tuple[Any, ...]] = {
+            k: (t.access_request, *t.input_keys) for k, t in env.items()
```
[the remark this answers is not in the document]
sanders41: Ah, got it. I believe Python 3.11 is adding a way to do this, but I don't think it is possible now with needing the variable length. If you are comfortable with the the values you are going to get I think what you have is fine.

docs/fidesops/docs/guides/manual_data.md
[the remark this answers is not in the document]
pattisdr: cool thank you! it's a PATCH to `/api/v1/connection/connection_key/dataset`.  Until recently you had to convert the dataset yaml file into JSON to send in the request body, but I think with @nschoenbrot's work you can now send in the yaml directly

src/fidesops/task/graph_task.py
```diff
@@ -486,6 +500,26 @@ def collect_queries_fn(
...
+    If there's no cached data, the dsk dictionary won't change.
+    """
+
+    cached_results: Dict[str, Optional[List[Row]]] = resources.get_all_cached_objects()
+
+    for collection_name in cached_results:
+        dsk[CollectionAddress.from_string(collection_name)] = (
+            start_fn(cached_results[collection_name]),
+        )
+
```
pattisdr, line 521: This is a primary change.  This will update the graph when we attempt to re-run - it will still have the same number of collections as before, we just may not need to run actual queries on certain nodes, we'll use the data we already have.

src/fidesops/task/graph_task.py
```diff
@@ -526,11 +563,14 @@ def termination_fn(*dependent_values: List[Row]) -> Dict[str, List[Row]]:
...
-        dsk[ROOT_COLLECTION_ADDRESS] = (start_function(traversal.seed_data),)
+        dsk: Dict[CollectionAddress, Tuple[Any, ...]] = {
+            k: (t.access_request, *t.input_keys) for k, t in env.items()
+        }
+        dsk[ROOT_COLLECTION_ADDRESS] = (start_function([traversal.seed_data]),)
         dsk[TERMINATOR_ADDRESS] = (termination_fn, *end_nodes)
-        v = dask.delayed(get(dsk, TERMINATOR_ADDRESS))
+        update_mapping_from_cache(dsk, resources, start_function)
 
+        v = dask.delayed(get(dsk, TERMINATOR_ADDRESS, num_workers=1))
```
pattisdr, line 573: This is another primary change.  Using one worker on the threaded scheduler will force our tasks to run sequentially for a given privacy request.  

For some background, Pausing a privacy request throws an exception to stop execution. Dask Delayed assumes that every other task should stop.   Running sequentially here helps make sure a closed session doesn't affect other tasks that are in flight, as well as makes sure all execution logs get written in full.  With this design, only one collection can be paused at one time for a privacy request.  There's pros and cons with this, but I think it's the cleanest way to stop PrivacyRequest execution with Dask Delayed.

This also feels more similar to Sean's proposed design too which sequentially ordered these tasks in Celery.

src/fidesops/task/graph_task.py
```diff
@@ -497,14 +531,16 @@ def run_access_request(
...
+            seed: List[Dict[str, Any]]
+        ) -> Callable[[], List[Dict[str, Any]]]:
             """Return a function that returns the seed value to kick off the dask function chain.
 
             The first traversal_node in the dask function chain is just a function that when called returns
             the graph seed value."""
 
             def g() -> List[Dict[str, Any]]:
-                return [seed]
+                return seed
```
pattisdr, line 543: This modification just allows us to just reuse this function for both root nodes and for the nodes that just return data they already have on restart.

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -225,6 +211,47 @@ def run(
...
+                    request_id=privacy_request.id,
+                    data=filtered_results,
+                    storage_key=rule.storage_destination.key,
+                )
+            except common_exceptions.StorageUploadError as exc:
+                logging.error(
+                    f"Error uploading subject access data for rule {rule.key} on policy {policy.key} and privacy request {privacy_request.id} : {exc}"
+                )
+                privacy_request.status = PrivacyRequestStatus.error
+
```
pattisdr, line 254: I just pulled this out because PrivacyRequestRunner.run is getting long, and it's a specific step that should not run if we're resuming the privacy request from the erasure portion.  There's some logic that's been added to PrivacyRequestRunner.run to allow it to be used to resume from both the access and erasure portions of the request.

src/fidesops/task/graph_task.py
```diff
@@ -72,6 +73,13 @@ def result(*args: Any, **kwargs: Any) -> List[Optional[Row]]:
                         self.log_start(action_type)
                     # Run access or erasure request
                     return func(*args, **kwargs)
+                except PrivacyRequestPaused as ex:
+                    logger.warning(
+                        f"Privacy request {method_name} paused {self.traversal_node.address}"
+                    )
+                    self.log_paused(action_type, ex)
+                    # Re-raise to stop privacy request execution.
+                    raise
```
pattisdr, line 82: This is a primary change.  

It's worth pointing out that in our current design, if a node fails three times on retry, we catch that final exception, and just pass an empty list or a 0 downstream, still allowing the request to complete.  The Pause feature (and soon, the Restart after Exception feature) actually stops execution, cancelling other remaining tasks, rather than sending dummy values downstream.

seanpreston:
Some initial thoughts:

> This implementation assumes that the user hasn't updated their collections since the privacy request has paused

This is great to log, and I don't think this should be a blocker for now. The reality is that if collection datasets change midway through a request's execution, it's likely the underlying datastores have also changed, and the existing request will be liable to error due to misconfiguration. If we proceed with this change we have a slightly different manifestation of the problem, but the problem nonetheless.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -569,6 +576,93 @@ def resume_privacy_request(
...
+    dependencies=[
+        Security(verify_oauth_client, scopes=[PRIVACY_REQUEST_CALLBACK_RESUME])
+    ],
+)
+def resume_with_manual_input(
+    privacy_request_id: str,
+    *,
+    db: Session = Depends(deps.get_db),
+    cache: FidesopsRedis = Depends(deps.get_cache),
+    manual_rows: List[Optional[Row]],
```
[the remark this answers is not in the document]
seanpreston: > we could surface it on any API endpoint that needs this for the frontend

This will be necessary for the UI. The user will need to know:

- Why is the graph paused?
- Where am I looking for data?
- What fields are we looking for?
- Which identity does this pertain to? (cc @NevilleS as this is another place we'll need to reveal identity data, with potentially more significant impact here as the data required for one datastore might be found inside the customer's internal systems vs. passed in from the requester themselves)
- Where do I submit the data to once found?

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -569,6 +576,93 @@ def resume_privacy_request(
...
+
+    If there's no manual data to submit, pass in an empty list to resume the privacy request.
+    """
+    privacy_request: PrivacyRequest = get_privacy_request_or_error(
+        db, privacy_request_id
+    )
+    if privacy_request.status != PrivacyRequestStatus.paused:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Invalid resume request: privacy request '{privacy_request.id}' status = {privacy_request.status.value}.",
```
seanpreston, line 633: Can we change this to indicate that a privacy request must be paused to achieve success?

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -569,6 +576,93 @@ def resume_privacy_request(
...
+        )
+
+    paused_step: Optional[ActionType]
+    paused_collection: Optional[CollectionAddress]
+    paused_step, paused_collection = privacy_request.get_paused_step_and_collection()
+    if not paused_collection:
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail=f"Cannot resume privacy request '{privacy_request.id}'; no paused collection.",
```
seanpreston, line 642: Are there any valid states in the flow where a request can have a paused state, but no paused collection? If not, we might want to move the request into the errored state here.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
pattisdr: Yes, exposing this information isn't included in this PR because it's getting long and there were unknowns around what we should expose and where. This is a good starting list; I think we should reticket.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: Yes I can rephrase, thanks

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: Yes, the privacy request can also be paused due to webhooks that run before we even build the graph. In that case, the privacy request is automatically resumed via a `reply-to` URL that we pass as a header when we call the  client-defined endpoint originally.

docs/fidesops/docs/guides/database_connectors.md
```diff
@@ -254,6 +254,17 @@ you should adjust the ConnectionConfig Secrets properties through additional cal
...
+Once you have a working ConnectionConfig, it can be associated to an existing [dataset](datasets.md) by calling the `/dataset` endpoint, with a JSON version of your dataset as the request body:
+
+```json title="<code>PATCH /api/v1/connection/my_connection_key/dataset</code>"
+[{
+    "fides_key": "example_test_dataset",
+    "name": "Example Test Dataset",
+    "description": "Example of a dataset containing a variety of related tables like customers, products, addresses, etc.",
+    "collections": [...]
+}]
+```
```
conceptualshark, line 268: Ended up putting this in the connectors guide for now, just for simplicity; I think it will move to datasets once the docs get cleaned up a bit further.

pattisdr: wonderful thank you

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
pattisdr: We also don't necessarily need the identity data, we need whatever data was passed into that node.  It could have been identity data, but it could also be any other type of data that the user has configured as being an upstream dependency.

src/fidesops/service/connectors/manual_connector.py
```diff
@@ -0,0 +1,62 @@
...
+from fidesops.service.connectors.base_connector import DB_CONNECTOR_TYPE, BaseConnector
+from fidesops.util.collection_util import Row
+
+
+class ManualConnector(BaseConnector[None]):
+    def query_config(self, node: TraversalNode) -> None:
+        """No query_config for the Manual Connector"""
+        return None
+
+    def create_client(self) -> DB_CONNECTOR_TYPE:
```
seanpreston, line 16: Should this be `Optional[DB_CONNECTOR_TYPE]` given this method returns `None`?

src/fidesops/task/graph_task.py
seanpreston: As I understand it the pause feature and auto retry feature are distinct. Am I right in thinking that pause on error will only pause after the configured auto retries are exhausted?

pattisdr: they are different features, but they interact because auto-retry catches the first paused exception here. And no, pause re-raises here, so it doesn't automatically retry.  If a node was just paused, you wouldn't have any "retry" execution logs created for it.

pattisdr: unlike the "pause" feature, in which I don't have any retries, for the similar feature to "restart after failure", I'm thinking we still retry for the configured amount of times in case it is a temporary glitch, and then re-raise the error the final time, cancelling the remaining tasks, rather than sending the zero values downstream.

tests/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1610,3 +1613,177 @@ def test_resume_privacy_request(
...
+        )
+        response = api_client.post(url, headers=auth_header, json=[{"mock": "row"}])
+        assert response.status_code == 422
+        assert (
+            response.json()["detail"]
+            == "Cannot save manual rows. No 'mock' field defined on the 'manual_input:filing_cabinet' collection."
+        )
+
+    @pytest.mark.usefixtures(
+        "postgres_example_test_dataset_config", "manual_dataset_config"
```
seanpreston, line 1705: If I'm understanding this correctly `manual_dataset_config` overlays `postgres_example_test_dataset_config` for these tests, and in test scenarios where we don't want to test manual input, we can overlay the same `dataset_config` as before?

pattisdr: no, this is something Paul showed me, `pytest.mark.usefixtures` lets you separate the fixtures that you don't actually need to access within the function.   These tests still use both the postgres and manual datasets however.

pattisdr:
> Ideally we'd add some extra execution logs for these manual nodes, one when the request pauses and another when the request resumes (logging whether data is submitted, or the request has been resumed after an empty search).

Makes sense, I have a lot of this, I recommend fine-tuning in a follow-up ticket where we surface information for the UI.  I'm currently creating "paused" execution logs, and then another in-processing log when the request resumes. We could add more detail to these logs. I'm also caching the current paused collection, and the current step that is paused (access or erasure).  

> Why is the graph paused? (if there's a collection cached, it's because we're waiting on manual input)
  Where am I looking for data? (should return the name of the cached paused collection)
  What fields are we looking for? (should return the fields defined on the collection in the dataset)
  Which identity does this pertain to? (should return the data that is cached for any upstream dependencies of this node)
  Where do I submit the data to once found? (there will be two API endpoints, one for access and one for erasure)

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
pattisdr: Followup added here, https://github.com/ethyca/fidesops/issues/570

tests/api/v1/endpoints/test_privacy_request_endpoints.py
seanpreston: Got it, thanks

PR #577, opened 2022-05-27
Refactor auth and enable static file serving

elliotbonneville:
# Purpose

Make it possible to run the Next application without relying on the Node server by completely rewriting authentication layer and disabling image optimization.

# Changes

- Rewrite authentication layer and introduce auth slice to store auth state client-side
- Remove `next-redux-wrapper` and all hydration since we no longer render anything serverside
- Clean up user management code to use new auth slice
- Disable image optimization
- Update the `build` script to execute `next build && next export` to the `out` dir for static files

# Checklist

# Ticket

Fixes https://github.com/ethyca/fidesui/issues/2


clients/admin-ui/src/features/privacy-requests/DenyPrivacyRequestModal.tsx
```diff
@@ -44,37 +44,37 @@ const DenyPrivacyRequestModal = ({
     returnFocusOnClose={false}
   >
     <ModalOverlay />
-    <ModalContent width='100%' maxWidth='456px'>
+    <ModalContent width="100%" maxWidth="456px">
```
elliotbonneville, line 47: @TheAndrewJackson I know you had set `jsxSingleQuote` to true, but is there any chance we can keep double quotes in JSX? I've never worked with JSX single quotes before. I don't want to be pedantic but I'm wondering what the purpose of them is? Just consistency?

TheAndrewJackson: We can definitely keep double quotes. I used single because I was following someone else who used single quotes. If double quotes are preferred then let's use them instead. Especially if the `fidesctl` repo is using double quotes. 

It's not a big deal anyway. As long as the formatter automatically sets it I'm fine with either.

elliotbonneville: Cool. Let's update the Prettier file to use double quotes then – thanks for your flexibility here!

clients/admin-ui/README.md
```diff
@@ -17,12 +17,8 @@ Admin UI for managing FidesOps privacy requests. A web application built in Next
 5. Submit a privacy request through the Privacy Request center.
 6. View that request in the Admin UI and either approve or deny it.
 
-## Authentication
+## Unit test locations
 
-To enable stable authentication you must supply a `NEXTAUTH_SECRET` environment
-variable. The best way to do this is by creating a `.env.local` file, which Next
```
TheAndrewJackson, line 23: Is the `.env.local` file no longer needed for auth to work properly?

clients/admin-ui/src/pages/_app.tsx
```diff
@@ -4,23 +4,31 @@ import '@fontsource/inter/700.css';
...
 import theme from '../theme';
 
 if (process.env.NEXT_PUBLIC_MOCK_API) {
   // eslint-disable-next-line global-require
   require('../mocks');
 }
 
+const SafeHydrate: React.FC = ({ children }) => (
+  <div suppressHydrationWarning>
+    {typeof window === 'undefined' ? null : children}
```
TheAndrewJackson, line 20: Why is the hydration warning saying and doing? I'm guessing it's an annoying error message that we don't need while developing

clients/admin-ui/src/pages/user-management/new.tsx
```diff
@@ -6,58 +6,42 @@ import {
...
-        <Heading fontSize='2xl' fontWeight='semibold'>
-          User Management
-          <Box mt={2} mb={7}>
-            <Breadcrumb fontWeight='medium' fontSize='sm'>
-              <BreadcrumbItem>
-                <BreadcrumbLink href='/user-management'>
-                  User Management
-                </BreadcrumbLink>
-              </BreadcrumbItem>
+  <ProtectedRoute>
```
TheAndrewJackson, line 17: The `ProtectedRoute` component seems more elegant than the original `getServerSideProps`
solution.  Just wrap the JSX with 1 more component and it's being authenticated.

TheAndrewJackson (commented):
It looks good! I like a lot of the refactors and the tests. I finished my first pass of reading all of the code. I just have a couple small nits and questions. I learned a few things while reading this PR!

I'm going to do some functional testing now and see if anything visually pops up that I couldn't see while reviewing it.

clients/admin-ui/src/pages/subject-request/[id].tsx
```diff
@@ -8,94 +8,67 @@ import {
...
-const SubjectRequestDetails: NextPage<{}> = () => {
-  const filters = useAppSelector(selectPrivacyRequestFilters);
-  const { data } = useGetAllPrivacyRequestsQuery(filters);
+const useSubjectRequestDetails = () => {
+  const router = useRouter();
+  const { id = '' } = router.query;
+  const { data } = useGetAllPrivacyRequestsQuery({
+    id: Array.isArray(id) ? id[0] : id,
+    verbose: true,
+  });
```
Comment (TheAndrewJackson) [inline clients/admin-ui/src/pages/subject-request/[id].tsx:25]: In my functional testing I found that this page was doing 2 requests to the backend. The first was with no filtering id so it would get back multiple requests. The second would have the id from the url as expected so it would filter and get the data.

This update makes it skip getting the data the url until the `const { id = '' } = router.query;` line above has parsed the id as expected. Now it will only do 1 request.

```suggestion
  const { data } = useGetAllPrivacyRequestsQuery(
    {
      id: Array.isArray(id) ? id[0] : id,
      verbose: true,
    },
    {
      skip: id === '',
    }
  );
```

clients/admin-ui/src/pages/subject-request/[id].tsx
```diff
@@ -8,94 +8,67 @@ import {
...
+
+  return data;
+};
+
+const SubjectRequestDetails: NextPage = () => {
+  const data = useSubjectRequestDetails();
   const body =
-    data?.items.length === 0 ? (
+    !data || data?.items.length === 0 ? (
       <Text>404 no subject request found</Text>
```
Comment (TheAndrewJackson) [inline clients/admin-ui/src/pages/subject-request/[id].tsx:34]: It looks like this page is flashing this message for a split second while the page is loading the data. If the user has a slow connection it will display the 404 message while the page is loading. I guess we'll need to incorporate some kind of loading behavior now since we're no longer doing SSR. That might be out of scope of this PR though

clients/admin-ui/__tests__/index.test.tsx
```diff
@@ -1,16 +1,19 @@
...
 describe('Home', () => {
-  it('renders the Subject Requests page by default', () => {
-    render(
-      <SessionProvider>
-        <Home />
-      </SessionProvider>
-    );
+  it('renders the Subject Requests page by default when logged in', () => {
+    render(<Home />, {
+      preloadedState: {
```
allisonking, line 8: whoa this is very cool 🤩

clients/admin-ui/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "fidesui-template",
+  "name": "fidesops-admin-ui",
   "private": true,
   "scripts": {
     "dev": "next dev",
     "dev:mock": "echo '🚨 Running with mock API'; NEXT_PUBLIC_MOCK_API=true next dev",
-    "build": "next build",
+    "build": "next build && next export",
```
allisonking, line 7: we'll probably want to standardize this across the repos. fidesctl looks like this right now: https://github.com/ethyca/fides/blob/aking-706-edit-collection/clients/admin-ui/package.json/#L15-L17

clients/admin-ui/src/constants.ts
```diff
@@ -1,51 +1,9 @@
...
-  username?: string;
-  password?: string;
-  created_at?: string;
-}
+import { UserPrivileges } from './features/user-management/types';
 
-export interface UserResponse {
-  id: string;
-}
+// eslint-disable-next-line import/prefer-default-export
```
allisonking, line 3: I think there are >1 exports now, so don't need this line anymore?

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -0,0 +1,101 @@
+import {
+  createListenerMiddleware,
+  createSlice,
+  PayloadAction,
+} from '@reduxjs/toolkit';
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
+
+import type { RootState } from '../../app/store';
```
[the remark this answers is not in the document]
allisonking: same!

clients/admin-ui/src/pages/login.tsx
```diff
@@ -11,206 +11,223 @@ import {
...
+    if (!values.email) {
+      errors.email = 'Required';
+    }
 
-      if (!values.password) {
-        errors.password = 'Required';
-      }
+    if (!values.password) {
+      errors.password = 'Required';
+    }
```
allisonking, line 67: a note for the future, I'd love to start using [yup](https://github.com/jquense/yup) for validation

allisonking (commented):
so much refactoring, amazing! 🤩  I don't have much to add outside of what Andrew already caught (👏 ), just some comments here and there. thanks for doing this!!

clients/admin-ui/README.md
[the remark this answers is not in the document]
elliotbonneville: Nope. This was an artifact of the previous library, which we're no longer using.

clients/admin-ui/src/features/auth/auth.slice.ts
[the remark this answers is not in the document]
elliotbonneville: It's most helpful when you need types from two files that depend on one another – it avoids the circular dependency problem because the types can be statically inferred and the files don't need to be imported and executed, which is where you'd get that import loop.

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+});
+
+export const selectAuth = (state: RootState) => state.auth;
+export const selectUser = (state: RootState) => selectAuth(state).user;
+export const selectToken = (state: RootState) => selectAuth(state).token;
+
+export const { login, logout } = authSlice.actions;
+
+export const credentialStorage = createListenerMiddleware();
```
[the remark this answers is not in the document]
elliotbonneville: Yup! They make it really easy to build middleware in, and this is where any side effects of actions should be specified (you don't want to do that in reducers, since they need to be pure and not affect any external state whatsoever).

clients/admin-ui/src/features/auth/auth.slice.ts
```diff
@@ -0,0 +1,101 @@
...
+  baseQuery: fetchBaseQuery({
+    baseUrl: process.env.NEXT_PUBLIC_FIDESOPS_API!,
+    prepareHeaders: (headers, { getState }) => {
+      const token = selectToken(getState() as RootState);
+      headers.set('Access-Control-Allow-Origin', '*');
+      if (token) {
+        headers.set('authorization', `Bearer ${token}`);
+      }
+      return headers;
+    },
```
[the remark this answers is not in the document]
elliotbonneville: Yep, pretty sure they do! That's a great idea.

clients/admin-ui/src/pages/_app.tsx
elliotbonneville: That's correct. Basically it's noting that what gets rendered on the server and what gets rendered on the client isn't the same, because we're no longer taking advantage of serverside rendering. This won't apply at all without a server, so the warning is meaningless.

clients/admin-ui/src/pages/user-management/new.tsx
elliotbonneville: Yep! To be fair, we can definitely improve on the `getServerSideProps` pattern, but I hadn't cleaned that up because I knew we were intending to rip it out.

I will say that one thing I noticed with this pattern is that it's easily possible to forget to authenticate a page, so we just need to either a) wrap all of the pages by default at the `App.tsx` level, or stay on top of that while developing. I'm not super concerned, since the requests wouldn't be authenticated anyhow, so even if we do forget to wrap a page it's not the end of the world – and we don't create pages that frequently, so the error surface area is low. But it's just another place where we could introduce an error.

To be fair, that same logic applied to the previous solution too, so I guess it's not a new problem. Something to think about...

clients/admin-ui/src/pages/subject-request/[id].tsx
Comment (elliotbonneville) [reply clients/admin-ui/src/pages/subject-request/[id].tsx:40]: Good spot. Between Chakra UI and Redux Toolkit Query we actually have everything we need to solve this in like 5min, so I pushed up a quick fix for this. :)

clients/admin-ui/package.json
elliotbonneville: Yeah. Do you know off the top of your head if Next can export to a location in a parent directory? Would simplify a lot of the copy / paste / file moving stuff.

clients/admin-ui/src/constants.ts
elliotbonneville: Yup yup yup!

elliotbonneville:
Thanks very much for your thorough review and great catches, folks! Really appreciate you taking the time to look this over and again, sorry about such a giant PR! Heh.

clients/admin-ui/src/features/user-management/EditUserForm.tsx
```diff
@@ -194,13 +201,13 @@ const EditUserForm: NextPage<{
...
-                isDisabled={!hasAdminPermission}
+                isReadOnly={!canUpdateUser}
+                isDisabled={!canUpdateUser}
               />
             </FormControl>
 
-            {/* Only the associated user can change their own password */}
-            {id === user.id && <UpdatePasswordModal id={id} />}
+            {/* Only the owner of this profile can change their password */}
+            {isOwnProfile ? <UpdatePasswordModal id={profileId} /> : null}
```
LKCSmith, line 210: 🎉

clients/admin-ui/src/pages/login.tsx
LKCSmith: @allisonking Is there a benefit to using yup that I'm not seeing? 👀

LKCSmith (changes requested):

clients/admin-ui/src/pages/login.tsx
allisonking: I like that `yup` makes validation declarative (you declare all the requirements a field has to meet at the same time you define the form (i.e. required, minimum 8 letters, etc.) and then it'll do the validation on submit (or I think you can also do it on touch) so then you don't have to do manual validation by looking through all the fields yourself with a bunch of `if`s. since validation is defined in a schema, you can also reuse that schema which is handy. it's also nice that formik supports yup out of the box and they recommend it in their tutorial too: https://formik.org/docs/tutorial#schema-validation-with-yup

TheAndrewJackson (approved):
I just finished my second pass and I think everything looks good now 🙂  I will be comfortable with merging once this PR also gets an approval from @allisonking and @LKCSmith

clients/admin-ui/package.json
allisonking: looks like the answer is no, it has to stay in the project dir 

https://nextjs.org/docs/api-reference/next.config.js/setting-a-custom-build-directory

> distDir should not leave your project directory. For example, ../build is an invalid directory.

allisonking (approved):
🚀

Issue #570, opened 2022-05-26
API: Surface Pause/Resume Privacy Request Details 


pattisdr:
As a followup to #521 and #522 which add support for pausing and resuming privacy requests during access and erasure requests, surface details about where the privacy request is paused and what attributes are needed to resume the request for the UI.

- Why is the graph paused?
- Where am I looking for data?
- What fields are we looking for?
- Which upstream data is needed?
- Where do I submit the data to once found?

# Questions
- Should we surface in a new endpoint or an existing endpoint?

Issue #361, opened 2022-04-05
Hubspot Owners endpoint


adriaaaa:
### Is your feature request related to a specific problem?

There seems to be an issue with getting a response from the Owner's endpoint in hubspot. We will need to investigate further and potentially reach out to their support department to figure out how to get it working. Once we do, we will need to add that endpoint to our Connector.

### Acceptance Criteria
1. Support access and erasure request to the `owners` endpoint, if possible
2. Update the [spec](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2212135123/Hubspot+Connector+Specification) with relevant info 

Aha! Link: https://ethyca.aha.io/features/OPS-90

PR #578, opened 2022-05-28
Restart Graph from Failure [#574]

### LINKED ISSUE #574: Ability to Restart Graph From Failure
# What
If a particular collection fails during an access or an erasure request, raise an exception and cancel other tasks in the graph. Allow the privacy request to be restarted from the failure point.

# Why
We currently retry a failed collection a specified number of times. If the collection continues to fail after a certain number of retries, we **continue with the graph execution**.  We assume that the failed collection didn't return data, and that no data was masked. Downstream collections still run.  

This can be problematic for both data in the failed collection and downstream collections whose data potentially wasn't retrieved or masked.  Running another privacy request may not rectify the issue because data may have been destroyed that prevents us from reaching the collection in question again.

We will still keep the retry, in case the failure was just temporary, but stopping execution entirely allows the user to go correct something on their end before resuming.

pattisdr:
# Purpose

We currently can't rerun a graph from a failed node, our only choice is to attempt to run a completely new privacy request.

If a node in the graph fails a certain number of times, we continue with graph execution, just passing in empty values downstream to dependent nodes.  This is problematic for a couple of reasons: we may not have properly retrieved/masked data on the failed collection, or downstream collections.  Second, if we want to run another privacy request to rectify this, we may no longer be able to execute the graph because data has been destroyed.

# Changes
- Raise an exception when a graph node fails (instead of ignoring after "x" retries), and cancel remaining graph tasks.    Sending a POST request to `/privacy-request/{privacy_request_id}/retry` will restart the graph, only running remaining graph tasks.
- Fix existing issues with Bigquery/Hubspot/Stripe either in the code or the tests, whose collection failures were previously failing silently.  Now that a collection failure causes the entire privacy request to fail, more errors are being raised than there were previously.

# Note
- This doesn't yet expose to the frontend which node is the "failed" node; that will be in follow-up #570 
- This still preserves the "retry" behavior, except after "x" retries we re-raise an exception and exit, instead of continuing on.

# Checklist

# Ticket

Fixes #574


src/fidesops/models/privacy_request.py
```diff
@@ -378,7 +377,9 @@ def start_processing(self, db: Session) -> None:
         """Dispatches this PrivacyRequest throughout the Fidesops System"""
         if self.started_processing_at is None:
             self.started_processing_at = datetime.utcnow()
-            self.save(db=db)
+        if self.status == PrivacyRequestStatus.pending:
+            self.status = PrivacyRequestStatus.in_processing
+        self.save(db=db)
```
pattisdr, line 382: If a privacy request does not need approval, it seems like this should go to in_processing when it starts processing instead of staying on 'pending'

pattisdr:
@ethyca/docs-authors minor edit to guide added here

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -192,8 +192,11 @@ def run(
 
             except BaseException as exc:  # pylint: disable=broad-except
                 privacy_request.status = PrivacyRequestStatus.error
+                privacy_request.save(db=session)
                 # If dev mode, log traceback
                 _log_exception(exc, config.dev_mode)
+                session.close()
+                return
```
seanpreston, line 199: This will stop the post processing webhooks running when the request errors out, which is an interesting nuance that I don't think we've thought about at a product level. For example, in the event that webhooks are used to teardown infra (a requested case previously in Atlas), this would have the implication of that teardown work not being done. This may also mean we need to reconsider the pre-execution hooks in the event of a resume too. @adriaaaa do we have a preference here?

seanpreston: If we decide to leave this, then we no longer need the check on line `:212`

seanpreston:
Thanks @pattisdr, I just have that one product level question re: the webhooks

src/fidesops/service/privacy_request/request_runner_service.py
pattisdr: From offline discussion, we'll go with the current design - we'll assume webhooks should just run once. In the future, we can build out fancier webhooks that encompass a single collection or only run on failure, or only run on pause, for example.

Removed the check on L212.

CHANGELOG.md
```diff
@@ -40,7 +42,6 @@ The types of changes are:
 * Added `FIDESOPS__DATABASE__ENABLED` and `FIDESOPS__REDIS__ENABLED` configuration variables to allow `fidesops` to run cleanly in a "stateless" mode without any database or redis cache integration [#550](https://github.com/ethyca/fidesops/pull/550)
 * Pause Access Request Execution / Resume on Manual Input in [#554](https://github.com/ethyca/fidesops/pull/554)
 * A `[package]` section of the `fidesops.toml` configuration file may specify the path to the `fidesops` package itself [#566](https://github.com/ethyca/fidesops/pull/566)
-* Pause Erasure Request Execution / Resume on Manual Input in [#571](https://github.com/ethyca/fidesops/pull/571/)
```
pattisdr, line 43: This is incorrect, it should be in the current release

conceptualshark (approved):

tests/integration_tests/saas/test_stripe_task.py
```diff
@@ -1071,9 +1071,8 @@ def test_stripe_erasure_request_task(
         ],
     )
 
-    # run erasure with MASKING_STRICT to execute the update actions
-
-    config.execution.MASKING_STRICT = True
+    # Run erasure with masking_strict = False so both update and delete actions can be used
+    config.execution.MASKING_STRICT = False
```
pattisdr, line 1076: Stripe tests were incorrectly being run with `config.execution.MASKING_STRICT = True` which is invalid for Stripe, because there are both updates and deletes defined in the config.  Nodes with `delete`-only configs were failing silently and then Stripe tests were re-run with `config.execution.MASKING_STRICT = False` below to get delete-specific behavior.

Because a collection no longer fails silently, we were seeing failures here. To address, we can just run a single erasure request with `config.execution.MASKING_STRICT = False` so stripe can use the update if defined, otherwise it uses the delete.  The counts below have been updated to reflect this.

data/saas/dataset/hubspot_dataset.yml
```diff
@@ -91,7 +91,6 @@ dataset:
           - name: id
             data_categories: [user.derived.identifiable.unique_id]
             fidesops_meta:
-              primary_key: True
```
pattisdr, line 94: We currently shouldn't attempt to run an erasure against hubspot owners' endpoint. https://github.com/ethyca/fidesops/issues/361.  Our tests were attempting to run this and failing silently.

This PR doesn't allow it to fail silently anymore, so this adjustment prevents us from running an erasure against hubspot owners' until we can sort out how to connect to that endpoint.

tests/integration_tests/saas/test_stripe_task.py
```diff
@@ -1071,9 +1071,8 @@ def test_stripe_erasure_request_task(
...
```
seanpreston, line 1075: Ideally we wouldn't set this in the test since if execution halts mid-test the value won't be reset and could cause cascade failures within other tests. Looks like you're only updating the value here so let's change these as part of a [subsequent ticket](https://github.com/ethyca/fidesops/issues/606).

seanpreston (approved):
Thanks @pattisdr

PR #579, opened 2022-05-30
Improve configuration variable troubleshooting with DEBUG logs, more helpful errors, and unit tests

NevilleS:
# Purpose
After watching @RobertKeyser run into some mysterious errors setting up the `fidesops` configuration values recently, I thought I'd make a few QOL improvements to the debug-ability of our config.

# Changes
- Changed the error message for `FIDESOPS__SECURITY__APP_ENCRYPTION_KEY` to output the number of characters to aid troubleshooting
- Added `FIDESOPS__SECURITY__LOG_LEVEL` configuration variable to allow controlling the log level [#579](https://github.com/ethyca/fidesops/pull/579)
- Added `DEBUG` logs at startup to view all configuration values [#579](https://github.com/ethyca/fidesops/pull/579)
- Added unit tests for `FidesopsConfig` loading and tests for `APP_ENCRYPTION_KEY` validation

# Checklist
- [X] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [X] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [X] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [X] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [X] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
  - [X] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - n/a documentation issue created (tag docs-team to complete issue separately)
- [X] Good unit test/integration test coverage
- n/a This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [X] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket
n/a


src/fidesops/core/config.py
```diff
@@ -266,15 +302,15 @@ def get_config() -> FidesopsConfig:
     """
     try:
         return FidesopsConfig.parse_obj(load_toml("fidesops.toml"))
-    except (FileNotFoundError, ValidationError) as e:
```
NevilleS, line 269: Catching the `ValidationError` here was odd; it basically made it try loading the config twice:
- once from the file, which if invalid threw an error
- then once again from defaults, which may throw an error

This is surprising, because if you have an invalid config it fails twice, and usually with different error messages, which is misleading.

NevilleS: By removing this catch, the validation error simply crashes startup and throws a more relevant warning log and stacktrace.

src/fidesops/main.py
```diff
@@ -38,6 +38,14 @@
 def start_webserver() -> None:
     """Run any pending DB migrations and start the webserver."""
     logger.info("****************fidesops****************")
+
+    if logger.getEffectiveLevel() == logging.DEBUG:
+        logger.warning(
+            "WARNING: log level is DEBUG, so sensitive or personal data may be logged. "
+            "Set FIDESOPS__SECURITY__LOG_LEVEL to INFO or higher in production."
+        )
```
NevilleS, line 46: I'm increasingly of the opinion we should remove the `LOG_PII` variable, and just use this single `LOG_LEVEL` instead. Having DEBUG vs INFO logs is a standard config option most apps support, so it'll be more intuitive to use this.

NevilleS: NOTE: I really wanted to DEBUG log all the `ENV` vars at startup too, but by the time this method runs the `FidesopsConfig` has already been loaded (and potentially crashed), so it's too late. Since we're using the config to set the log level simultaneously, it's awkward to use DEBUG logs to see the `ENV` vars... kind of a chicken & egg situation.

tests/core/test_config.py
```diff
@@ -0,0 +1,109 @@
...
+import os
+from unittest.mock import patch
+
+import pytest
+from pydantic import ValidationError
+
+from fidesops.core.config import get_config
+
+
+def test_config_from_default() -> None:
```
NevilleS, line 11: These first three tests (load from default, load from path, load from ENV) didn't have any tests yet so I added some basic ones

src/fidesops/core/config.py
```diff
@@ -214,6 +243,14 @@ class Config:  # pylint: disable=C0115
         f'Startup configuration: pii logging = {os.getenv("FIDESOPS__LOG_PII") == "True"}'
     )
 
+    def log_all_config_values(self) -> None:
+        """Output DEBUG logs of all the config values."""
+        for settings in [self.database, self.redis, self.security, self.execution]:
+            for key, value in settings.dict().items():
+                logger.debug(
+                    f"Using config: {NotPii(settings.Config.env_prefix)}{NotPii(key)} = {NotPii(value)}"
```
eastandwestwind, line 251: In order for vals to be masked properly, we have to use `%` string interpolation instead of `f`. This is because we intercept the Python `LogRecord` when we check for PII args, and `LogRecord` is only compatible with `%` str interpolation- https://docs.python.org/3/library/logging.html#logrecord-attributes. We could not find a solution to intercept `f` formatted strings.

So, even though we're saying `NotPii` here, mind updating to `%` for consistency in approach?

src/fidesops/core/config.py
eastandwestwind: good catch, I'll actually have to update this on my fideslog integration branch

src/fidesops/core/config.py
```diff
@@ -214,6 +243,14 @@ class Config:  # pylint: disable=C0115
         f'Startup configuration: pii logging = {os.getenv("FIDESOPS__LOG_PII") == "True"}'
     )
 
+    def log_all_config_values(self) -> None:
```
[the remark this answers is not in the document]
eastandwestwind: this will be super helpful, thanks!

data/config/fidesops.toml
```diff
@@ -0,0 +1,33 @@
+# Configuration TOML used in select unit tests
+
+[database]
```
eastandwestwind, line 3: so will this config get loaded on top of the base `fidesops.toml` for unit/integration tests, or instead of? 

Can you explain more re in which circumstances this file will load?

eastandwestwind (commented):
Looking good @NevilleS, just wanted additional clarify on the `fidesops.toml` you've added, due to the fact that I'm working on testing new env variables for fideslog that may need to be added to this toml, too.

data/config/fidesops.toml
NevilleS: ah- this file is currently only used in the *one or two* tests for `test_config.py`, where you can see the `FIDESOPS__CONFIG_PATH` get set. It's not needed anywhere else

src/fidesops/core/config.py
NevilleS: Ah, OK I see. Will do

src/fidesops/core/config.py
NevilleS: Ideally let's get this merged to `main` so you can just rebase 👍

src/fidesops/core/config.py
NevilleS: OK, updated

eastandwestwind (approved):

PR #614, opened 2022-06-08
Bump fastapi dependencies

sanders41:
# Purpose

Bumps FastAPI and FastAPI's dependencies to allow installing fideslib.

There are 2 failing tests, [test_generate_field](https://github.com/ethyca/fidesops/blob/d4ebae80d6848105ac092e8acafde0c6322e9f9f/tests/graph/test_config.py#L240) and [test_field_collect_matching](https://github.com/ethyca/fidesops/blob/d4ebae80d6848105ac092e8acafde0c6322e9f9f/tests/graph/test_config.py#L332) after updating the dependencies. The dependency causing an issue is Pydantic. Both tests are failing for the same reason. They are comparing two objects with nested objects for equality. In older versions of Pydantic, the nested objects were created with pointers so the equality test would pass because the objects were at the same memory address. Now Pydantic creates the objects by creating new instances of the objects. Because of the, even though the values contained in the objects are the same, the tests fail because the test object id's are no longer the same. Any suggestions on getting around this?

# Changes
- FastAPI bumped and dependencies installed with `fastapi[all]` removed from requirements.txt

# Checklist

# Ticket

Fixes #


PSalant726:
> even though the values contained in the objects are the same, the tests fail because the test object id's are no longer the same. Any suggestions on getting around this?

I believe we'll have to implement a "deep equality" check, like:
```python
class Something:
    def __eq__(self, other):
        return self.__dict__ == other.__dict__
```

sanders41:
The version of black that was pinned turned out to be incompatible with the newer version of Click so I'm bumping that also.

sanders41:
Thanks @PSalant726, I tried this initially and it didn't work, but I only did it at the last level and not the nested classes also. I'm thinking now maybe it will work if I take it all the way down the chain.

requirements.txt
```diff
@@ -1,36 +1,29 @@
 alembic==1.6.5
 APScheduler==3.8.0
-bcrypt~=3.2.0
 boto3~=1.18.14
-click==7.1.2
+click==8.1.3
```
seanpreston, line 4: What was the secret to getting this to pass? [I struggled with this yesterday](https://ethyca.slack.com/archives/C029HFVSC8L/p1654638146084379?thread_ts=1654616643.541969&cid=C029HFVSC8L)

PSalant726: Looks like `black` was also bumped to v22.3.0 in https://github.com/ethyca/fidesops/pull/614/commits/ada24f8e2ff3ad230c07d4703304ca5bcf4e74fa.

Issue #525, opened 2022-05-17
Ability to revoke/cancel a request


adriaaaa:
### Description
A privacy request manager may need to cancel a request in case it was submitted in error or approved in error. Furthermore, as part of Consumer Report's DRP specification, they do require support for revoking a request in the event that the consumer chooses to do so.

This may be challenging to do for in progress requests at this time. We have limited the scope of this ticket to support cancelling a request only if the request is in a NEW state.

### Acceptance Criteria
1. Demonstrate that a request can be cancelled (limit to only requests that haven't started processing aka NEW requests)
2. If a request was cancelled, demonstrate that the status of the request moves to "Cancelled"
- on the SR landing/list page, we should update the status dropdown to include "cancelled" as a filterable status
- you should not be able to restart cancelled requests (users should submit new requests)

Aha! Link: https://ethyca.aha.io/features/OPS-51

pattisdr:
https://github.com/ethyca/fidesops/issues/418

PR #621, opened 2022-06-09
Celery privacy request dispatch

seanpreston:
# Purpose

# Changes
-

# Checklist

# Ticket

Fixes #


docker-compose.yml
```diff
@@ -58,6 +58,25 @@ services:
     ports:
       - "0.0.0.0:6379:6379"
 
+  celery:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: celery -A fidesops.tasks beat -l info --logfile=/var/log/celery.log
```
NevilleS, line 65: This works in development, but for a production deployment we need to make the celery worker configuration more foolproof

Can we add a new CLI command like `fidesops worker`, which acts similarly to `fidesops webserver` except it starts celery instead of uvicorn?

seanpreston: In the interest of [unblocking the DRP revoke / cancel request feature](https://github.com/ethyca/fidesops/issues/525) I've added [this follow-up ticket](https://github.com/ethyca/fidesops/issues/663) for this suggestion

tests/tasks/test_celery.py
```diff
@@ -1,10 +1,10 @@
-def test_create_task(celery_app, celery_worker):
-    @celery_app.task
+def test_create_task(celery_session_app, celery_session_worker):
```
seanpreston, line 1: `celery_app` and `celery_worker` are function scoped fixtures, so a worker is started on a new thread for each function, which is very slow. `celery_session_app` and `celery_session_worker` are session scoped fixtures, so we'll only use one worker per test session, which is much faster.

pattisdr (commented):
nice job getting the pytest portion of this working, I'd just like to verify things are working when we run locally.  Fidesops doesn't seem to be talking to the celery worker, as if it doesn't have the right configuration.

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -177,10 +177,7 @@ def create_privacy_request(
             )
 
             if not config.execution.REQUIRE_MANUAL_REQUEST_APPROVAL:
-                PrivacyRequestRunner(
-                    cache=cache,
-                    privacy_request=privacy_request,
-                ).submit()
+                run_privacy_request.delay(privacy_request.id)
```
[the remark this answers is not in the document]
seanpreston: This would imply `fidesops` isn't talking to the Redis backend as you say. Let me double check the connection details

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -28,231 +28,212 @@
     run_access_request,
     run_erasure,
 )
+from fidesops.tasks import app as celery_app
 from fidesops.tasks.scheduled.scheduler import scheduler
-from fidesops.util.async_util import run_async
```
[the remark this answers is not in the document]
seanpreston: @eastandwestwind was considering using these in a PR earlier I think. Catherine can you confirm if you'll need them?

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
seanpreston: You're right, the config wasn't being loaded in correctly. Have rectified.

src/fidesops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
eastandwestwind: we don't need async_util here

src/fidesops/tasks/__init__.py
```diff
@@ -1,8 +1,17 @@
...
-app.autodiscover_tasks(["fidesops.tasks", "fidesops.tasks.scheduled"])
+app.conf.update(broker_url=config.execution.CELERY_BROKER_URL)
+app.conf.update(result_backend=config.execution.CELERY_RESULT_BACKEND)
+app.autodiscover_tasks(
+    [
+        "fidesops.tasks",
+        "fidesops.tasks.scheduled",
+        "fidesops.service.privacy_request.request_runner_service",
+    ]
+)
```
pattisdr, line 14: have you managed to get this running locally outside of integration tests @seanpreston? I am not seeing any tasks being received or running at all.  I'm no longer getting 500's, but I don't think tasks are getting sent.  I'm not even sure that your run_privacy_task request is being registered:

```python3
>>> from celery import current_app
>>> current_app.tasks.keys()
dict_keys(['celery.chord', 'celery.accumulate', 'celery.chunks', 'celery.chord_unlock', 'celery.group', 'celery.map', 'celery.chain', 'celery.starmap', 'celery.backend_cleanup'])

```

src/fidesops/core/config.py
```diff
@@ -39,8 +39,8 @@ class ExecutionSettings(FidesSettings):
     TASK_RETRY_BACKOFF: int
     REQUIRE_MANUAL_REQUEST_APPROVAL: bool = False
     MASKING_STRICT: bool = True
-    CELERY_BROKER_URL: str
-    CELERY_RESULT_BACKEND: str
+    CELERY_BROKER_URL: str = "redis://:testpassword@redis:6379/1"
+    CELERY_RESULT_BACKEND: str = "redis://:testpassword@redis:6379/1"
```
pattisdr, line 43: Should we move this back to the fidesops.toml file?

src/fidesops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
pattisdr: Surfacing this again @seanpreston

fidesops.toml
```diff
@@ -29,8 +29,6 @@ DRP_JWT_SECRET = "secret"
 LOG_LEVEL = "INFO"
 
 [execution]
-CELERY_BROKER_URL = "redis://:testpassword@redis:6379"
-CELERY_RESULT_BACKEND = "redis://:testpassword@redis:6379"
```
pattisdr, line 33: these vars are still in the test toml file, but I think we'd want to restore them?

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -22,237 +23,219 @@
...
+) -> None:
+    # pylint: disable=too-many-locals
+    """
+    Dispatch a privacy_request into the execution layer by:
+        1. Generate a graph from all the currently configured datasets
+        2. Take the provided identity data
+        3. Start the access request / erasure request execution
+        4. When finished, upload the results to the configured storage destination if applicable
+    """
+    logger.info(f"Executing request with ID: {privacy_request_id}")
```
pattisdr, line 139: Now that you have logging working, is this still needed, given that we log a very similar thing on L149?

pattisdr (commented):
this is working well now Sean, nice work getting the all the config and the worker pieces sorted out.

my lingering concern is that that all tests were passing in earlier rounds of CR while there were several celery pieces still broken.  Have you thought about ways to end-to-end test this reliably?

src/fidesops/core/config.py
seanpreston: I thought about this. It can always be overridden from `fidesops.toml`, but the reality is that since the webserver always depends on the Redis cache it's safe to include as a default.

src/fidesops/util/cache.py
```diff
@@ -147,3 +147,7 @@ def get_all_cache_keys_for_privacy_request(privacy_request_id: str) -> List[Any]
     return cache.keys(f"{privacy_request_id}-*") + cache.keys(
         f"id-{privacy_request_id}-*"
     )
+
+
+def get_async_task_tracking_cache_key(privacy_request_id: str) -> str:
+    return f"id-{privacy_request_id}-async-execution"
```
pattisdr, line 153: In our current design I believe there could be multiple celery tasks for the same privacy request. For example, if you "pause" a request, behind the scenes, this stops executing remaining tasks and exits.  Resuming a privacy request would create another task for the same privacy request id.

I'm thinking we might need to differentiate here?

seanpreston: I may be wrong in my understanding here — does the initial task finish when the privacy request execution pauses? If so, we should be OK overwriting the privacy request's `task_id` on resume

src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -122,6 +128,29 @@ def upload_access_results(
             privacy_request.status = PrivacyRequestStatus.error
 
 
+def queue_privacy_request(
+    privacy_request_id: str,
+    from_webhook_id: Optional[str] = None,
+    from_step: Optional[str] = None,
+) -> str:
```
pattisdr, line 135: good idea here, since we've got this extra step now of caching the task idea

pattisdr (approved):
This looks good to me @seanpreston thanks for adding the extra step of caching the request id and asserting this in tests

PR #622, opened 2022-06-09
559-endpoint-log-events - Adds middleware for calling analytics events for each endpoint

### LINKED ISSUE #559: [Fideslog] Send an analytics event for each endpoint
For each deployment, we need to log:

* which API was accessed (and how - via the front end or directly via the API)
* if the call was successful or not
* if the call was not successful, we need to capture the error

eastandwestwind:
# Purpose
Adds middleware for analytics events on endpoints

# Changes
- Adds http middleware to `fidesops` endpoints that calls `fideslog` with endpoint path, response code, source, and error class
- Adds `X-Fides-Source` header to all endpoint calls from admin ui and privacy center ui
- Turns off `prefer-default-export` for eslint in admin-ui
- Turns on dev mode by default in fidesops- https://github.com/ethyca/fidesops/pull/622/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R27.

# Checklist

# Ticket

Fixes #559

### DIFF
--- a/src/fidesops/main.py
+++ b/src/fidesops/main.py
@@ -80,7 +80,7 @@ async def dispatch_log_request(request: Request, call_next: Callable) -> Respons
 
 def prepare_and_log_request(
     endpoint: str,
-    hostname: str,
+    hostname: Optional[str],
     status_code: int,
     event_created_at: datetime,
     fides_source: Optional[str],
--- a/src/fidesops/analytics.py
+++ b/src/fidesops/analytics.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from platform import system
+from typing import Optional
 
 from fideslog.sdk.python.client import AnalyticsClient
 from fideslog.sdk.python.event import AnalyticsEvent
@@ -17,7 +18,7 @@ def in_docker_container() -> bool:
     return bool(os.getenv("RUNNING_IN_DOCKER") == "true")
 
 
-def accessed_through_local_host(hostname: str) -> bool:
+def accessed_through_local_host(hostname: Optional[str]) -> bool:
     """`True`if the event was submitted through a local host, e,g, 127.0.0.1."""
     # testserver is hostname in unit tests
     LOCAL_HOSTS = ["localhost", "0.0.0.0", "127.0.0.1", "testserver"]


docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
         read_only: False
       - /fidesops/src/fidesops.egg-info
     environment:
-      - FIDESOPS__DEV_MODE=${FIDESOPS__DEV_MODE}
+      - FIDESOPS__DEV_MODE=True
```
eastandwestwind, line 27: Used to set `developer` (`bool`) when calling fideslog. Our reasoning why we hard-code this to `true` here is because if someone is using our default docker-compose file, that would be considered dev/test mode.

Someone actually using this in prod would need to create their own `docker-compose.yml` with specific config per their environment, as detailed here- https://ethyca.github.io/fidesops/getting_started/

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
...
+        # More context- https://github.com/tiangolo/fastapi/issues/1840
+        try:
+            if response.status_code >= 400:
+                response.background = BackgroundTask(
+                    prepare_and_log_request,
+                    endpoint,
+                    response.status_code,
+                    now,
+                    fides_source,
+                    "HTTPException",
```
eastandwestwind, line 65: error class can be assumed to be `HTTPException` if status code >=400 and it wasn't a thrown exception. In Starlette, only non-HTTP exceptions are thrown

tests/graph/test_graph.py
```diff
@@ -70,6 +70,7 @@ def test_retry_decorator(privacy_request, policy):
         CollectionAddress("postgres_example", "payment_card")
     ]
 
+    config.dev_mode = False
```
eastandwestwind, line 73: allow retries in test

clients/admin-ui/.eslintrc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     // causes bug in re-exporting default exports, see
     // https://github.com/eslint/eslint/issues/15617
     'no-restricted-exports': [0],
+    'import/prefer-default-export': 'off',
```
eastandwestwind, line 18: allows us to directly function method here- https://github.com/ethyca/fidesops/pull/622/files#diff-adfcb02b29653b385133da876800f29f50b8c5bc39875695ebdf2324c63525f7R6

We also have other instances throughout admin-ui where we don't use default exports, so I'm unsure why this wasn't caught before.

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
         allow_headers=["*"],
     )
 
+
+@app.middleware("http")
+async def dispatch_log_request(request: Request, call_next: Callable) -> Response:
+    fides_source: Optional[str] = request.headers.get("X-Fides-Source")
+    now: datetime = datetime.now(tz=timezone.utc)
+    endpoint = f"{request.method}: {request.url.path}"
```
eastandwestwind, line 50: right now HTTP method gets filtered out by a validator in fideslog, but this will be fixed in https://github.com/ethyca/fideslog/pull/66

clients/admin-ui/package-lock.json
```diff
@@ -19427,11 +19252,6 @@
       "integrity": "sha512-PkUpF+qoXTqhOeWL9fu7As8LXsIUZ1WYaJiY/a7McAQzxjk82OF0tibkFXVCDImZtWxbvojFjerkiLb0/q8mew==",
       "dev": true
     },
-    "pretty-format": {
-      "version": "3.8.0",
-      "resolved": "https://registry.npmjs.org/pretty-format/-/pretty-format-3.8.0.tgz",
-      "integrity": "sha1-v77VbV6ad2ZF9LH/eqGjrE+jw4U="
-    },
```
pattisdr, line 19434: Just making sure all these changes to the package-lock file were intentionally committed.

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
...
+            logger.warning(
+                f"Analytics event for endpoint {request.url} failed to send: {exc}"
+            )
+            return response
+
+        return response
+    except Exception as e:
+        logger.warning("exception caught")
+        prepare_and_log_request(
+            endpoint, e.args[0], now, fides_source, e.__class__.__name__
```
pattisdr, line 87: I don't think an exception is guaranteed to have an argument - in which case e.args[0] would throw an indexerror or similar. You also have status_code being an `int` below, but that's not guaranteed for this first argument to an exception.

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
         allow_headers=["*"],
     )
 
+
+@app.middleware("http")
+async def dispatch_log_request(request: Request, call_next: Callable) -> Response:
+    fides_source: Optional[str] = request.headers.get("X-Fides-Source")
```
pattisdr, line 48: If you're calling the API directly, I assume `fides_source` is None - is that how you're separating the data in fideslog?  You just assume if there's no fides_source, then it's an API call?

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
...
+                    now,
+                    fides_source,
+                    None,
+                )
+        except Exception as exc:
+            # always continue if something went wrong with analytics
+            logger.warning(
+                f"Analytics event for endpoint {request.url} failed to send: {exc}"
+            )
+            return response
```
pattisdr, line 81: Will we actually catch anything here in practice, since it looks like `send_analytics_event` also has a try/except?  It would be nice to get rid of this doubly-nested try/except.

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
...
+    status_code: int,
+    event_created_at: datetime,
+    fides_source: Optional[str],
+    error_class: Optional[str],
+) -> None:
+    analytics_event = AnalyticsEvent(
+        docker=in_docker_container(),
+        event=Event.endpoint_call.value,
+        event_created_at=event_created_at,
+        local_host=running_on_local_host(),
```
pattisdr, line 103: Why is this always true?

src/fidesops/main.py
```diff
@@ -40,6 +42,75 @@
         allow_headers=["*"],
     )
 
+
+@app.middleware("http")
+async def dispatch_log_request(request: Request, call_next: Callable) -> Response:
```
pattisdr, line 47: It seems like this middleware could be unit-tested, you could assert that certain calls are made.  That brings up a second question, how are you preventing analytics from being sent with unit tests?

src/fidesops/main.py
eastandwestwind: we're always running on localhost from the perspective of 1. always either running within the docker container or 2. running locally e.g. using pyenv or something. In this way, all ports/addresses are referencing internal/local addresses.

src/fidesops/main.py
eastandwestwind: Good call, I'll add some tests here. 

To answer your question, we add the `ANALYTICS_OPT_OUT` flag within `Makefile` when we run unit/integration tests or linters. This means that analytics will still be sent when bypassing Makefile and calling `pytest` directly from `make server` or pyenv. But, at least internal folks will have the `INTERNAL` client id to help with this.

sanders41: Could you add an autouse fixture to `conftest.py` to prevent it from sending? I haven't looked through yet to see if there is a good patch point for that so the answer maybe no :smile:

src/fidesops/main.py
eastandwestwind: Thanks, I'll check this out. 

This is also a rare exception clause that will catch if, for some reason, a non-HTTP exception is thrown from our endpoint that has a status code of <400. I don't expect this will ever happen, but may be good to have in case.

src/fidesops/main.py
eastandwestwind: correct, the only way we can discern direct api call vs not is by setting specific source headers from our clients.

src/fidesops/main.py
eastandwestwind: This `Exception` may catch issues with `BackgroundTask`. It was helpful when debugging async issues with tasks, but I guess it's not needed now that it's properly working. I can remove!

clients/admin-ui/package-lock.json
eastandwestwind: hmmm this may have been generated during a build? I'm unsure if I should commit - cc @TheAndrewJackson

src/fidesops/main.py
eastandwestwind: testing out something like this:

```
@pytest.fixture(autouse=True, scope="session")
def analytics_opt_out():
    """Disable sending analytics when running tests."""
    original_value = os.getenv("FIDESOPS__ROOT_USER__ANALYTICS_OPT_OUT")
    os.environ["FIDESOPS__ROOT_USER__ANALYTICS_OPT_OUT"] = "true"
    yield
    os.environ["FIDESOPS__ROOT_USER__ANALYTICS_OPT_OUT"] = original_value
```

clients/admin-ui/package-lock.json
TheAndrewJackson: I'm not 100% sure either. I double checked and it's not a dependency in our `package.json`. My guess is that it one our dependencies, let's say `prettier`, used to use this package and recently removed it. When you did an `npm install` it may have updated the `package-lock.json` and removed it because our dependency removed it. Since the tests are passing it seems safe.

Does everything still run fine on your local machine?

pattisdr: that makes sense Andrew. everything still runs fine, I was just making sure that that's what we're doing, we just commit the latest `package-lock.json` that was generated.

eastandwestwind: yup everything runs just fine locally

src/fidesops/main.py
eastandwestwind: Setting this os var isn't working as written above ^ and not sure why yet

eastandwestwind: looks like setting config worked:

```
@pytest.fixture(autouse=True, scope="session")
def analytics_opt_out():
    """Disable sending analytics when running tests."""
    original_value = config.root_user.ANALYTICS_OPT_OUT
    config.root_user.ANALYTICS_OPT_OUT = True
    yield
    config.root_user.ANALYTICS_OPT_OUT = original_value
```

src/fidesops/main.py
eastandwestwind: I guess technically a non-HTTP exception doesn't have a status code necessarily, and if it does there's no guarantee where status code would be. What do you think of just hard-coding `500` or something here?

pattisdr: well python exceptions don't have codes, the first argument that you're displaying above `e.args[0]` is usually the message.  For example, if I raised Exception("Bad code"), - "Bad code" is what your currently sending in as your status code.

pattisdr: 500 is fine because Python Exceptions don't have a code, but I'd think the actual message would be meaningful too, the fake 500 doesn't tell you much.

src/fidesops/main.py
sanders41: Looks like a good solution to me

tests/test_main.py
```diff
@@ -0,0 +1,139 @@
...
+            "headers": {}
+        },
+    )
+
+    datetime_mock.now.return_value = datetime(2022, 5, 22, tzinfo=timezone.utc)
+
+    f = asyncio.Future()
+    f.set_result(Response(status_code=200))
+
+    def mock_call_next(request: Request) -> Any:
```
eastandwestwind, line 47: @pattisdr I'm unsure if you'd have any pointer here, but I'm struggling to mock the return of an async function, while also preserving the `BackgroundTask` that is added to the response here- https://github.com/ethyca/fidesops/pull/622/commits/53f414e53cfa4396a8d27cc68bccd0da1268809d#diff-51868caf44d288309a19f7b59360f148adfa01226f70bf528441cef723439ffcR57

pattisdr: have you tried AsyncMocks? https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.AsyncMock

eastandwestwind: checking this out!

pattisdr:
thanks @eastandwestwind, I'll review this AM

src/fidesops/main.py
```diff
@@ -40,6 +42,65 @@
...
```
pattisdr, line 47: Can you add a descriptive docstring to this `dispatch_log_request` middleware?

src/fidesops/main.py
```diff
@@ -40,6 +42,65 @@
...
+        raise
+
+
+def prepare_and_log_request(
+    endpoint: str,
+    status_code: int,
+    event_created_at: datetime,
+    fides_source: Optional[str],
+    error_class: Optional[str],
+) -> None:
```
pattisdr, line 84: Similarly, could you add a good docstring to this function?

src/fidesops/main.py
pattisdr: fixture looks good @eastandwestwind 👍

pattisdr:
@eastandwestwind what about unit tests of this format? just testing that `prepare_and_log_request` was called with the correct parameters. I'm just arbitrarily adding them to the patch connections api endpoint tests:

```python3
class TestPatchConnections:
...
    @mock.patch('fidesops.main.prepare_and_log_request')
    def test_patch_connections_incorrect_scope_analytics(
        self, mocked_prepare_and_log_request, api_client: TestClient, generate_auth_header, payload
    ) -> None:
        url = V1_URL_PREFIX + CONNECTIONS
        auth_header = generate_auth_header(scopes=[STORAGE_DELETE])
        response = api_client.patch(url, headers=auth_header, json=payload)
        assert 403 == response.status_code
        assert mocked_prepare_and_log_request.called
        call_args = mocked_prepare_and_log_request._mock_call_args[0]

        assert call_args[0] == 'PATCH: http://testserver/api/v1/connection'
        assert call_args[1] == 403
        assert isinstance(call_args[2], datetime)
        assert call_args[3] is None
        assert call_args[4] == 'HTTPException'

    @mock.patch('fidesops.main.prepare_and_log_request')
    def test_patch_http_connection_successful_analytics(
        self, mocked_prepare_and_log_request, api_client, db: Session, generate_auth_header, url
    ):
        auth_header = generate_auth_header(scopes=[CONNECTION_CREATE_OR_UPDATE])
        payload = [
            {
                "name": "My Post-Execution Webhook",
                "key": "webhook_key",
                "connection_type": "https",
                "access": "read",
            }
        ]
        response = api_client.patch(url, headers=auth_header, json=payload)
        assert 200 == response.status_code
        body = json.loads(response.text)
        assert body["succeeded"][0]["connection_type"] == "https"
        http_config = ConnectionConfig.get_by(db, field="key", value="webhook_key")
        http_config.delete(db)

        call_args = mocked_prepare_and_log_request._mock_call_args[0]

        assert call_args[0] == 'PATCH: http://testserver/api/v1/connection'
        assert call_args[1] == 200
        assert isinstance(call_args[2], datetime)
        assert call_args[3] is None
        assert call_args[4] is None

```

pattisdr (approved):
nice work @eastandwestwind 

Let's also make sure that the landing page that @TheAndrewJackson  is working on is adding this `X-Fides-Source` to the request headers there too

PR #636, opened 2022-06-10
Fixing connector integration tests

galvana:
# Purpose
There were a few connector integration tests failing on main. Some were resolved by re-creating the necessary test data (Stripe, Segment) but we still need access to a premium account to fully test Sentry.

The long term fix is to include the data creation for the access requests in the fixtures, but that will be covered in a separate ticket. The goal of this ticket is to get the tests passing again in main.

# Changes
- Disabled Sentry task tests

# Checklist

### DIFF
--- a/tests/integration_tests/saas/test_sentry_task.py
+++ b/tests/integration_tests/saas/test_sentry_task.py
@@ -14,6 +14,7 @@ from fidesops.task.graph_task import get_cached_data_for_erasures
 from tests.graph.graph_test_util import assert_rows_match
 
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/saas/test_sentry_task.py
```diff
@@ -14,6 +14,7 @@
 from tests.graph.graph_test_util import assert_rows_match
 
 
+@pytest.mark.skip(reason="Pending account resolution")
```
pattisdr, line 17: nice @galvana

pattisdr (approved):

PR #654, opened 2022-06-14
Add Default Policies [#624]

### LINKED ISSUE #624: Migration to add default access + erasure policies to Fidesops
Right now a freshly deployed Fidesops instance has no way to process privacy requests until a Policy is configured. We should ship two default Policies with the keys `download` and `delete` to correspond with [the Privacy Centre's expectations](https://github.com/ethyca/fidesops/blob/main/clients/privacy-center/config/config.json#L7-L30). `Policy.drp_action` should also be set appropriately for both policies.

We can create both of these policies as a migration to be run when FIdesops starts up.

Aha! Link: https://ethyca.aha.io/features/OPS-33

pattisdr:
❗ Contains data migration; check downrev before merging
# Purpose
Ship Fidesops with two default Policies: `download` and `delete` so Fidesops instances can run privacy requests right away.  
The policies are created as a data migration.  

# Changes
Adds a data migration that:
- Creates a local StorageConfig
- Creates a Client with a specific fides_key
- If a `download` policy does not exist, create one, with a Rule using the localstorage and the the RuleTarget looking at identifiable data, with the autogenerated client above
- If a `delete` policy does not exist, create one, with a Rule that using a string masking strategy, targeting identifiable data, with the autogenerated client above
- Reverse migration only removes the download/delete policies and their associated resources if they were created by the autogenerated client.

Note: this does not create resources in test mode (pytest)

# Checklist

# Ticket

Fixes #624


src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -0,0 +1,213 @@
...
+
+    RuleTarget.create(
+        db=db,
+        data={
+            "name": "Fidesops Autogenerated Access Target",
+            "data_category": DataCategory("user.provided.identifiable").value,
+            "rule_id": access_rule.id,
+            "client_id": client.id,
+        },
+    )
```
pattisdr, line 173: I tried to use sane defaults for these, local storage which we already discussed, and targeting identifiable data, happy to adjust.

seanpreston: This is great, thanks @pattisdr

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -0,0 +1,215 @@
...
+
+    erasure_rule = Rule.create(
+        db,
+        data={
+            "name": "Fidesops Autogenerated Erasure Rule",
+            "action_type": ActionType.erasure.value,
+            "policy_id": erasure_policy.id,
+            "client_id": client.id,
+            "masking_strategy": {
+                "strategy": "null_rewrite",
```
seanpreston, line 201: Just thinking `null_rewrite` might not be the best strategy as it relies on fields being nullable. Let's use `string_rewrite`

pattisdr: good point, updated -

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -0,0 +1,384 @@
...
+
+storage_select_query: TextClause = text(
+    """SELECT storageconfig.id FROM storageconfig WHERE storageconfig.key = :storage_config_key"""
+)
+storage_insert_query: TextClause = text(
+    """
+    INSERT INTO storageconfig(id, name, type, details, key, secrets, format) 
+    VALUES(:storage_config_id, :storage_name, :storage_type, '{"naming": "request_id"}', 
+    :storage_key, NULL, 'json')
+    """
```
pattisdr, line 75: An earlier commit was using the SQLAlchemy ORM to run these queries, but this will break in the future as these models are updated, and we're importing the most recent model into older migrations where the columns don't exist yet.  These migrations should be able to stand on their own.

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+def upgrade() -> None:
+    """Data migration only.
+
+    Create an autogenerated client and storage destination, then use those to create
+    autogenerated 'download' and 'delete' policies if they don't already exist."""
+    if config.is_test_mode:
+        logger.info(f"Skipping data migration in test mode'")
+        return
```
pattisdr, line 130: It seemed like we shouldn't be adding these to the test database, so I run the migration, but don't insert data.

eastandwestwind (commented):
Couple questions related to docs:

1. Should we add to docs somewhere that these Policies are already created by default, to avoid confusion? 

Some places in docs where this might make sense- https://ethyca.github.io/fidesops/postman/using_postman/

https://ethyca.github.io/fidesops/guides/policies/

2. test mode / dev mode do diff things and can be confusing to someone first looking at config vars. I think it would be useful to better clarify in docs what setting these vars actually do, perhaps in a separate PR.

pattisdr:
Thanks @eastandwestwind I'll look into clarifying these items in the docs

pattisdr:
@ethyca/docs-authors minor docs added to describe the two autogenerated policies that ship with Fidesops and clarify that the TESTING variable is set by Pytest for when we run unit tests.

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
seanpreston: This is an interesting pattern to set. I could see us wanting to test these policies specifically at some stage, we can address this then.

docs/fidesops/docs/guides/configuration_reference.md
```diff
@@ -117,6 +117,8 @@ Please note: The configuration is case-sensitive, so the variables must be speci
 | `FIDESOPS__DEV_MODE` | False | If "True", the fidesops server will log error tracebacks, and log details of third party requests. This variable should always be set to "False" in production systems.|
 | `FIDESOPS__CONFIG_PATH` | None | If this variable is set to a path, that path will be used to load .toml files first. That is, any .toml files on this path will override any installed .toml files. |
 | `FIDESOPS__DATABASE__SQLALCHEMY_DATABASE_URI` | None | An optional override for the URI used for the database connection. |
+| `TESTING` | False | This variable does not need to be set - Pytest will set it to True when running unit tests, so we run against the test database. |
```
seanpreston, line 120: I intentionally left this out of the docs originally as it's more of an implementation detail. I can see it's good for transparency to show it, might it confuse some users too?

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
seanpreston: Have you tested that this is the case explicitly? As far as I'm aware this wouldn't be an issue since the migration is run at a specific point in the migration history, meaning we can guarantee the DB state at that point based on what's run before it.

docs/fidesops/docs/guides/configuration_reference.md
pattisdr: Your call, I can go either way.  Most users probably don't need this, but people digging in the code might wonder what test mode is.

conceptualshark: I'm okay with leaving this in for now, as I intend to refactor this page at some point, but I do think generally that anything targeted for testing or development should ideally live in the Development section - we just don't have a good spot for it yet.

docs/fidesops/docs/guides/policies.md
```diff
@@ -140,3 +148,12 @@ It's illegal to erase the same data twice within a Policy, so you should take ca
...
 
+## Default Policies
+
+Fidesops ships with two default Policies: `download` (for access requests) and `delete` (for erasure requests).  
+The `download` Policy is configured to retrieve `user.provided.identifiable` data and upload to a local storage location.
+The `delete` Policy is set up to mask `user.provided.identifiable` data with the string: `MASKED`.  
+
+You will want to create separate Policies that upload to a proper storage destination that target the fields you care about,
+and handle the fields in the appropriate way.  But these autogenerated Policies can get you up and running
+in a test environment.
```
conceptualshark, line 159:
```suggestion
These autogenerated Policies are intended for use in a test environment. In production deployments, you should configure separate Policies with proper storage destinations that target and process the appropriate fields.
```

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
seanpreston: > and then you have the latest version of the SQLAlchemy models trying to do things in this older migration that isn't possible yet.

Got it, that makes sense!

seanpreston (approved):

PR #661, opened 2022-06-15
Adding support for default values for connector params

### LINKED ISSUE #642: [SaaS Connectors] Support default values for connector params
A SaaS config should be able to define default values for specific connector param values. This can include a default `domain`, or a default `page_size` for pagination.

- Add a new `default_value` field for `connector_params`
- Update the `SaaSSchema` validator class to not require `connector_params` with default values
- Initialize empty connection_config secrets with default_values if available
- Do not override existing secrets with default values when updating the SaaS config

Aha! Link: https://ethyca.aha.io/features/OPS-28

galvana:
# Purpose
Adding support for default values for connector params. Connector params with default_values defined will not be required during the initial secrets configuration. Additionally, the default_value will be used to initialize the value for the secret. This initialization will only happen if the secret value is empty, the secret will not be overwritten if the user has already defined a value.

# Changes
- Added a new default_value field for connector_params
- Updated the SaaSSchema validator class to not require connector_params with default values
- Initialize empty connection_config secrets with default_values if available
- Misc cleanup for SaaS configs, any noteworthy changes will be called out in the PR comments

# Checklist

# Ticket

Fixes #642

### DIFF
--- a/src/fidesops/schemas/saas/saas_config.py
+++ b/src/fidesops/schemas/saas/saas_config.py
@@ -194,16 +194,6 @@ class SaaSConfig(BaseModel):
     test_request: SaaSRequest
     data_protection_request: Optional[SaaSRequest] = None  # GDPR Delete
 
-    @validator("connector_params")
-    def validate_connector_params(cls, v: List[ConnectorParam]) -> List[ConnectorParam]:
-        """Validate that the domain is present in the connector params."""
-
-        if not [param.name for param in v if param.name == "domain"]:
-            raise ValueError(
-                "SaaS config is missing a connector_param with name 'domain'"
-            )
-        return v
-
     @property
     def top_level_endpoint_dict(self) -> Dict[str, Endpoint]:
         """Returns a map of endpoint names mapped to Endpoints"""
--- a/tests/schemas/connection_configuration/test_connection_secrets_saas.py
+++ b/tests/schemas/connection_configuration/test_connection_secrets_saas.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict
+
 import pytest
 from pydantic import ValidationError
 
@@ -11,7 +13,7 @@ from fidesops.schemas.saas.saas_config import SaaSConfig
 @pytest.mark.unit_saas
 class TestSaaSConnectionSecrets:
     @pytest.fixture(scope="function")
-    def saas_config(self, saas_example_config) -> SaaSConfig:
+    def saas_config(self, saas_example_config: Dict[str, Any]) -> SaaSConfig:
         return SaaSConfig(**saas_example_config)
 
     def test_get_saas_schema(self, saas_config):
@@ -23,12 +25,14 @@ class TestSaaSConnectionSecrets:
         assert schema.__name__ == f"{saas_config.fides_key}_schema"
         assert issubclass(schema.__base__, SaaSSchema)
 
-    def test_validation(self, saas_config, saas_example_secrets):
+    def test_validation(
+        self, saas_config: SaaSConfig, saas_example_secrets: Dict[str, Any]
+    ):
         schema = SaaSSchemaFactory(saas_config).get_saas_schema()
         config = saas_example_secrets
         schema.parse_obj(config)
 
-    def test_missing_fields(self, saas_config):
+    def test_missing_fields(self, saas_config: SaaSConfig):
         schema = SaaSSchemaFactory(saas_config).get_saas_schema()
         config = {"domain": "domain", "username": "username"}
         with pytest.raises(ValidationError) as exc:
@@ -43,7 +47,9 @@ class TestSaaSConnectionSecrets:
             f"[{', '.join(required_fields)}]." in str(exc.value)
         )
 
-    def test_extra_fields(self, saas_config, saas_example_secrets):
+    def test_extra_fields(
+        self, saas_config: SaaSConfig, saas_example_secrets: Dict[str, Any]
+    ):
         schema = SaaSSchemaFactory(saas_config).get_saas_schema()
         config = {
             **saas_example_secrets,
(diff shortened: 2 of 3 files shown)


data/saas/config/mailchimp_config.yml
```diff
@@ -21,70 +21,73 @@ saas_config:
   test_request:
     method: GET
     path: /3.0/lists
-    
+
   endpoints:
-  - name: messages
```
galvana, line 26: Auto-formatting applied to SaaS configs.

src/fidesops/schemas/connection_configuration/connection_secrets_saas.py
```diff
@@ -49,10 +51,13 @@ def __init__(self, saas_config: SaaSConfig):
     # Pydantic uses the shorthand of (str, ...) to denote a required field of type str
     def get_saas_schema(self) -> Type[SaaSSchema]:
         """Returns the schema for the current configuration"""
-        field_definitions: Dict[str, Any] = {
-            connector_param.name: (str, ...)
-            for connector_param in self.saas_config.connector_params
-        }
+        field_definitions: Dict[str, Any] = {}
```
galvana, line 54: The `field_definitions` are used to denote a parameter as optional or required. A parameter with a default value will be flagged as required and a parameter using Pydantic's `(str, ...)` notation will be flagged as required.

As stated in the comments
```
Pydantic uses the shorthand of (str, ...) to denote a required field of type str
```

tests/service/connectors/test_queryconfig.py
```diff
@@ -70,45 +70,57 @@ def found_query_keys(node: TraversalNode, values: Dict[str, Any]) -> Set[str]:
         }
 
         # values exist for all query keys
-        assert found_query_keys(
```
galvana, line 73: No changes here but I got tired of having to omit this auto-format from my commits.

adamsachs (commented):
left some minor comments - most are to improve my understanding, probably just a few tweaks needed

tests/schemas/connection_configuration/test_connection_secrets_saas.py
```diff
@@ -48,3 +52,9 @@ def test_extra_fields(self, saas_config, saas_example_secrets):
         with pytest.raises(ValidationError) as exc:
             schema.parse_obj(config)
         assert exc.value.errors()[0]["msg"] == "extra fields not permitted"
+
+    def test_default_value_fields(self, saas_config, saas_example_secrets):
+        schema = SaaSSchemaFactory(saas_config).get_saas_schema()
+        del saas_example_secrets["domain"]
+        config = saas_example_secrets
+        schema.parse_obj(config)
```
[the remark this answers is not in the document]
galvana: You're right, we want to assert the initial state to make sure this passes for the right reasons.

data/saas/config/saas_example_config.yml
```diff
@@ -6,10 +6,11 @@ saas_config:
 
   connector_params:
     - name: domain
+      default_value: localhost
     - name: username
     - name: api_key
     - name: api_version
```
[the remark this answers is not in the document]
galvana: Docs updated!

tests/fixtures/saas_example_fixtures.py
```diff
@@ -202,7 +202,7 @@ def oauth2_connection_config(db: Session, oauth2_configuration) -> Generator:
         "name": "OAuth2 Connector",
         "description": "Generic OAuth2 connector for testing",
         "version": "0.0.1",
-        "connector_params": [{"name": item} for item in secrets.values()],
+        "connector_params": [{"name": item} for item in secrets.keys()],
```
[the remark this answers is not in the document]
galvana: You are correct, this was a mistake that wasn't caught during initial development. We were adding the secret values as the names instead of the keys.

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -193,6 +194,16 @@ class SaaSConfig(BaseModel):
     test_request: SaaSRequest
     data_protection_request: Optional[SaaSRequest] = None  # GDPR Delete
 
+    @validator("connector_params")
+    def validate_connector_params(cls, v: List[ConnectorParam]) -> List[ConnectorParam]:
```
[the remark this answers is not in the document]
galvana: I was including this since we were talking about potentially using the domain to identify a SaaS connector. I got ahead of myself and agree with you that we should give this more thought, rolling back for now.

src/fidesops/schemas/connection_configuration/connection_secrets_saas.py
galvana: This comes down to Pydantic's `create_model` function https://pydantic-docs.helpmanual.io/usage/models/#dynamic-model-creation. We need to create a dynamic model for secret validation based on the connector params. A sample `connector_params` to `field_definitions` conversion would look like this:

```
connector_params:
  - name: domain
    default_value: localhost
  - name: username
```
```
{
  "domain": "localhost"
  "username": (str, ...)
}
```
Then the `cls.__fields__` value on line 23 would be
```
{
  'domain': ModelField(name='domain', type=str, required=False, default='localhost'),
  'username': ModelField(name='username', type=str, required=True)
}
```
And we can use the `required` param to determine which secrets will be required by the user.

adamsachs: thanks for explaining this, it makes sense now.

i think one thing that tripped me up a bit was a typo in your original comment here - you wrote
> A parameter with a default value will be flagged as required

but instead it should read:
> A parameter with a default value will be flagged as optional

assuming your later comment here is correct. right?

adamsachs (approved):
approving -- all looks good now, just a minor question i put in for my own clarification.

src/fidesops/schemas/connection_configuration/connection_secrets_saas.py
galvana: You're right, I fixed my typo in case we come back to this.

conceptualshark (approved):

PR #675, opened 2022-06-16
Add ConnectionConfig Filters

### LINKED ISSUE #670: [Datastore Management] Adding filtering capabilities to landing page BACK END
Will need to add filtering capabilities for connection type, status (enabled, disabled) and may need to code for the joining of filters that are selected.

pattisdr:
# Purpose
Add ConnectionConfig filters to support the frontend datastore management landing page.

# Changes
- Filters added for disabled: `api/v1/connection/?disabled=true` or `api/v1/connection/?disabled=false`
- Filters added for multiple connectiontypes: `?api/v1/connection/?connection_type=postgres&connection_type=redshift
   - If there are multiple connection_type query params, the backend looks for *any* that match.
- Filters for test_status: `?test_status=passed`, `?test_status=failed`, and `?test_status=untested`
- Filters for system_type: `?system_type=saas` and `?system_type=database`

# Checklist

# Ticket

Fixes #670


src/fidesops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -86,9 +86,15 @@ def get_connections(
     db: Session = Depends(deps.get_db),
     params: Params = Depends(),
     search: Optional[str] = None,
+    disabled: Optional[bool] = None,
+    connection_type: Optional[List[str]] = Query(default=None),  # type:ignore
```
pattisdr, line 90: to search for multiple connection_types do: `?connection_type=postgres&connection_type=mongo`.  

I would have preferred this syntax: `?connection_type=postgres,mongo` for an OR query, but I am using the syntax in the docs https://fastapi.tiangolo.com/it/tutorial/query-params-str-validations/#query-parameter-list-multiple-values which works well with the autogenerated API docs:

Screen Shot 2022-06-16 at 6 33 51 PM

src/fidesops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -102,6 +108,13 @@ def get_connections(
                 ConnectionConfig.description.ilike(f"%{escape_like(search)}%"),
             )
         )
+
+    if connection_type:
+        query = query.filter(ConnectionConfig.connection_type.in_(connection_type))
+
+    if disabled is not None:
+        query = query.filter(ConnectionConfig.disabled == disabled)
+
```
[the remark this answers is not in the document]
TheAndrewJackson: It looks like there are 2 more filters that are needed for the frontend. 

One is `Testing Status`. This would be filtering based on the `last_test_succeeded` field. I think it needs to be able to filter based on `true`, `false`, and `null`. `null` would be for connections that have never been tested.

The second is `System Type`. In the mockups it looks it's for filtering for all SAAS connections or all Database connections. I'm not sure if this is possible with the current data model. I think it would conflict with the `connection_type` filter you added because it would be a second filter based on the same field. Does that seem correct?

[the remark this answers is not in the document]
pattisdr: Thanks @TheAndrewJackson - I can add both testing status and system type. it's fine if it's a second filter on the same field, just important to note that if the user selects "saas" and then filters for "postgres" then they'll get no results.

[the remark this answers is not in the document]
pattisdr: Re: Testing status, let's make sure we're being careful about the "null" status.  Usually a None would mean, "don't filter", and in this case, you're wanting it to mean "we never tested it".  Maybe we do string filters here, "passed", "failed", "untested", and None, or similar.

src/fidesops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -102,6 +116,30 @@ def get_connections(
...
+    if system_type:
+        if system_type == SystemType.saas:
+            query = query.filter(
+                ConnectionConfig.connection_type == ConnectionType.saas
+            )
+        elif system_type == SystemType.database:
+            query = query.filter(
+                ConnectionConfig.connection_type.notin_(
+                    [ConnectionType.saas, ConnectionType.manual]
```
pattisdr, line 139: Pointing out that there's really 3 system types right now, saas, database, and manual. I've only added system_type filtering for the two, but I can change.

src/fidesops/api/v1/endpoints/connection_endpoints.py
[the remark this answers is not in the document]
pattisdr: OK I added filtering on `?test_status=passed`, `?test_status=failed`, and`?test_status=untested`, and then there's
`?system_type=saas` and `?system_type=database`.  

If someone filters on conflicting fields, they'll get no results: `?system_type=saas&connection_type=postgres`.

src/fidesops/schemas/connection_configuration/connection_config.py
```diff
@@ -30,6 +31,26 @@ class Config:
         extra = Extra.forbid
 
 
+class TestStatus(Enum):
+    passed = "passed"
+    failed = "failed"
+    untested = "untested"
+
+    def str_to_bool(self) -> Optional[bool]:
```
TheAndrewJackson, line 39: I like this! Good idea

src/fidesops/api/v1/endpoints/connection_endpoints.py
TheAndrewJackson: @pattisdr Does that backend support manual connection types right now?

pattisdr: yes it does but there's not an easy way to interact with them yet, it is a connection type that requires a user to go manually perform some action and upload data back to fidesops.

TheAndrewJackson: I see. Maybe we should add it in in the backend just to get ahead of the future update? The frontend can always leave the filtering option out if it isn't supposed to handle it yet.

TheAndrewJackson (approved):

PR #717, opened 2022-06-26
Parallelize CI safe checks

sanders41:
# Purpose

Parallelize the safe CI checks to reduce run times.

Currently the safe CI checks take an average of around 13 minutes to run. I did some testing on a private fork and was getting an average run time of around 8:45 running the jobs in parallel, which is about a 33% reduction in time.

# Changes
- Modify the tests in `safe_pr_checks.yml` to run in parallel rather than serially.
- Run linting outside of docker. Doing this further reduces run times. Pylint is the longest running job in this section at about 2 minutes.

# Checklist

# Ticket

Fixes #

### DIFF
--- a/.github/workflows/safe_pr_checks.yml
+++ b/.github/workflows/safe_pr_checks.yml
@@ -20,7 +20,7 @@ jobs:
         pip install -U pip
         pip install -r dev-requirements.txt
         pip install -e .
-    - name: run iort
+    - name: run isort
       run: isort --check-only src tests
 
   Black:
--- a/.github/workflows/safe_pr_checks.yml
+++ b/.github/workflows/safe_pr_checks.yml
@@ -6,26 +6,88 @@ on:
   pull_request:
 
 jobs:
-  Check-All:
+  Import-Sorting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
+        cache: "pip"
+    - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


ThomasLaPiana:
@sanders41 another solution is to copy how we do it over on `fidesctl`, but this does definitely have an immediate positive impact as-is 🙂

sanders41:
@ThomasLaPiana I tried copying it but was getting errors, I think because of the way the tag is set, `IMAGE_TAG := $(shell git fetch --force --tags && git describe --tags --dirty --always)`.

Because of this I tried the way I have it now, and to my surprise it was significantly faster than building the image once and sharing it. The reason is the upload/down load of the image to storage is slow, and each job could build it faster itself rather than waiting on it to be built, uploaded, then downloading it. Not the result I was expecting, but I couldn't argue with the numbers. With the pre-built image the first job would fail at around 13 minutes. When done the way it is here all tests are done in around 8 minutes 45 seconds.

PSalant726 (approved):
Have you tried breaking out the python setup and dependency installation into their own job, and making that job a dependency of the others that require them? That seems like a nice way to DRY things up and further reduce run times.

Otherwise, this LGTM.

sanders41:
@PSalant726 I am already doing this everywhere that doesn't use a database. I considered also doing this everywhere, i.e. only running external services in Docker (i.e. postgres, redis, etc) and not the app itself. It would make things much faster, but I don't know how everyone feels about doing this. Thoughts?

seanpreston:
Have merged this now to reap the benefits, we can add only running external services in Docker as a separate incremenn

PR #720, opened 2022-06-27
Serve UI from root

### LINKED ISSUE #666: [Spike] Serve Admin UI from `/` instead of `/static/index.html`
Timebox - 1 day
Right now when you serve admin UI it serves on something that isn't ideal. It's not conventional to access UIs from a static directory.
Ideally we would serve it from the root and not a static html.

TheAndrewJackson:
# Purpose
Update webserver to serve the admin-ui from the root (`/`) instead of `/static/index.html`

# Changes
- Add new `/` and `/{catchall:path}` routes that handle serving all of the UI files
- Remove `BASE_ASSET_URN` constant from the frontend because it's no longer needed
- Remove `/static` mounted directory

# Checklist

# Ticket

Fixes #666


src/fidesops/main.py
```diff
@@ -115,20 +116,27 @@ def prepare_and_log_request(
 for handler in ExceptionHandlers.get_handlers():
     app.add_exception_handler(FunctionalityNotConfigured, handler)
 
+WEBAPP_DIRECTORY = Path("src/fidesops/build/static")
+WEBAPP_INDEX = WEBAPP_DIRECTORY / "index.html"
 
-@app.on_event("startup")
-async def create_webapp_dir_if_not_exists() -> None:
```
allisonking, line 120: This function may still be useful to have. It's mostly for the case of someone is starting up the server for the first time and they've never built the frontend before, so don't have `build/static/index.html` on their local filesystem. on the fidesctl side, I added some lines to the Dockerfile to create these files, but even that doesn't work all the time since we frequently mount our own local directory over the built image.  

it's always annoying to test this, but these are the two things I double checked for while working on this on the fidesctl side:

* delete your local `build/static` directory, then navigate to localhost:8080 and make sure it doesn't error out (I _think_ without these lines it would, but am not positive, maybe it's more graceful than that!)
* build and run the docker image without volume mounting to make sure that is getting the most up to date static files properly hosted at localhost:8080

if those two are working 💯  then this is probably good to go!

pattisdr: confirmed deleting `build/static` and navigating to localhost:8080 causes a 500 error

TheAndrewJackson: Good find. I'll add some code that handles that situation now

src/fidesops/util/api_router.py
```diff
@@ -0,0 +1,35 @@
...
+        alternate_path = path + "/"
+        add_alternate_path = super().api_route(
+            alternate_path, include_in_schema=False, **kwargs
+        )
+
+        def decorator(func: DecoratedCallable) -> DecoratedCallable:
+            add_alternate_path(func)
+            return add_path(func)
+
+        return decorator
```
pattisdr, line 35: Interesting solution, to handle trailing slashes. Main thing, we talked about a test for this one, otherwise, it's looking good

seanpreston: Would echo what Dawn has said there. It's nice for the server to behave the same regardless of a trailing slash. Django for instance solves this with regex in the URL def where you can say something like:
```
    re_path(r'^api/v1/some-endpoint/?', comments),
```
where the last `?` makes the preceding `/` optional. Since we're effectively doing this automatically now, it would be good to also document that somewhere in the API docs.

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ The types of changes are:
 
 ## [1.6.1](https://github.com/ethyca/fidesops/compare/1.6.0...1.6.1)
 
+### Changed
+* Update admin ui to be served from the root route `/` [#720](https://github.com/ethyca/fidesops/pull/720)
+
```
pattisdr, line 55: Small merging issue, move this up to "Unreleased"

pattisdr (approved):
Good work @TheAndrewJackson, since we need a slight tweak to the changelog, do you mind documenting that the endpoints work with both a trailing slash and non trailing slash per Sean's request?

CHANGELOG.md
pattisdr: thank you Andrew, that looks fine!

PR #808, opened 2022-07-06
update celery config defaults, set to redis settings if none provided

### LINKED ISSUE #771: Make Celery connection defaults more sane
We've had some feedback that for anyone deploying Fidesops with Celery the config vars have confusing defaults (because they won't actually work outside of the Docker Compose environments). We should update these to something that fits better for deployed use cases. e.g. forcing them as the Redis connection string that's already defined and explicitly incrementing the index.

seanpreston:
# Purpose

Changes the default Celery implementation settings to default to the Redis cache.

# Changes
- Adds a `redis_connection_url` that can be manually specified
- Set `result_backend` to `redis_connection_url` if no `celery_result_backend` is supplied
- Set `broker_url` to `redis_connection_url` if no `celery_broker_url` is supplied

# Checklist

# Ticket

Fixes #771

### DIFF
--- a/src/fidesops/core/config.py
+++ b/src/fidesops/core/config.py
@@ -39,8 +39,8 @@ class ExecutionSettings(FidesSettings):
     TASK_RETRY_BACKOFF: int
     REQUIRE_MANUAL_REQUEST_APPROVAL: bool = False
     MASKING_STRICT: bool = True
-    CELERY_BROKER_URL: str = "redis://:testpassword@redis:6379/1"
-    CELERY_RESULT_BACKEND: str = "redis://:testpassword@redis:6379/1"
+    CELERY_BROKER_URL: Optional[str] = None
+    CELERY_RESULT_BACKEND: Optional[str] = None
 
     class Config:
         env_prefix = "FIDESOPS__EXECUTION__"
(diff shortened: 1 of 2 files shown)


src/fidesops/core/config.py
```diff
@@ -59,6 +59,26 @@ class RedisSettings(FidesSettings):
...
+    @classmethod
+    def assemble_redis_connection_url(
+        cls,
+        v: Optional[str],
+        values: Dict[str, str],
+    ) -> str:
+        """Join Redis connection credentials into a connection string"""
+        if isinstance(v, str):
+            # If the whole URL is provided via the config, preference that
+            return v
```
sanders41, line 74: Does this need further validation to make sure it is actually a URL? Currently any string would be allowed, but wouldn't actually work.

src/fidesops/core/config.py
```diff
@@ -59,6 +59,26 @@ class RedisSettings(FidesSettings):
...
+
+        return "redis://:{password}@{host}:{port}/{db_index}".format(
+            password=values["PASSWORD"],
+            host=values["HOST"],
+            port=values["PORT"],
+            db_index=values.get("DB_INDEX"),
+        )
```
sanders41, line 81: f-strings are slightly faster. Also `DB_INDEX` is a required field so you don't need the `.get` since you know it will have a value.

```suggestion
        return f"redis://:{values['PASSWORD']}@{values['HOST']}:{values['PORT']}/{values['DB_INDEX']}"
```

seanpreston: Have opted to make `DB_INDEX` optional, as the connection string will still work without it

src/fidesops/core/config.py
seanpreston: I thought about this — ideally yes it would. I'll need to research the range of connection URIs across vanilla Redis, Elasticache, and whatever other equivalents exist, in order to add anything meaningful though (without resorting to generic [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) validation). In case we want to pursue that I did add a `USER` to match Redis' own implementation ([even though it's unused](https://stackoverflow.com/a/44353927)).

src/fidesops/core/config.py
```diff
@@ -73,12 +74,7 @@ def assemble_connection_url(
             # If the whole URL is provided via the config, preference that
             return v
 
-        return "redis://:{password}@{host}:{port}/{db_index}".format(
-            password=values["PASSWORD"],
-            host=values["HOST"],
-            port=values["PORT"],
-            db_index=values.get("DB_INDEX"),
-        )
+        return f"redis://{values.get('USER', '')}:{values['PASSWORD']}@{values['HOST']}:{values['PORT']}/{values.get('DB_INDEX')}"
```
sanders41, line 77: If `DB_INDEX` is `NONE` the string will be `redis://{some url}/None` so maybe default to `''` if `DB_INDEX` is `None`.

```suggestion
        return f"redis://{values.get('USER', '')}:{values['PASSWORD']}@{values['HOST']}:{values['PORT']}/{values.get('DB_INDEX', '')}"
```

seanpreston: Good spot, thanks!

PR #835, opened 2022-07-08
adds concurrency to unsafe check jobs

seanpreston:
# Purpose

This PR adds concurrency to our unsafe PR checks, similar to the change we made to our safe checks


sanders41:
@seanpreston I re-ran this test and it failed in the same place, but I don't see anything you changed that would make that happen. Could the SalesForce connection just be down?

PR #844, opened 2022-07-08
Populate dataset

galvana:
# Purpose
Adds a `update_dataset` function which takes a `ConnectionConfig`, `DatasetConfig`, and API data to update the existing Dataset. This is meant to be used for SaaS connector development.

# Functionality
- Updates an existing dataset based on the API data
  - Can also write the generated dataset to a separate file from the existing dataset for comparisons
- The collection order is derived from the collection order in the SaaS config
  - Default behavior preserves the collection order of the existing dataset
- Preserves any user-defined data categories during the dataset generation
- Invoked from the `test_*_access_request_task` after the `run_access_request` has returned a response
  ```
  v = graph_task.run_access_request(
      privacy_request,
      policy,
      graph,
      [connection_config],
      {"email": identity_email},
  )

  update_dataset(
      connection_config,
      dataset_config,
      v,
      "generated_dataset.yml",
  )
  ```

# Checklist


tests/test_helpers/dataset_utils.py
```diff
@@ -0,0 +1,258 @@
...
+            {
+                "name": collection_name,
+                "fields": fields,
+            }
+        )
+
+    return collections
+
+
+def generate_fields(
```
galvana, line 137: We have a `merge_datasets` function but we need to preserve the order of the fields and we lose control of that as soon as we convert a simple dict to a `FidesopsDataset`, the fields are alphabetized due to default Pydantic functionality.

tests/test_helpers/test_dataset_utils.py
```diff
@@ -0,0 +1,667 @@
...
+                        {
+                            "name": "b",
+                            "data_categories": ["user.provided"],
+                            "fidesops_meta": {"data_type": "string"},
+                        }
+                    ],
+                },
+            ],
+        }
+        api_data = {}
```
adamsachs, line 577: would a better test here include an `api_data` object that has at least one of the existing collection elements placed in a different order than in `existing_dataset`?

tests/test_helpers/test_dataset_utils.py
```diff
@@ -0,0 +1,667 @@
+from tests.test_helpers.dataset_utils import generate_collections, generate_dataset
+
+
```
adamsachs, line 3: nice job with these tests - they're a great starting point for understanding the expected functionality of the utilities.

one follow up - would it be feasible/useful to include more of an "integration" level test that actually tests and demonstrates how the utility can work on a _real_ API response, i.e. using a saas provider we already integrate with as an example? that test would obviously need to be marked as an `integration_saas` test since it would require external communication, but just seems like it could be a nice-to-have. (if you think this is feasible, maybe we can split out a separate issue/PR for it, since i don't think it should necessarily block this PR)

tests/test_helpers/dataset_utils.py
```diff
@@ -0,0 +1,258 @@
...
+    if not collection_order:
+        collection_order = [
+            collection["name"] for collection in existing_dataset["collections"]
+        ]
+
+    # remove the dataset name from the keys in the api_data map before passing
+    # into generate_collections
+    generated_collections = generate_collections(
+        {
+            collection_name.replace(f"{existing_dataset['fides_key']}:", ""): collection
```
adamsachs, line 76: is this logic/processing covered in the unit tests? i know we're probably not going to be able to cover every piece of this, but it's some behavior that i think would be good to have covered both to ensure correct functionality and to clarify expected behavior

adamsachs: also, i think my suggestion about an "integration" level test may help in providing coverage for this piece of things? that would address my concern, no need to make it specifically a unit test.

maybe just another reason to look into having that sort of "integration" test with more realistic data

tests/test_helpers/dataset_utils.py
```diff
@@ -0,0 +1,258 @@
...
+
+def generate_collections(
+    api_data: Dict[str, List[Row]], dataset: Dict[str, Any] = None
+) -> List[Dict[str, Any]]:
+    """
+    Generates a list of collections based on the response data or returns
+    the existing collections if no API data is available.
+    """
+
+    # convert FidesopsDataset to Dataset to be able to use the Collection helpers
```
adamsachs, line 107: just noting that i don't totally follow this, but probably due to my lack of understanding, so may be good to review together offline

tests/test_helpers/dataset_utils.py
```diff
@@ -0,0 +1,258 @@
...
+            object["data_categories"] = field.data_categories
+        if field.data_type() != "None":
+            object["fidesops_meta"] = {"data_type": field.data_type()}
+        if isinstance(field, ObjectField) and field.fields:
+            object["fields"] = get_simple_fields(field.fields.values())
+        object_list.append(object)
+    return object_list
+
+
+def get_data_type(value) -> Optional[str]:
```
adamsachs, line 227: to me, it feels like this function should live more centrally. what would you think of defining it somewhere within `fidesops.graph.data_type.py`? from what i can tell, there's nothing yet defined there that takes an arbitrary input and gives the corresponding proper `Fidesops` `DataType`, but I think that's what you want here.

If you add that utility method over there, then I think here you should be relying upon the proper `DataType` enum - passing around string literals here just feels a bit sloppy to me. Specifically, I think it's ripe for unnoticed regressions if/when any of the typing logic changes in the fidesops "taxonomy" (for lack of a better term), and it will be burdensome to maintain.

tests/test_helpers/dataset_utils.py
```diff
@@ -0,0 +1,258 @@
...
+            collection.name: collection for collection in graph.collections
+        }
+
+    collections = []
+    for collection_name in set().union(
+        api_data.keys(),
+        collection_map.keys(),
+    ):
+        if len(rows := api_data.get(collection_name, [])):
+            fields = generate_fields(rows[0], collection_name, collection_map)
```
adamsachs, line 123: not necessarily a feature for now, but a potential enhancement could be to allow for a configurable "sampling size" here so that potentially more than just the first row of response data is being used here. obviously we'd then need to build in some more advanced type-inference logic to take into account multiple values for each column, but just calling it out as an area for fine tuning.

adamsachs:
it looks like there are some unit tests failing that i _think_ are real failures? once that's addressed, i think i'm good to approve - there is one additional nitpick i threw on there for the `get_data_type`, let me know what you think. i don't think it should block the PR but if you agree with it and it's a quick change, may as well get it in there.

adamsachs (approved):
nice, this all looks great - thanks for making those updates to move the data type logic more centrally, i think it will help for maintainability. 

excited to see this in action!

PR #860, opened 2022-07-12
Return persisted identities in `get_request_status` view

seanpreston:
# Purpose

This PR updates the way we return identities to use the data stored in the database, rather than the cache.

# Changes
- Return identity date from `ProvidedIdentity` table
- Persist identity data for privacy requests created by test fixtures

# Checklist

# Ticket
Fixes NA


src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -491,7 +491,7 @@ def get_request_status(
         # Conditionally include the cached identity data in the response if
         # it is explicitly requested
         for item in paginated.items:  # type: ignore
-            item.identity = item.get_cached_identity_data()
+            item.identity = item.get_persisted_identity().dict()
```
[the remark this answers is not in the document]
pattisdr: Creating the request body for a webhook, creating the requests for saas configs retrieve/update statements, and feeding the initial seed data into the traversal all still use the cache, not the database.

Do we do this because it's easier to access the cache sometimes, we don't always have a readily available session? especially in the traversal? I'm a little worried about having different locations storing what the identity is, some pull from one, others pull from another.

tests/fixtures/application_fixtures.py
```diff
@@ -839,7 +848,12 @@ def succeeded_privacy_request(cache, db: Session, policy: Policy) -> PrivacyRequ
             "client_id": policy.client_id,
         },
     )
-    pr.cache_identity({"email": "email@example.com"})
+    identity_kwargs = {"email": "email@example.com"}
+    pr.cache_identity(identity_kwargs)
+    pr.persist_identity(
```
pattisdr, line 853: Bug: Policy webhooks can have derived_identities returned. Neither `PrivacyRequest.trigger_policy_webhook` nor `privacy_request_endpoints > resume_privacy_request` which both update the identity graph, persist the data to the database, they only add it to the redis cache.

pattisdr (commented):
Just some more cleanup needed @seanpreston

src/fidesops/api/v1/endpoints/privacy_request_endpoints.py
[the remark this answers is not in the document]
seanpreston: > Do we do this because it's easier to access the cache sometimes, we don't always have a readily available session?

For now, yes. Ideally I'd like everything to use the same source of truth for identity data, but that's a larger refactor for exactly this reason, the DB connection isn't piped into everywhere that would need it yet. I've made [this ticket](https://github.com/ethyca/fidesops/issues/867) to be actioned as a follow-up.

tests/fixtures/application_fixtures.py
seanpreston: > Would it be useful to have a method that both persists the identity in the cache and in the database at the same time?

I'm torn here. On the one hand it's nice to have consistency, on the other, as you rightly suggest above, it means we'll need to be plumbing the DB connection in more places. I'm not sure if it's better to have the execution update the cache with identity data at the very start before the traversal, such that we can guarantee the traversal will always use what was provided by the user on privacy request creation. That way the internals can still use the cache and benefit from the speed, and less refactoring is required.

seanpreston: > Bug: Policy webhooks can have derived_identities returned. Neither PrivacyRequest.trigger_policy_webhook nor privacy_request_endpoints > resume_privacy_request which both update the identity graph, persist the data to the database, they only add it to the redis cache.

We should these concerns for now. The `ProvidedIdentity` is useful for facilitating request search based on the exact identity provided, we don't currently want to search based on derived identities, or return them into the UI, so should be careful when we update the `ProvidedIdentity` table as that's what will get displayed in the UI (and doesn't currently support anything beyond `email` and `phone_number`).

pattisdr: >  That way the internals can still use the cache and benefit from the speed, and less refactoring is required. What do you think?

Thinking about this more, I agree, it's in line with our original design, in that we query everything up front, build the graph, and execute it. We're not regularly querying the database as we execute the traversal which I think is good for performance.

pattisdr: > We should separate these concerns for now. The ProvidedIdentity is useful for facilitating request search based on the exact identity provided, we don't currently want to search based on derived identities

OK, that makes sense

pattisdr:
@seanpreston thanks for your response to my comments. This all makes sense, the one thing I would add then, are code comments in the places where we're writing to just the cache and not the database.  I'd want to note why we're doing this and make it clear it's intentional.

CHANGELOG.md
```diff
@@ -59,6 +59,14 @@ The types of changes are:
 * Bump fideslib to handle base64 encoded password [#820](https://github.com/ethyca/fidesops/pull/820)
 * Stop masking uvicorn logs by default [#831](https://github.com/ethyca/fidesops/pull/831)
 
+## Changed
+* Changed wording on Admin UI login page [#774](https://github.com/ethyca/fidesops/pull/774)
+* Fixed typos in Admin UI [#774](https://github.com/ethyca/fidesops/pull/774)
+* Update clipboard icon in Admin UI [#838](https://github.com/ethyca/fidesops/pull/838)
+* Return identity data from application DB, instead of cache [#860](https://github.com/ethyca/fidesops/pull/860)
+* Update admin ui to be served from the root route `/` [#720](https://github.com/ethyca/fidesops/pull/720)
```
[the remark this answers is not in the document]
pattisdr: Duplicates most of "changed" section above

pattisdr (approved):

PR #888, opened 2022-07-15
[SaaS Connector] Salesforce (erasure)

### LINKED ISSUE #713: Salesforce (erasure)
### Description

Add erasure functionality for the following collections. Salesforce allows `PATCH` actions for these endpoints so we should be able to mask individual fields.

|Collection|Request|
|---|---|
|contacts|`PATCH` /services/data/v54.0/sobjects/Contact/<contact_id>|
|cases|`PATCH` /services/data/v54.0/sobjects/Case/<case_id>|
|leads|`PATCH` /services/data/v54.0/sobjects/Lead/<lead_id>|
|accounts|`PATCH` /services/data/v54.0/sobjects/Account/<account_id>|
|campaign_members|`PATCH` /services/data/v54.0/sobjects/CampaignMember/<campaign_member_id>|

galvana:
# Purpose
Adds erasure support for Salesforce.

# Changes
Added update requests for the following
Collection | Request
-- | --
contacts | PATCH /services/data/v54.0/sobjects/Contact/<contact_id>
cases | PATCH /services/data/v54.0/sobjects/Case/<case_id>
leads | PATCH /services/data/v54.0/sobjects/Lead/<lead_id>
accounts | PATCH /services/data/v54.0/sobjects/Account/<account_id>
campaign_members | PATCH /services/data/v54.0/sobjects/CampaignMember/<campaign_member_id>

I also made these core changes which were needed for Salesforce and are generic enough to apply to other connectors:
- The `read_only` metadata flag was not being applied to the `masked_object_fields` placeholder value only `all_object_fields`. Both of these are maps which contain masked data but we should omit read-only fields in both cases (refer to **placeholder options for updates** in the [SaaS Connector Documentation](https://ethyca.github.io/fidesops/saas_connectors/saas_config/#body))
- The `assign_placeholders` function would incorrectly return `None` when replacing a placeholder with an empty string
  - Added a test to cover this scenario 

Although we prefer keeping such changes separate, I think it's important for these changes to be included together so we can understand the reason the changes were made.

# Checklist

# Ticket

Fixes #713


tests/fixtures/saas/salesforce_fixtures.py
```diff
@@ -219,4 +220,17 @@ def salesforce_create_erasure_data(
...
+        url=f"{base_url}/services/data/v54.0/sobjects/Case/{case_id}", headers=headers
+    )
+    assert case_response.ok
+
+    account_response = requests.delete(
+        url=f"{base_url}/services/data/v54.0/sobjects/Account/{account_id}",
+        headers=headers,
+    )
+
+    assert account_response.ok
```
adamsachs, line 236: i assume it's not feasible to actually confirm with salesforce that the data was deleted? (i.e. it takes too much time to process the delete)

src/fidesops/service/connectors/saas_query_config.py
```diff
@@ -219,15 +224,17 @@ def generate_update_stmt(
...
         all_value_map: Dict[str, Any] = {}
         for field_path, field in self.field_map().items():
-            # only map scalar fields that are not read-only
-            if isinstance(field, ScalarField) and not field.read_only:
-                # only map if the value exists on the row
+            # only map scalar fields
+            if (
+                isinstance(field, ScalarField)
+                and pydash.get(row, field_path.string_path) is not None
+            ):
```
adamsachs, line 237: maybe i'm missing something, but isn't this check just repeated on the follow line? i can understand grouping them together in the same conditional, but i don't think we gain anything by repeating it, and the code becomes a bit less intuitive.

adamsachs, line 237: maybe i'm missing something, but isn't this check just repeated on the follow line? i can understand grouping them together in the same conditional, but i don't think we gain anything by repeating it, and the code becomes a bit less intuitive.

adamsachs:
some pretty minor comments that we should be able to clear up quickly. besides that, i think it just needs the `Run Unsafe PR Checks` label applied so that the right integration tests are run?

src/fidesops/service/connectors/saas_query_config.py
galvana: Nice catch, I was refactoring this to make it cleaner but I forgot to remove the inner conditional, fixed!

tests/integration_tests/saas/test_salesforce_task.py
```diff
@@ -359,3 +362,415 @@ def test_salesforce_access_request_task(
...
+    response = requests.get(
+        url=f"{base_url}/services/data/v54.0/sobjects/Lead/{lead_id}",
+        headers=headers,
+    )
+    lead = response.json()
+    assert lead["FirstName"] == "MASKED"
+    assert lead["LastName"] == "MASKED"
+
+    # reset
+    config.execution.MASKING_STRICT = True
```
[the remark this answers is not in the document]
galvana: Good suggestion, I changed this to store the `masking_strict` value and restore it at the end of the test.

src/fidesops/service/connectors/saas_query_config.py
adamsachs: i think GH was playing a joke on me - the UI glitched and had me add this comment twice. of course it was a comment that was pointing out redundancy in the code :D

adamsachs (approved):
all looks good to me, thanks for making the requested adjustments!

Issue #813, opened 2022-07-06
Fix MSSQL integration tests (Timebox: 3 days)


seanpreston:
The MSSQL tests are [inconsistently failing at the moment](https://github.com/ethyca/fidesops/runs/7218441274?check_suite_focus=true), and should be fixed.

Issue #439, opened 2022-04-28
[Enhancement] Convert project to use Nox 


TheAndrewJackson:
# Description
The Fidesops project uses `Make` to have a standard set of commands that are used for CI and development. It works well for a lot of tasks but can have issues when commands aren't Posix compliant or depend on directories to be named a certain way.

The solution to this issue is the tool [Nox](https://nox.thea.codes/en/stable/). The website describes it as:

> nox is a command-line tool that automates testing in multiple Python environments, similar to [tox](https://tox.readthedocs.org/). Unlike tox, Nox uses a standard Python file for configuration.

This will help standardize the development process across all environments. [FidesCtl](https://github.com/ethyca/fides) recently switched over to nox and it's worked really well for them.

Aha! Link: https://ethyca.aha.io/features/OPS-70

PR #890, opened 2022-07-18
Bump fideslang to 1.1.0

### LINKED ISSUE #757: Update to Fideslang 1.1.0
See the Fideslang 1.1.0 milestone.

ThomasLaPiana:
# Purpose

Bump fideslang to version 1.1.0

# Changes

- update `requirements.txt`
- remove `identifiable`, `nonidentifiable`, `derived` and `provided` sub levels (they all just become `user`
- fix tests

# Checklist
- [ ] ~This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging~

# Ticket

Fixes #757


ThomasLaPiana:
@conceptualshark would appreciate a look-see here as there might be some docs cleanup from the fideslang upgrade changes. Shouldn't be quite as much as on fidesctl but more than I trust myself to handle!

ThomasLaPiana:
@pattisdr it looks like you were the one who wrote the tests that are failing now, it looks like `customer_id` isn't getting pulled out and thats why the tests are failing. I'm having a hard time nailing down why my change broke this but I'm sure I'm just not seeing it, maybe something in the example dataset yaml file is no longer being denoted as dangerous?

conceptualshark (commented):
Docs look covered - thank you! 👍

ThomasLaPiana:
side note: I'm kind of surprised that this doesn't require a database migration given the additional fields for datasets but 🤷

docs/fidesops/docs/guides/query_execution.md
```diff
@@ -99,9 +99,9 @@ dataset:
           - name: address_id
 ```
 
-We trigger a retrieval with identity data, such as an email address or user ID, that's provided by the user. What we do is... 
+We trigger a retrieval with identity data, such as an email address or user ID, that's provided by the user. What we do is...
 
-1. Identify the collections that contain the identity data that the user provided.
+1. Identify the collections that contain the identity data that the user.
```
pattisdr, line 104: Unintentional find/replace

requirements.txt
```diff
@@ -9,7 +9,7 @@ emails
 fastapi-caching[redis]
 fastapi-pagination[sqlalchemy]~= 0.9.3
 fastapi[all]==0.78.0
-fideslang==1.0.0
+fideslang==1.1.0
```
pattisdr, line 12: 🎉 
For my own understanding, can you explain why we're moving away from the provided/derived and identifiable/non-identifiable categories?

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -331,7 +331,7 @@ def autogenerate_erasure_policy(connection: Connection, client_id: str) -> None:
             "target_id": rule_target_id,
             "target_name": "Fidesops Autogenerated Erasure Target",
             "target_key": "fidesops_autogenerated_erasure_target",
-            "data_category": DataCategory("user.provided.identifiable").value,
+            "data_category": DataCategory("user").value,
```
pattisdr, line 334: This edits a past data migration.  If any fidesops user has already run this data migration, these changes will not take affect. Instead of editing an old migration, I'd create a new data migration that updates these targets to the new categories.

tests/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -140,6 +140,7 @@ def test_put_validate_dataset_nested_collections(
             validate_dataset_url, headers=auth_header, json=invalid_dataset
         )
         json_response = json.loads(response.text)
+        print(json_response)
```
pattisdr, line 143: Committed print statement

tests/service/connectors/test_queryconfig.py
```diff
@@ -189,9 +189,9 @@ def test_update_rule_target_fields(
 
         # Make target more broad
         target = rule.targets[0]
-        target.data_category = DataCategory("user.provided.identifiable").value
+        target.data_category = DataCategory("user.contact").value
```
pattisdr, line 192: Is this what you intended to do? Swapping user.provided.identifiable for user.contact?

tests/service/connectors/test_queryconfig.py
```diff
@@ -534,6 +535,11 @@ def test_generate_update_stmt_multiple_fields(
             }
         }
 
+        print(mongo_statement[1])
+        print(expected_result_1)
```
pattisdr, line 539: Committed print statements

tests/task/test_filter_results.py
```diff
@@ -898,24 +895,42 @@ def test_filter_data_categories_limited_results():
...
+            }
+        ],
+        "postgres_example:login": [
+            {"time": datetime(2021, 1, 6, 1, 0), "customer_id": 3}
         ],
     }
+
+    print(filtered_results)
+    print("-" * 10)
+    print(expected_results)
```
pattisdr, line 935: Committed print statement

pattisdr (commented):
Understand this is still in draft, just did a quick skim while I was in here looking at the side effects of removing derived/provided categories. Will test more thoroughly when it's ready for review.

tests/task/test_filter_results.py
ThomasLaPiana: deliberately added, the stdout of this test will only get shown on a failure in this test, and the current error message is very difficult to debug with. This will make future debugging easier (apply for all places where I've added a print statement to a test)

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
ThomasLaPiana: This was unavoidable, the package itself has changed so `user.provided.identifiable` now throws an error and breaks the whole migration

pattisdr: Instead of using `DataCategory` then in this migration, I'd update to use strings: "user.provided.identifiable" here, and still add a new migration to update the existing autogenerated rule target.

tests/integration_tests/test_mongo_task.py
```diff
@@ -283,7 +283,7 @@ def test_mongo_erasure_task(db, mongo_inserts, integration_mongodb_config):
         "mongo_test:customer": 1,
         "mongo_test:payment_card": 0,
         "mongo_test:orders": 0,
-        "mongo_test:address": 2,
+        "mongo_test:address": 0,
```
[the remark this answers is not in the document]
pattisdr: This test tries to remove a lot of variables and just focus on masking behavior, so we assign arbitrary data categories that are different from the fideslang data categories, and builds a very small graph to mask.

The address table is expecting record with id 1000 and record with id 1002 to have their city and state fields masked.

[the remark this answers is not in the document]
ThomasLaPiana: @pattisdr would you be able to help me get this PR over the line? We can either pair sync or just collab on this PR, but I don't know enough about what these tests are doing to intelligently fix things without potentially voiding what the test was checking in the first place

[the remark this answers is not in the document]
pattisdr: @ThomasLaPiana Yes, I can look into this and other failing tests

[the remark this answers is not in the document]
ThomasLaPiana: @pattisdr appreciated! 🙏 I'll go through your other comments as well

Makefile
```diff
@@ -211,7 +211,6 @@ teardown:
 	-f docker/docker-compose.integration-mysql.yml \
 	-f docker/docker-compose.integration-postgres.yml \
 	down \
-	--volumes \
```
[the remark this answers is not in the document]
ThomasLaPiana: I seemed to be having an issue where the database wasn't getting initialized on startup, as well as some idiosyncrasies with testing the integrations. I'm fine leaving it in (i was the one that  put it there) but it seemed to be causing some problems. I'll revisit that more thoroughly in #439

pattisdr:
Remaining test failures as of [9454f91](https://github.com/ethyca/fidesops/pull/890/commits/9454f911aca9d6bc68159833608513bafda7b266) are our existing MSSQL failures https://github.com/ethyca/fidesops/issues/813

tests/service/connectors/test_queryconfig.py
ThomasLaPiana: don't believe I did this on purpose (my memory isn't trustworthy), probably a sloppy find/replace

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
ThomasLaPiana: very elegant solution 👍

tests/service/connectors/test_queryconfig.py
ThomasLaPiana: I remember now that I did it to try and understand what the test was doing/what would pass. have reverted to the `user`

ThomasLaPiana:
next step here is for me to manually write the migration and confirm that everything works. It is unable to be autogenerated it seems

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+        "user.derived",
+        "user.derived.identifiable",
+        "user.derived.nonidentifiable",
+        "user.derived.nonidentifiable.sensor",
+        "user.provided",
+        "user.provided.identifiable",
+        "user.provided.nonidentifiable",
+    )
+
+    return privacy_key in removed
```
sanders41, line 120: Once removed in an upgrade, a downgrade will not be able to add these back because there will be no way to know where they should go. I don't think there is anything we can do about this.

ThomasLaPiana:
@sanders41 now that's a migration...

sanders41:
I believe the failing test is related to #992. I will restart the test in a while to see if it resolves.

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
```diff
@@ -140,6 +140,7 @@ def test_put_validate_dataset_nested_collections(
             validate_dataset_url, headers=auth_header, json=invalid_dataset
         )
         json_response = json.loads(response.text)
+        print(json_response)
```
seanpreston, line 143: We'll want to remove the `print`s in this file I think

tests/ops/service/connectors/test_queryconfig.py
```diff
@@ -534,6 +535,11 @@ def test_generate_update_stmt_multiple_fields(
             }
         }
 
+        print(mongo_statement[1])
+        print(expected_result_1)
```
seanpreston, line 539: We'll want to remove these `print`s

tests/ops/api/v1/endpoints/test_dataset_endpoints.py
ThomasLaPiana: @seanpreston i left these in specifically. Prints are very helpful in tests, especially with complicated object comparisons. They also don't actually get printed unless that specific test fails, so it also doesn't hurt anything to leave it in

tests/ops/service/connectors/test_queryconfig.py
ThomasLaPiana: see answer above, without these prints I had no way to know why and what broke

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -276,7 +276,7 @@ def autogenerate_access_policy(
             "target_id": rule_target_id,
             "target_name": "Fidesops Autogenerated Access Target",
             "target_key": "fidesops_autogenerated_access_target",
-            "data_category": DataCategory("user.provided.identifiable").value,
+            "data_category": DataCategory("user").value,
```
seanpreston, line 279: Should we be changing legacy migrations? This ought to be handle in the new migration that's been added. I'm not sure it actually makes a difference here, but technically speaking the migration history should be preserved.

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def update_field(field: dict, up_down: str) -> dict:
+    data_categories = field.get("data_categories")
+    updated = []
+    if up_down == "up":
```
seanpreston, line 126: Nit: could you name this something more obvious?

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+        if data_categories:
```
seanpreston, line 127: I think we could avoid this extra check if you set `data_categories = field.get("data_categories", [])` on line 124

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+                        new = DATA_MAP_UPGRADE.get(data_category)
+                        if new:
+                            updated.append(new)
+                        else:
+                            updated.append(data_category)
+                    else:
+                        updated.append(data_category)
+        field["data_categories"] = updated
+
+        if field.get("fields"):
```
seanpreston, line 144: This catches nested fields right? Since we're calling this in both `up_down` blocks, we could call this only once at the end outside of the `if` / `else` blocks.

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+                update_field(field, up_down)
+
+    return field
+
+
+def upgrade() -> None:
+    sessionlocal = get_db_session(config)
+    with sessionlocal() as session:
+        try:
+            datasets = DatasetConfig.all(db=session)
```
seanpreston, line 178: Something I've found really useful with (potentially) large data migrations in the past is some sense of a progress indicator via logging. Now we've fetched the entire dataset it'd be good to log the count here, and current iteration below (or maybe every `x` to avoid spamming the logs).

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+            pass
+
+
+def downgrade() -> None:
+    sessionlocal = get_db_session(config)
+    with sessionlocal() as session:
+        try:
+            datasets = DatasetConfig.all(db=session)
+
+            for dataset in datasets:
```
seanpreston, line 203: As above re: progress check

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
```diff
@@ -0,0 +1,217 @@
...
+                                field = update_field(field, "up")
+
+        # If the datasetconfig table doesn't exist a ProgrammingError will be thrown.
+        # If that happens we know there are no records to update so we can ignore the
+        # error.
+        except ProgrammingError:
+            pass
+
+
+def downgrade() -> None:
```
seanpreston, line 197: This is more or less a copy of the `upgrade` function — consider DRY-ifying them both into a single function that takes an `up` / `down` arg.

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
sanders41: Because of the recursion it could be tricky to give an accurate count. The only length we will know is the top level, but we won't know how deep each level goes. Thoughts on how to handle that?

seanpreston: Great point. I don't have a good solution for that. As long as the user gets some input, anything is better than nothing.

src/fidesops/migrations/versions/55d61eb8ed12_add_default_policies.py
ThomasLaPiana: @seanpreston this could be changed to just a string, "user.provided.identified", but the current version of fideslang will break if we try to use a data category that doesn't exist

I think we made that change elsewhere too and it seemed to work

src/fidesops/migrations/versions/7abe778b7082_update_fideslang_data_categories.py
sanders41: I added some more logging. How helpful it will be is going to depend on how deep vs how wide things are. See if you think it will be good enough.

ThomasLaPiana:
marked all `outdated` conversations are resolved

ThomasLaPiana:
It looks like the migration order got messed up, will fix

ThomasLaPiana:
seems like a legit failure @sanders41 but I can't tell for sure

sanders41:
The hubspot tests have been failing on and off. There is an open issue on this #992. Seeing if a restart passes.

sanders41:
@adamsachs or @galvana does this looks like a real issue or is it a random hubspot failure?

galvana:
The failing tests for Hubspot are not caused by this change. There is a second round of work being done right now for Hubspot where we'll address the intermittent test failures.

sanders41:
@seanpreston the rebase is done and the only failing test is the hubspot one so we are ready for final review now.

ThomasLaPiana:
@sanders41 ty ty 💯

seanpreston (approved):
Thanks folks!

PR #795, opened 2022-07-05
Endpoint: Return Secrets for a Connector Type [#753]

pattisdr:
❗  Dependent on #768 . Merge that first.

# Purpose
To enable UI users to configure datastores from the UI, we need to create a way to tell the frontend what types of secrets are needed for each connector. Because each database and SaaS app have differing authentication requirements, we need a modular front end that can display 1 to many secrets (on the scale of 2 to 6).

# Changes
- Add an API endpoint that surfaces the secrets needed for both database and saas connectors

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [x] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [x] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [x] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [x] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #753

PR #768, opened 2022-06-30
GET Available Connectors [#706]

pattisdr:
❗ Dependent on #706 - Merge that first

# Purpose
As a Fidesops admin user, need the ability to see a list of available connector types to choose from so that I can select one to connect my fidesops environment to.

# Changes
-  Adds an endpoint to get available connectors that are ready to go (this is a combination of datastore and saas connectors minus ("manual", "custom", generic "saas", and "https" connectors.) `GET api/v1/connection_type`
    - Can be filtered on whether search param is a substring of the connector name `GET api/v1/connection_type?search=red`

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [x] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [x] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [x] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [x] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #752

Issue #760, opened 2022-06-30
Add a Connection - Select a connector to configure (front end)


adriaaaa:
### Use Cases

As the owner of a system that contains PII, I need the ability to select a connector to configure so that I can specify the connection details so Fides can query that system to fulfill a privacy request.

### Assumptions
* The scope of this ticket is limited to selecting a SaaS type connector
* We do not have the concepts of "categories" yet, so exclude this from the "Choose your Connection" screen for now.
* use feature flag for this body of work
* dual sorting will be handled on the FE
* reminder - do not include the categories menu on left

### Acceptance Criteria
1. Demonstrate that a user can add a new connection
* If there are no connection configured, show that "no connections" landing page (note: this page may already exist, will need to confirm if it does or not. also confirm that it matches the above mockup)
* If there already connections configured, the button for "Create a Connection" should appear above the list of configured connections.
* When use selects "Add a New Connection" they should be taken to a page where they can choose from a list of supported connectors.
* The list of supported connections should be displayed in ascending alphanumeric order. 
* No pagination, just display an entire list of supported connectors
2. Demonstrate that a user can narrow down the list of available connectors
* This will be done via dropdown with options for  "Third party connectors" and "Database connectors"
* The dropdown for type of connector should default to showing all connector types.
* "Database connectors" should be greyed out/read-only. When user hovers over the "Database connectors" option they should see a tooltip that says "You cannot add database connections from here at this time. Please refer to our [documentation](https://ethyca.github.io/fidesops/guides/database_connectors/) on how you can do so from our API."(note this is temporary until we can support adding database connectors)
3. Demonstrate that a user can search for a specific connector.
* search by the name of the type of connector (e.g. "Slack" or "Hubspot")
* When searching for a connection, use the "TypeAhead" we discussed.
4. Show that a user can go back to the landing page by selecting the breadcrumb under the title OR selecting the "Datastore Connections" menu option.

### Mockups
[Figma](https://www.figma.com/file/MTbZohIGvB4bq2f2wkiDHO/Fidesops-Management-UI?node-id=106%3A1469)
![Image](https://user-images.githubusercontent.com/87208929/179291403-451c7bde-b3cc-4dbc-8097-adf72cc86029.png)

![Image](https://user-images.githubusercontent.com/87208929/179293845-b5a82d4a-8038-4f87-8dc9-4d48d1308be3.png)

[...]

PR #937, opened 2022-07-25
Add db vs saas to connection type api

### LINKED ISSUE #910: Backend Support: Select a Connector to Configure

Add any changes needed or document existing endpoints to support selecting a connector to configure #760 

### Background 

We already have `GET /api/v1/connection_type` which returns a combined list of both database and saas options, but the latest UI designs show you being able to search just saas or just database connectors.   We need to surface if each option is database or saas so they can be sorted client-side.

Current:
`GET /api/v1/connection_type`
```json
{
    "items": [
        "bigquery",
        "hubspot",
        "mailchimp",
        "mariadb",
        "mongodb",
        "mssql",
        "mysql",
        "outreach",
        "postgres",
        "redshift",
        "salesforce",
        "segment",
        "sentry",
        "snowflake",
        "stripe",
        "zendesk"
    ],
    "total": 15,
    "page": 1,
    "size": 50
}
```

- Modify `GET /api/v1/connection_type` to return a list of dictionaries instead of a list of strings.  We return both the identifier of the connection type and whether it is "saas connector" or a "database connector".  Also remove pagination from the backend.  This can then be searched client-side.  If the user narrows down to search just third party or database, we can use the "type" param to limit just those options from being displayed/searched client side.
```json
{
    "items": [
        {"identifier": "bigquery", "type": "database"},
        {"identifier": "hubspot", "type": "saas"}
        ...
    ],
    "total": 15,
    "page": 1,
    "size": 50
}
```
- 

### Additional context

eastandwestwind:
# Purpose
Add db vs saas to connection type api

# Changes
- Return db vs saas system type from connection type api
- Adds query param to filter for `system_type`, in addition to the already existing `search` query param. e.g `?search=str&system_type=saas`
- Remove pagination from endpoint
- Update tests

# Checklist

# Ticket

Fixes #910

### DIFF
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -93,10 +93,7 @@ def get_connection_type_secret_schema(
     to authenticate.
     """
     connection_system_types: List[ConnectionSystemTypeMap] = get_connection_types()
-    if not any(
-        item.identifier == connection_type
-        for item in connection_system_types
-    ):
+    if not any(item.identifier == connection_type for item in connection_system_types):
         raise HTTPException(
             status_code=HTTP_404_NOT_FOUND,
             detail=f"No connection type found with name '{connection_type}'.",
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -94,7 +94,7 @@ def get_connection_type_secret_schema(
     """
     connection_system_types: List[ConnectionSystemTypeMap] = get_connection_types()
     if not any(
-        item.identifier and item.identifier.value == connection_type
+        item.identifier == connection_type
         for item in connection_system_types
     ):
         raise HTTPException(
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -94,7 +94,7 @@ def get_connection_type_secret_schema(
     """
     connection_system_types: List[ConnectionSystemTypeMap] = get_connection_types()
     if not any(
-        item.identifier.value == connection_type for item in connection_system_types
+            item.identifier and item.identifier.value == connection_type for item in connection_system_types
     ):
         raise HTTPException(
             status_code=HTTP_404_NOT_FOUND,
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -94,7 +94,8 @@ def get_connection_type_secret_schema(
     """
     connection_system_types: List[ConnectionSystemTypeMap] = get_connection_types()
     if not any(
-            item.identifier and item.identifier.value == connection_type for item in connection_system_types
+        item.identifier and item.identifier.value == connection_type
+        for item in connection_system_types
     ):
         raise HTTPException(
             status_code=HTTP_404_NOT_FOUND,
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -35,7 +35,7 @@ def get_connection_types(search: Optional[str] = None) -> List[ConnectionSystemT
         return search in elem if search else True
 
     connection_system_types: List[ConnectionSystemTypeMap] = []
-    database_types: List[ConnectionType.value] = sorted(
+    database_types: List[str] = sorted(
         [
             conn_type.value
             for conn_type in ConnectionType
@@ -50,7 +50,7 @@ def get_connection_types(search: Optional[str] = None) -> List[ConnectionSystemT
             for item in database_types
         ]
     )
-    saas_types: List[SaaSType.value] = sorted(
+    saas_types: List[str] = sorted(
         [
             saas_type.value
             for saas_type in SaaSType
--- a/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
+++ b/src/fidesops/api/v1/endpoints/connection_type_endpoints.py
@@ -1,10 +1,8 @@
 import logging
 from typing import Any, Dict, List, Optional
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, HTTPException
 from fastapi.params import Security
-from fastapi_pagination import Page, Params, paginate
-from fastapi_pagination.bases import AbstractPage
 from starlette.status import HTTP_404_NOT_FOUND
 
 from fidesops.api.v1.scope_registry import CONNECTION_TYPE_READ
@@ -18,6 +16,10 @@ from fidesops.schemas.connection_configuration import (
     SaaSSchemaFactory,
     secrets_validators,
 )
+from fidesops.schemas.connection_configuration.connection_config import (
+    ConnectionSystemTypeMap,
+    SystemType,
+)
 from fidesops.schemas.saas.saas_config import SaaSConfig, SaaSType
 from fidesops.util.oauth_util import verify_oauth_client
 from fidesops.util.saas_util import load_config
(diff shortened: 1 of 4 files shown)


src/fidesops/api/v1/endpoints/connection_type_endpoints.py
```diff
@@ -27,41 +29,55 @@
...
 def get_all_connection_types(
-    *, params: Params = Depends(), search: Optional[str] = None
-) -> AbstractPage[str]:
+    *, search: Optional[str] = None
+) -> List[ConnectionSystemTypeMap]:
     """Returns a list of connection options in Fidesops - includes only database and saas options here."""
 
-    connection_types: List[str] = get_connection_types(search)
-
-    return paginate(connection_types, params)
```
seanpreston, line 64: Is there a technical constraint that requires us to remove the pagination?

eastandwestwind: @pattisdr had suggested this in the ticket- https://github.com/ethyca/fidesops/issues/910 

Alongside this, the server-side logic to get all connection types for these endpoints isn't costly enough and the data not large enough to really require pagination.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
```diff
@@ -27,41 +29,55 @@
...
 @router.get(
     CONNECTION_TYPES,
     dependencies=[Security(verify_oauth_client, scopes=[CONNECTION_TYPE_READ])],
-    response_model=Page[str],
+    response_model=List[ConnectionSystemTypeMap],
 )
 def get_all_connection_types(
-    *, params: Params = Depends(), search: Optional[str] = None
-) -> AbstractPage[str]:
+    *, search: Optional[str] = None
```
seanpreston, line 76: If we're adding the pagination `params` back into this endpoint, a query param for the type of connector could be useful. ie. `?connection_type=saas` or `?connection_type=db`.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
```diff
@@ -27,41 +29,55 @@
...
+        [
+            ConnectionSystemTypeMap(identifier=item, type=SystemType.database)
+            for item in database_types
+        ]
+    )
+    saas_types: List[str] = sorted(
+        [
+            saas_type.value
+            for saas_type in SaaSType
+            if saas_type != SaaSType.custom and is_match(saas_type.value)
```
seanpreston, line 57: This endpoint will never return custom SaaSType connections, should we also include those?

pattisdr: On the initial version of this endpoint, we had a discussion with Adria that we're not returning custom connector at this time as an option, https://github.com/ethyca/fidesops/pull/795#discussion_r914117170, the UI isn't yet supporting that flow.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
pattisdr: I specified no pagination in the endpoint just to support @adriaaaa's acceptance criteria in the frontend for adding a connector

https://github.com/ethyca/fidesops/issues/760

> No pagination, just display an entire list of supported connectors

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
eastandwestwind: I wanted to follow the pattern we had before, where we're only including pre-registered / existing connectors for now, see https://github.com/ethyca/fidesops/pull/768#discussion_r911470444

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
eastandwestwind: good idea!

eastandwestwind: fyi we already have a `?search` param , where we filter by connection type (postgres/stripe), not system type (saas / db). 

So for this new param, I'm thinking naming it `system_type`

pattisdr: I was under the impression we were just going to be giving the entire list to @chriscalhoun1974 and he was going to be handling everything else client side.

This comment is just related to client-side _sorting_, but maybe he can clarify on whether all of these pieces are needed server side.

https://github.com/ethyca/fidesops/issues/760#issuecomment-1190312249

chriscalhoun1974: Correct, I will handle the sorting of the result set via client-side.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
eastandwestwind: @seanpreston good point with the consistency. I'm OK adding it back, just let me know which you'd prefer.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
eastandwestwind: I guess this is a question of whether we should implement convenience features vs features that are actually needed by client side now. Open to either, but will require more testing for adding the new param

eastandwestwind: going ahead and adding it

seanpreston: > I guess this is a question of whether we should implement convenience features vs features that are actually needed by client side now. Open to either, but will require more testing for adding the new param

Given these a relatively light lifts, it doesn't warrant compromising the quality of our APIs. Thanks for adding!

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
seanpreston: Thanks! I prefer it in, let's keep consistency unless there's a really clear incentive not to. In this case we can have both client side awareness of the whole set and pagination for relatively low cost, so let's keep it

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
```diff
@@ -27,41 +29,63 @@
...
+    connection_system_types: List[ConnectionSystemTypeMap] = []
+    if system_type == SystemType.database or system_type is None:
+        database_types: List[str] = sorted(
+            [
+                conn_type.value
+                for conn_type in ConnectionType
+                if conn_type
+                not in [
+                    ConnectionType.saas,
+                    ConnectionType.https,
```
seanpreston, line 48: We should probably remove this value all together, since it's not used and is a duplicate of `saas`. We can do that in a follow-up.

tests/api/v1/endpoints/test_connection_template_endpoints.py
```diff
@@ -34,33 +41,82 @@ def test_get_connection_types(
...
+        } in data
+        assert {
+            "identifier": SaaSType.stripe.value,
+            "type": SystemType.saas.value,
+        } in data
+
+        assert "saas" not in [item["identifier"] for item in data]
+        assert "https" not in [item["identifier"] for item in data]
+        assert "custom" not in [item["identifier"] for item in data]
+        assert "manual" not in [item["identifier"] for item in data]
```
seanpreston, line 58: Thanks for being thorough!

seanpreston (approved):

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
pattisdr: What do you mean it isn't used? `https` connectors are the type of connector you create if you're making a pre- or post-execution webhook.

pattisdr: Ah, spoke too soon, you're talking about from the ConnectionType enum.

src/fidesops/api/v1/endpoints/connection_type_endpoints.py
adriaaaa: @seanpreston @eastandwestwind @pattisdr - one concern about pagination though is how we search for integrations from the front end. We specified that it would be that "typeahead" method in refinement. If we do pagination on this page, would we be limited to the results on that page and therefore it would be considered an incomplete search? May not be the case, but I do just want to confirm.

adriaaaa: @eastandwestwind @seanpreston @pattisdr one concern I have about including pagination when a user searches on the front end using that "typeahead" method we discussed. If this UI is paginated, would a search only return matching connections that are on that page? If so, that would be considered an incomplete search by the end user. May not be the case but want to bring it up just in case. 

Also, may be good to run this by the design team quickly because this would be a change with their designs they provided us.

eastandwestwind: @adriaaaa , if we had pagination here, you're correct in that if the amount of data exceeds a certain threshold, a search would only return 1 page of connections, with the need for another api call to return the next page. As we scale and/or add more data to this endpoint, @seanpreston is right that pagination will probably be necessary in the future.

seanpreston: >  would a search only return matching connections that are on that page?

Search via the API is done at the database level, which always has full context of the available dataset, so if the search term is submitted to the API then what you're worried about here won't be a problem.

If the search isn't submitted via the API, and instead done via the client-side code checking results from the API, then including pagination means we have to be sure that the client has a full collection of the results. We can increase the page size of what's requested in this http request to guarantee we'll fetch the whole lot in the first go, so this shouldn't be an issue.

PR #945, opened 2022-07-25
Dispatch a repository event on new published releases

PSalant726:
# Purpose

There is no way for a GH Actions workflow to read the releases in another repository. In order for an issue to be automatically created in the fidesops-plus repository whenever a new fidesops release is published, the act of publishing the fidesops release must dispatch a repository event to the fidesops-plus repo. This event is used to trigger a separate GH Actions workflow in the fidesops-plus repository, which creates the desired issue.

# Changes

* Add a new GH Actions workflow to dispatch an event on new releases

# Checklist

# Ticket

Related to ethyca/fidesops-plus#43
Blocks ethyca/fidesops-plus#50

### DIFF
--- a/.github/workflows/release_event.yml
+++ b/.github/workflows/release_event.yml
@@ -1,0 +1,17 @@
+name: Release Dispatch Event
+
+on:
+  release:
+    types: [released]
+
+jobs:
+  Notify-of-New-Release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.DISPATCH_ACCESS_TOKEN }}
+          repository: ethyca/fidesops-plus
+          event-type: new-fidesops-release
+          client-payload: '{"tag": "${{ github.event.release.tag_name }}", "url": "${{ github.event.release.html_url }}"}'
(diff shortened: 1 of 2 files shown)


.github/workflows/release_event.yml
```diff
@@ -0,0 +1,17 @@
...
+    types: [released]
+
+jobs:
+  Notify-of-New-Release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.DISPATCH_ACCESS_TOKEN }}
```
PSalant726, line 14: This secret must be added to the repository. It must be a personal access token with the repo scope. I have created one in my own GH account which we can use for testing, but the token used in perpetuity should be associated with an organization-level GH account (if we have one).

I don't have permissions to add secrets to the repo, so someone with admin privileges will have to do so on my behalf. Reach out to me via Slack for the token value.

sanders41:
@PSalant726 this is the same action you were testing on fidesctl and it was working there correct?

PSalant726:
> this is the same action you were testing on fidesctl and it was working there correct?

Worked first try with no changes needed 😎

sanders41 (approved):

.github/workflows/release_event.yml
seanpreston: > the token used in perpetuity should be associated with an organization-level GH account (if we have one)

@NevilleS do we have one of these?

PSalant726: We don't have any such account right now.

PR #954, opened 2022-07-26
Allow `run_infrastructure.py` to bring up no datastores.

seanpreston:
# Purpose

I've noticed that in certain places in our util scripts we're passing in `postgres` perhaps thinking this spins up the application database. It actually spins up the `example_postgres` datastore that we use for integration testing. We may also be doing this to avoid spinning up _all_ datastores in those utli commands, because the `run_infrasructure.py` script will create all datastores by default if none are specified (so specifying only `postgres` is the lesser of those evils).

This PR updates `run_infrastructure.py` to allow no `datastores` to be spun up if an empty list is explicitly passed, meaning we don't have to wait for `example_postgres` to come-up in our util commands.

### DIFF
--- a/noxfiles/utils_nox.py
+++ b/noxfiles/utils_nox.py
@@ -25,13 +25,13 @@ COMPOSE_DOWN = (
 @nox.session()
 def create_user(session: nox.Session) -> None:
     """Create a super user in the fidesops database."""
-    run_infrastructure(datastores=["postgres"], run_create_superuser=True)
+    run_infrastructure(datastores=[], run_create_superuser=True)
 
 
 @nox.session()
 def seed_test_data(session: nox.Session) -> None:
     """Seed test data in the Postgres application database."""
-    run_infrastructure(datastores=["postgres"], run_create_test_data=True)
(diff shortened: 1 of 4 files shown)


noxfiles/utils_nox.py
```diff
@@ -25,13 +25,13 @@
...
     """Create a super user in the fidesops database."""
-    run_infrastructure(datastores=["postgres"], run_create_superuser=True)
+    run_infrastructure(datastores=[], run_create_superuser=True)
 
 
 @nox.session()
 def seed_test_data(session: nox.Session) -> None:
     """Seed test data in the Postgres application database."""
-    run_infrastructure(datastores=["postgres"], run_create_test_data=True)
+    run_infrastructure(datastores=[], run_create_test_data=True)
```
seanpreston, line 34: I noticed we can execute the infra commands like this now, allowing us to retain existing functionality but explicitly specify if we don't want any datastores to come up.

ThomasLaPiana:
@seanpreston smart change here! good catch

seanpreston:
I'm closing this for now, until we've migrated fully over to using `nox` to avoid wasting time updating two use cases.

PR #962, opened 2022-07-27
Update OAuth strategy to be able to perform local testing

### LINKED ISSUE #961: Update OAuth strategy to be able to perform local testing

With the addition of our Ethyca-hosted reverse-proxy callback server, we should be able to test the entire OAuth workflow for any connector. We need to add config for a developer to specify which ngrok tunnel to use for testing.

Add an `oauth_instance` Fidesops config which is prefixed to the `state` value of an OAuth authorization request when dev mode is enabled.

galvana:
# Purpose
With the addition of our Ethyca-hosted reverse-proxy callback server, we should be able to test the entire OAuth workflow for any connector. We need to add config for a developer to specify which ngrok tunnel to use for testing.

# Changes
- Added an oauth_instance Fidesops config which is prefixed to the state value of an OAuth authorization request when dev mode is enabled.
- Fixed the HTTP verb for the `/oauth/callback` endpoint from `POST` to `GET`

# Checklist

# Ticket

Fixes #961

### DIFF
--- a/fidesops.toml
+++ b/fidesops.toml
@@ -1,4 +1,5 @@
 port = 8080
+oauth_instance = "local"
 
 [database]
 server = "db"
--- a/fidesops.toml
+++ b/fidesops.toml
@@ -1,5 +1,4 @@
 port = 8080
-oauth_instance = "local"
 
 [database]
 server = "db"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,6 @@ The types of changes are:
 * Access and erasure support for Adobe Campaign [#905](https://github.com/ethyca/fidesops/pull/905)
 * Added db vs saas to connection type api [#937](https://github.com/ethyca/fidesops/pull/937)
 * Retry a DSR (FE) [#863](https://github.com/ethyca/fidesops/pull/938)
-* Added the `oauth_instance` system property to specify a value to prefix to the state of an outbound OAuth request [#962](https://github.com/ethyca/fidesops/pull/962)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The types of changes are:
 * Access and erasure support for Adobe Campaign [#905](https://github.com/ethyca/fidesops/pull/905)
 * Added db vs saas to connection type api [#937](https://github.com/ethyca/fidesops/pull/937)
 * Retry a DSR (FE) [#863](https://github.com/ethyca/fidesops/pull/938)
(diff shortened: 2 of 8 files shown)


fidesops.toml
```diff
@@ -1,4 +1,5 @@
 port = 8080
+oauth_instance = "local"
```
adamsachs, line 2: `local` will never really be a valid value here, if i'm following correctly - it's going to be something that uniquely identifies a given developer's ngrok tunnel. if that's the case, then giving a default value here that's not supposed to be used may be confusing.

this is really only meant to be used for internal testing/dev, right? if devs are going to need to update the value manually anyway (again, assuming i'm following this correctly), then maybe we should just not include here in `fidesops.toml` and have people set it via env var? the reason i suggest this is because this file, in the repo, can be used as a reference for clients/users of fidesops, and i'd worry about the confusion this may cause.

take the above with a large grain of salt - i'm still not sure i totally follow how this all works and the purpose of this config.

adamsachs (changes requested):
this all looks fine technically.

but i'm still not 100% clear on the purpose of the added prefix to the generated state string. perhaps this comes down to my lack of expertise in oauth2. i've tried to piece together  different documentation (both general oauth2 references and our internal docs), and my working understanding is that our internal reverse proxy that we use for dev work/testing will, as a matter of convention, take the prefix from the state param and route the incoming callbacks to the appropriate ngrok tunnel based on this identifier -- essentially, what seems to have been done for ethyca-pro, except tunneling to local dev environments instead.

if that's accurate, then i think this all makes sense, i just have a minor comment about how we should declare a default value for the `oauth_instance` config. if it's not accurate, then we'll need to touch base and talk through things.

either way, i do think the code could use some more documentation that explains _why_ we're doing what we're doing here. i've been digging through a lot of documentation that's not within the codebase and it's still not clear to me, so i think to ensure this context doesn't get lost going forward, it would help to provide some more explicit context somewhere in the code itself (e.g. [the `generate_state()` docstring](https://github.com/ethyca/fidesops/pull/962/files#diff-f4061918ba133c479f2cd00b3bd96cfa84e6125ab7f1656cc440bb3b493d7d66R81)). (to be clear, i'm referring specifically to any conventions/practices that we're relying upon here that's not "general oauth2 domain knowledge")

actually, on that note - we definitely need to update https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/guides/configuration_reference.md to include this and i'll also label the issue such that @ethyca/docs-authors is aware of it! even with that, though, i still think the suggestions mentioned above are worth considering

adamsachs (approved):
nice, this all looks good to me. thanks for addressing the suggestions and talking things through with me!

i do think we should update some documentation (internal if nothing else?) to make reference to the new config property, but that can be split out into its own issue

PR #966, opened 2022-07-28
Update the compose file and workflows to expect an already-built image

### LINKED ISSUE #932: Fully replace `make` with `nox`

When #919 gets merged, we'll have `nox` targets to fully replace the Makefile and should do so.

There will be a few steps here:

* [x]  Socialize the dev flow changes to the fidesops team (a short video should suffice going through some of the new usage patterns)
* [x]  updating the `docker-compose` file to match how `fidesctl` handles image building/usage (#966)
* [x]  Updating the docs where there is a reference to the Makefile (#960)
* [x]  The riskiest part, which is updating all of the CI workflows (#966)

ThomasLaPiana:
# Purpose

Updates the infra and workflows to follow the same pattern as fidesctl, this gives the following advantages:
* properly parallelize github workflows without rebuilding the image
* run tests in a container that matches prod

To that end, here are the major changes and some explanation of each:
1. You'll see that the `docker-compose.yml` no longer builds the docker image, and instead looks for a specific image named `ethyca/fidesops:local`. This has the advantage of letting us build images using various target stages and spinning them up with the compose file. The easiest example of this is testing, where we are now running our CI checks inside of a production container instead of a dev container. This means that we are testing the actual container our users get their hands on, which eliminates an entire class of "surprise" we could get after a release.
2. Enable by the above change ^ we can now run CI jobs in a more efficient way. Instead of build the container in every step, there is now a single initial step that builds the container and then stores it as a build artifact. All of the other jobs then spin up in parallel and load in that build artifact. This means that the container is now only being built _once_ and then reused across all of the jobs, with the exception of the `unsafe` jobs that require a second build due to their separate nature.
3. All of the CI checks have been switched over to using `nox` instead of `make` for their commands.
4. ⚠️  The `Makefile` is being removed in this PR! ⚠️ 

# Changes
* [x] update `docker-compose.yml` to look for the specific `ethyca/fidesops:local` image
* [x] update the CI workflows to build the image and then store it for use by the other jobs
* [x] add `check_migrations` to the nox targets
* [x] update `run_infrastructure` to use `COMPOSE_SERVICE_NAME` instead of `IMAGE_NAME` when calling the compose file
* [x] remove the Makefile and the `scripts/run_infrastructure.py` file used by it (there is a separate & updated `noxfiles/run_infrastructure.py`)

# Checklist

# Ticket

Fixes #932


ThomasLaPiana:
@seanpreston was there a reason that the unsafe CI checks are in a different file? I was wondering if they can live in the other file since at the job-level they still have the check for the `unsafe` label

ThomasLaPiana:
currently because of the separate `safe` and `unsafe` checks we're building the image twice, but everything is running in parallel as desired

ThomasLaPiana:
@seanpreston @sanders41 I'm working on this pylint error but otherwise this is ready for a prelim review! Thank you 🙏

.github/workflows/fidesops_unsafe_pr_checks.yml
```diff
@@ -0,0 +1,106 @@
...
+  pull_request:
+    types: [labeled]
+
+env:
+  CONTAINER: fidesops-local
+  IMAGE: ethyca/fidesops:local
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
```
sanders41, line 15: Do we need this here also?

```yaml
    if: contains(github.event.pull_request.labels.*.name, 'run unsafe ci checks')
```

docker-compose.yml
```diff
@@ -1,9 +1,6 @@
 services:
   webserver:
-    container_name: webserver
-    build:
```
sanders41, line 4: Just a question here. Why would we not want the image to build? Makes the test faster?

noxfiles/run_infrastructure.py
```diff
@@ -54,7 +55,7 @@ def run_infrastructure(
         _run_cmd_or_err(f'echo "datastores specified: {", ".join(datastores)}"')
 
     # De-duplicate datastores
-    datastores = set(datastores)
+    datastores = list(set(datastores))
```
sanders41, line 58: No problem with this, just a question. Is this because nox wants a list for the datastores?

ThomasLaPiana:
authors note: I've pinpointed the issue to the `get_current_tag` function causing the pylint break

noxfiles/ci_nox.py
```diff
@@ -106,7 +126,9 @@ def pytest_unit(session: nox.Session) -> None:
 def pytest_integration(session: nox.Session) -> None:
```
sanders41, line 126: I'm getting a bunch of errors on the startup of the integration test and getting some test failures, but I don't think it has anything to do with this PR. @seanpreston any thoughts on this?

```console
python: can't open file '/fidesops/tests/integration_tests/setup_scripts/postgres_setup.py': [Errno 2] No such file or directory
ERROR: 2
no custom setup logic found for postgres, skipping
Attempting to create schema and seed initial data for mongodb from tests/integration_tests/setup_scripts/mongodb_setup.py...
WARNING: The FIDESOPS__LOG_PII variable is not set. Defaulting to a blank string.
WARNING: The FIDESOPS__HOT_RELOAD variable is not set. Defaulting to a blank string.
Creating fidesops_webserver_run ... done
python: can't open file '/fidesops/tests/integration_tests/setup_scripts/mongodb_setup.py': [Errno 2] No such file or directory
ERROR: 2
no custom setup logic found for mongodb, skipping
Attempting to create schema and seed initial data for mariadb from tests/integration_tests/setup_scripts/mariadb_setup.py...
WARNING: The FIDESOPS__LOG_PII variable is not set. Defaulting to a blank string.
WARNING: The FIDESOPS__HOT_RELOAD variable is not set. Defaulting to a blank string.
Creating fidesops_webserver_run ... done
python: can't open file '/fidesops/tests/integration_tests/setup_scripts/mariadb_setup.py': [Errno 2] No such file or directory
ERROR: 2
no custom setup logic found for mariadb, skipping
Attempting to create schema and seed initial data for mysql from tests/integration_tests/setup_scripts/mysql_setup.py...
WARNING: The FIDESOPS__LOG_PII variable is not set. Defaulting to a blank string.
WARNING: The FIDESOPS__HOT_RELOAD variable is not set. Defaulting to a blank string.
Creating fidesops_webserver_run ... done
python: can't open file '/fidesops/tests/integration_tests/setup_scripts/mysql_setup.py': [Errno 2] No such file or directory
ERROR: 2
no custom setup logic found for mysql, skipping
Attempting to create schema and seed initial data for mssql from tests/integration_tests/setup_scripts/mssql_setup.py...
WARNING: The FIDESOPS__LOG_PII variable is not set. Defaulting to a blank string.
WARNING: The FIDESOPS__HOT_RELOAD variable is not set. Defaulting to a blank string.
Creating fidesops_webserver_run ... done
python: can't open file '/fidesops/tests/integration_tests/setup_scripts/mssql_setup.py': [Errno 2] No such file or directory
ERROR: 2
no custom setup logic found for mssql, skipping
```

ThomasLaPiana: @sanders41 that's `run_infrastructure` trying to seed the data it looks like

also it should be `fidesops/tests/ops/` I think? I'm not sure how its wrong but still working

noxfiles/run_infrastructure.py
ThomasLaPiana: mypy was complaining, it was a vscode nit lol

docker-compose.yml
ThomasLaPiana: @sanders41 I left some explanation in the description above

.github/workflows/fidesops_unsafe_pr_checks.yml
ThomasLaPiana: most likely, but can we just combine the files altogether and use this check?

.github/workflows/fidesops_safe_pr_checks.yml
```diff
@@ -0,0 +1,239 @@
...
+      - name: Install Nox
+        run: pip install nox>=2022
+
+      - name: Check fidesops installation
+        run: nox -s check_install
+
+      - name: Check migrations
+        run: nox -s check_migrations
+
+  isort:
```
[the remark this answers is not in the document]
ThomasLaPiana: I actually have an idea to improve this even further...will update

.github/workflows/fidesops_safe_pr_checks.yml
```diff
@@ -0,0 +1,239 @@
...
+    branches:
+      - main
+  pull_request:
+
+env:
+  CONTAINER: fidesops-local
+  IMAGE: ethyca/fidesops:local
+
+jobs:
+  Build:
```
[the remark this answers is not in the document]
ThomasLaPiana: 5 minutes? that seems kind of high, but also I don't think 5 minutes should make a huge difference here, CI latency shouldn't be _too_ important and this method feels more efficient. If people are debugging in CI that's an anti-pattern and we should figure out the cause and fix it in my opinion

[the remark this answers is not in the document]
sanders41: I was surprised by the 5 minutes, I was expecting they way you did it here to be faster. But I was consistently seeing it take about that much longer.

.github/workflows/fidesops_unsafe_pr_checks.yml
sanders41: The reason this _may_ not work...the code will get pushed and checks run, then later the label gets added to run the unsafe checks. This label could get added after the safe check have already finished because the reviewer decides it is need. Would this be a problem?

docker-compose.yml
sanders41: This is fine with me, we may just want a note somewhere (maybe the README?) about this. I'm guessing we will get questions about it since it is different and people could get "errors" if they needed to run the build first now.

ThomasLaPiana: oh you're totally right....`nox -s dev` will automatically run it, but everywhere that has `docker compose up webserver` it won't exist...

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
ThomasLaPiana: @sanders41 I'm going to open a separate issue for this, but I'm going to refactor the static analysis checks to handle setting up their own environments and not rely on docker, good call

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
ThomasLaPiana: The isort job took 51 seconds total, and the initial build took ~5 minutes. seems fine to me?

.github/workflows/fidesops_unsafe_pr_checks.yml
ThomasLaPiana: hmmm good question....at that point could they restart the job? Worst-case scenario they'd have to push a dummy commit to trigger the job again right?

noxfiles/ci_nox.py
ThomasLaPiana: alright I fixed the directory name in `noxfiles/run_infrastructure.py` and that seems to have done it! good catch

sanders41: Looks good now! There are still a couple errors (mongodb_setup.py and mssql_setup.py), but those error have always been there because the files really don't exist. The tests don't seem to need the setup from those 2 files.

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
sanders41: Well, initial build takes around 8 minutes which adds about 5 minutes total because the other stages are faster, they just took longer to get started, but that is splitting hairs.

Really it's just that there is 5is minutes of unneeded wait time before tests start. Looks like the plan in #990 is to make the static checks run with nox outside of docker. At that point the the static checks become a non issue and the only question remains with the other tests (unit, integration). There the test suite goes from taking around 10 minutes to around 15 minutes because of waiting for the build before starting the tests.

.github/workflows/fidesops_unsafe_pr_checks.yml
sanders41: Right now the job automatically starts as soon as the label is applied and that still may happen, I'm not sure how that would work.

I don't think a blank commit would work because, with the way it works now, a new commit doesn't re-trigger the unsafe checks. You have to remove the label and re-apply it to start the checks again. I actually found this confusing and surprising.

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
sanders41: Just to clarify my real question here: is there a benefit to sharing the build that I am not seeing that makes it worth it to add the extra time to CI? I wouldn't want to add time to CI if there isn't a good reason, but if there is a benefit that would be different.

docker-compose.yml
ThomasLaPiana: I'm going to fix all of those new docs issues in the other docs PR #981

.github/workflows/fidesops_unsafe_pr_checks.yml
ThomasLaPiana: so the unsafe checks only run once?  that doesn't seem optimal....
I think its worth opening an entirely other issue to figure out a more elegant solution to the unsafe checks in general, we can test this type of stuff out there. Thanks for the information!

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
ThomasLaPiana: The only real benefit here is minimizing CPU time in our CI pipelines. From an actual cost perspective is it actually impactful? No, not really, CI time is cheap and when we halve the number of repos it'll matter even less.

_but_...I'd like to think that CI runtime isn't very latency sensitive (to a point) and at the very least we're saving the environment a bit lol

Shouldn't the developer know the outcome of the CI checks before they even run?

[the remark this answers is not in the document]
ThomasLaPiana: To clarify, I'm not at all willing to die on this hill, just seems wasteful somehow to build the container 10 times in a single CI run

noxfiles/ci_nox.py
ThomasLaPiana: Yep, `run_infrastructure.py` is written to check for _all_ setups files and just silently error if it doesn't exist (probably a better way to do that, but thats for another PR)

.github/workflows/fidesops_safe_pr_checks.yml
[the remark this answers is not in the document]
sanders41: I looked at some new bench marks and think we can both be happy here. By removing the build from the docker compose file it looks to have sped up bringing the containers up enough to negate the benefits of the separate builds making total CI times roughly equal. That removes my only issue so let’s go with it.

sanders41 (approved):

sanders41:
@ThomasLaPiana this is still labeled don't merge. Are you ready to merge or are you still working on things?

ThomasLaPiana:
@sanders41 I was asked to let @seanpreston merge these larger changes in case they conflict with other work, but from my end this is ready

noxfiles/run_infrastructure.py
```diff
@@ -98,29 +99,31 @@ def run_infrastructure(
...
 ) -> None:
     """
     Seed the datastores with initial data as defined in the file at `setup_path`
     """
     _run_cmd_or_err('echo "Seeding initial data for all datastores..."')
     for datastore in datastores:
         if datastore in DOCKERFILE_DATASTORES:
-            setup_path = f"tests/integration_tests/setup_scripts/{datastore}_setup.py"
+            setup_path = (
+                f"tests/ops/integration_tests/setup_scripts/{datastore}_setup.py"
```
eastandwestwind, line 120: can we store `tests/ops` as a global const `ops_test_dir` that can be shared here and in  our `ci_nox.py` too?

ThomasLaPiana:
@seanpreston this should be good to go now, I'm hitting the hay

seanpreston:
Thanks for all the work on this @ThomasLaPiana + @sanders41

PR #986, opened 2022-07-28
saas request overrides

### LINKED ISSUE #815: Python overrides for SaaS config endpoints

We currently denote the logic needed to access a SaaS endpoint by configuring a `request` inside of a SaaS config:
```
endpoints:
    - name: messages
      requests:
        read:
          method: GET
          path: /3.0/conversations/<conversation_id>/messages
          param_values:
            - name: conversation_id
              references:
                - dataset: mailchimp_connector_example
                  field: conversations.id
                  direction: from
          data_path: conversation_messages
          postprocessors:
            - strategy: filter
              configuration:
                field: from_email
                value:
                  identity: email
...
```

This allows us to abstract out a few common use-cases such as how to get the data needed to make the request, how to post-process an API response, and how to paginate to get more results. The problem arises when a certain endpoint requires logic that cannot be described by our SaaS config (for example building non-standard request or more complex post-processing/pagination). The SaaS config could be extended to account for these non-standard scenarios but it would complicate the main SaaS request workflow and make it more difficult to support/troubleshoot.

Add functionality to be able to define a Python override function (i.e. a `request_override`) for any endpoint in a SaaS config, e.g.:
```
endpoints:
    - name: messages
      requests:
        read:
          request_override: mailchimp_messages_access
          param_values:
            - name: conversation_id
              references:
                - dataset: mailchimp_override_connector_example
                  field: conversations.id
                  direction: from
```
Note that the `param_values` field is still needed since this information is used for graph traversal and to denote which values are needed to make this request. The `grouped_inputs` may also be defined and leveraged. Also note that this means when a `request_override` field is specified, if any other fields on the request config are defined _besides_ `param_values` or `grouped_inputs`, the config will be rejected as invalid; if a `request_override` field  is _not_ specified, then the config validation works as it did before -- the `path` and `method` fields are required, and all other fields are optional.

The `request_override` field will point to a function that has been defined with the appropriate input parameters and returns the appropriate data type; this function will need to be registered with a central override function registry, using a function decorator. We will likely want to add contributor/developer documentation to provide a clear reference for how these overrides can be properly implemented and registered, as it's meant to be a true extension point.

Using the above example, the `mailchimp_messages_access` function may look something like this:

```
@SaaSRequestOverrideFactory.register(
    "mailchimp_messages_access", [SaaSRequestType.READ]
)
def mailchimp_messages_access(
    node: TraversalNode,
    policy: Policy,
    input_data: Dict[str, List[Any]],
    secrets: Dict[str, Any],
) -> List[Row]:

    # gather request params
    conversation_ids = input_data.get("conversation_id")

    # build and execute request for each input data value
    processed_data = []
    if conversation_ids:
        for conversation_id in conversation_ids:
            try:
                response = get(
                    url=f'https://{secrets["domain"]}/3.0/conversations/{conversation_id}/messages',
                    auth=(secrets["username"], secrets["api_key"]),
                )

            # here we mimic the sort of error handling done in the core framework
            # by the AuthenticatedClient. Extenders can chose to handle errors within
            # their implementation as they wish.
            except Exception as exc:  # pylint: disable=W0703
                if config.dev_mode:  # pylint: disable=R1720
                    raise ConnectionException(
                        f"Operational Error connecting to Mailchimp API with error: {exc}"
                    )
                else:
                    raise ConnectionException(
                        "Operational Error connecting to MailchimpAPI."
                    )
            if not response.ok:
                raise ClientUnsuccessfulException(status_code=response.status_code)

            # unwrap and post-process response
            response_data = pydash.get(response.json(), "conversation_messages")
            filtered_data = pydash.filter_(
                response_data,
                {"from_email": privacy_request.get_cached_identity_data().get("email")},
            )

            # build up final result
            processed_data.extend(filtered_data)

    return processed_data
```

### Additional Information

A more detailed proof-of-concept can be found here [https://github.com/ethyca/fidesops/pull/311/](https://github.com/ethyca/fidesops/pull/311/). Note that this POC does _not_ leverage the registration/factory decorator pattern that we've decided to use going forward.

### Override Interface Specification (API)
For override functions of `read` requests, your method signature must have the following arguments and return type in order for the override to work as expected:
```
def my_read_override_function(
    node: TraversalNode,
    policy: Policy,
    input_data: Dict[str, List[Any]],
    secrets: Dict[str, Any],
) -> List[Row]:
```

For override functions of `update`, `delete`, or `data_protection_request` requests, your method signature must have the following arguments and return type in order for the override to work as expected:
```
def my_update_override_function(
    param_values_per_row: List[Dict[str, Any]],
    policy: Policy,
    secrets: Dict[str, Any],
) -> int:
```

To register your override function, you must include the following decorator, where the first argument specifies the "id" by which the function is referenced in your SaaS configs, and the second argument is a `List` of `SaaSRequestType` enum values: ```@SaaSRequestOverrideFactory.register(
    "my_read_override_function", [SaaSRequestType.READ]
)```

The `SaaSRequestType` enum values are: 
```
class SaaSRequestType(Enum):
    """
    An `Enum` containing the different possible types of SaaS requests
    """

    READ = "read'"
    UPDATE = "update"
    DATA_PROTECTION_REQUEST = "data_protection_request"
    DELETE = "delete"
```

adamsachs:
# Purpose
Enables user (developer) to provide custom python extensions/overrides of saas request execution.

# Changes
- initial cut of saas request overrides framework
- includes mailchimp override as an example and test case
- also includes minor refactor of some of the saas request execution to enable smoother override

# Checklist
  - [x] documentation issue created (tag docs-team to complete issue separately)

# Ticket

Fixes #815


data/saas/config/request_override/mailchimp_override_config.yml
```diff
@@ -0,0 +1,81 @@
...
+
+  test_request:
+    method: GET
+    path: /3.0/lists
+
+  endpoints:
+    - name: messages
+      requests:
+        read:
+          request_override: mailchimp_messages_access
```
galvana, line 30: I like the name `request_override` 👍

src/fidesops/schemas/saas/saas_config.py
```diff
@@ -159,6 +160,36 @@ def validate_grouped_inputs(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                    k
+                    for k in values.keys()
+                    if values.get(k)
+                    and k not in ("request_override", "param_values", "grouped_inputs")
+                ),
+                None,
+            )
+            if invalid:
+                raise ValueError(
+                    f"Invalid property {invalid} on a request with request_override specified"
```
galvana, line 188: If I'm reading this correctly, we'll only return the first invalid field. I think it would be better if we returned a list of all the invalid fields at once, so the user can fix them all at once instead of removing one then getting a different error.

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -310,3 +338,90 @@ def _unwrap_response_data(saas_request: SaaSRequest, response: Response) -> Any:
...
+            return override_function(
+                update_param_values,
+                policy,
+                privacy_request,
+                privacy_request.get_cached_identity_data(),
+                secrets,
+            )  # type: ignore
+        except Exception:
+            logger.error(
+                "Encountered error executing override mask function '%s",
```
galvana, line 421: Missing closing single-quote

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -310,3 +338,90 @@ def _unwrap_response_data(saas_request: SaaSRequest, response: Response) -> Any:
...
+        """
+        override_function: Callable[
+            ..., Union[List[Row], int]
+        ] = SaaSRequestOverrideFactory.get_override(
+            override_function_name, SaaSRequestType(query_config.action)
+        )
+        try:
+            # if using a saas override, we still need to use the core framework
+            # to generate updated (masked) parameter values, and pass these
+            # into the overriden function
```
galvana, line 405: typo: overridden

src/fidesops/service/saas_request/saas_request_override_factory.py
```diff
@@ -0,0 +1,159 @@
...
+            raise TypeError(
+                "At least one SaaSRequestType must be specified when registering SaaS request override function {name}"
+            )
+
+        def wrapper(
+            override_function: Callable[..., Union[List[Row], int]],
+        ) -> Callable[..., Union[List[Row], int]]:
+            for request_type in request_types:
+                logger.debug(
+                    f"Registering new SaaS request override function '{override_function}' under name '{name}' for SaaSRequestType {request_type}"
```
galvana, line 65: Does a Callable turn into a user-friendly string? I tried it locally and got `<function add at 0x102722ca0>`. Maybe we can use `override_function.__name__`?

src/fidesops/service/saas_request/saas_request_override_factory.py
```diff
@@ -0,0 +1,159 @@
...
+    valid_overrides: Dict[SaaSRequestType, str] = {}
+
+    # initialize each request type's inner dicts with an empty dict
+    for request_type in SaaSRequestType:
+        registry[request_type] = {}
+        valid_overrides[request_type] = ""
+
+    @classmethod
+    def register(
+        cls, name: str, request_types: List[SaaSRequestType]
```
galvana, line 44: Is `request_types` a list for flexibility? I can't think of a use-case for this right now but it's better to have this just in case right?

src/fidesops/service/saas_request/saas_request_override_factory.py
```diff
@@ -0,0 +1,159 @@
...
+        Raises NoSuchSaaSRequestOverrideException if the named override
+        does not exist.
+        """
+        try:
+            override_function: Callable[..., Union[List[Row], int]] = cls.registry[
+                request_type
+            ][override_function_name]
+        except KeyError:
+            raise NoSuchSaaSRequestOverrideException(
+                f"Custom saas override '{override_function_name}' does not exist. Valid custom saas override classes for SaasRequestType {request_type} are [{cls.valid_overrides[request_type]}]"
```
galvana, line 111: typo: SaaS

tests/ops/integration_tests/saas/request_override/test_mailchimp_override_task.py
```diff
@@ -0,0 +1,215 @@
...
+
+from fidesops.graph.graph import DatasetGraph
+from fidesops.models.privacy_request import ExecutionLog, PrivacyRequest
+from fidesops.schemas.redis_cache import PrivacyRequestIdentity
+from fidesops.task import graph_task
+from fidesops.task.graph_task import get_cached_data_for_erasures
+from tests.ops.graph.graph_test_util import assert_rows_match, records_matching_fields
+
+"""
+Integration-level tests to validate the SaaS request override functionality by
```
galvana, line 13: Useful comment! 💯

src/fidesops/service/connectors/saas_connector.py
```diff
@@ -310,3 +338,90 @@ def _unwrap_response_data(saas_request: SaaSRequest, response: Response) -> Any:
...
+                query_config.generate_update_param_values(
+                    row, policy, privacy_request, masking_request
+                )
+                for row in rows
+            ]
+            return override_function(
+                update_param_values,
+                policy,
+                privacy_request,
+                privacy_request.get_cached_identity_data(),
```
galvana, line 416: What are your thoughts on just passing `privacy_request` and getting the identity data if needed from the privacy request?

src/fidesops/service/saas_request/override_implementations/mailchimp_request_overrides.py
```diff
@@ -0,0 +1,136 @@
...
+            processed_data.extend(filtered_data)
+
+    return processed_data
+
+
+@SaaSRequestOverrideFactory.register(
+    "mailchimp_member_update", [SaaSRequestType.UPDATE]
+)
+def mailchimp_member_update(
+    update_param_values: List[Dict[str, Any]],
```
galvana, line 97: The name `update_param_values` is a bit confusing since we have the concept of `param_values`. At first glance the `for update_params in update_param_values` part makes me think that we're iterating through each param_value. I think we need something that better describes this input as a list of param_values per row to update. Maybe param_values_list, or something along those lines?

src/fidesops/service/saas_request/saas_request_override_factory.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Specifically, the validation checks that function's return type is `List[Row]`
+    and that it declares at least 6 parameters.
+    """
+    sig: Signature = signature(f)
+    if sig.return_annotation is not List[Row]:
+        raise InvalidSaaSRequestOverrideException(
+            "Provided SaaS request override function must return a List[Row]"
+        )
+    if len(sig.parameters) < 6:
```
galvana, line 133: What are your thoughts around validating the kwarg names and types? I imagine the signature should be consistent since we're calling the function with fixed inputs. What is some of the flexibility you're referring to in the comment?

src/fidesops/service/saas_request/override_implementations/mailchimp_request_overrides.py
```diff
@@ -0,0 +1,136 @@
...
+from fidesops.models.policy import Policy
+from fidesops.models.privacy_request import PrivacyRequest
+from fidesops.service.saas_request.saas_request_override_factory import (
+    SaaSRequestOverrideFactory,
+    SaaSRequestType,
+)
+from fidesops.util.collection_util import Row
+
+
+@SaaSRequestOverrideFactory.register(
```
galvana, line 19: I like that we can keep all the overrides for a specific connector in a single place 👍

src/fidesops/service/saas_request/override_implementations/mailchimp_request_overrides.py
```diff
@@ -0,0 +1,136 @@
...
+    identity_data: Dict[str, Any],
+    secrets: Dict[str, Any],
+) -> List[Row]:
+    """
+    Equivalent SaaS config for the code in this function.
+
+    Request params still need to be defined for endpoints with overrides.
+    This is to provide the necessary reference and identity data as part
+    of graph traversal. The resulting values are passed in as parameters
+    so we don't need to define the data retrieval here.
```
galvana, line 36: nit: can we add a newline after the description and before the sample config?

src/fidesops/service/saas_request/saas_request_override_factory.py
```diff
@@ -0,0 +1,159 @@
...
+        cls, name: str, request_types: List[SaaSRequestType]
+    ) -> Callable[
+        [Callable[..., Union[List[Row], int]]], Callable[..., Union[List[Row], int]]
+    ]:
+        """
+        Decorator to register the custom-implemented saas request override
+        with the given name.
+        """
+
+        if isinstance(request_types, SaaSRequestType):
```
galvana, line 53: What are lines 53-54 doing?

src/fidesops/service/saas_request/saas_request_override_factory.py
adamsachs: yup. the use case i could imagine was _potentially_ around update/delete? probably wouldn't make sense at this point, but just thought i'd keep things flexible.

src/fidesops/service/connectors/saas_connector.py
adamsachs: yup, i like that - anything to reduce the number of params is better IMO

src/fidesops/service/saas_request/override_implementations/mailchimp_request_overrides.py
adamsachs: hmm, i agree the current name is confusing. i'm not sure `param_values_list` clears things up that much though. maybe just being very explicit and `param_values_per_row` ?

src/fidesops/service/saas_request/saas_request_override_factory.py
adamsachs: ah it's just handling the case where the user passes in a single `SaaSRequestType` by converting it to a list behind the scenes for them. do you know of a more idiomatic way of doing that? is that a bad thing to do in your opinion? it seemed convenient and harmless to me, and also seemed like the sorta thing that people do in python, but i wouldn't take my word for it 😅

src/fidesops/service/saas_request/saas_request_override_factory.py
adamsachs: yeah, i thought back and forth on this a bit, and am definitely not convinced on what is the best approach.

i guess i just don't really see the benefit of preventing an implementer from naming or type-annotating the args however they want in their implementation. whatever we do here in this validation, it's not going to really give them the IDE/static support to automatically generate their methods, or to inform them of what the arguments are, etc. - to get that understanding, they're going to need to refer to examples and/or documentation. so at that point, if they want to for whatever reason adjust the argument names from the example that they're referring to, or e.g. provide a more generic type annotation,  i don't really see why we (the framework) should care at runtime.

i guess what i mean is, with the way things are set up now, you still really have to know what you're doing in order to be able to implement an override here. so it just seems unlikely that they'd end up in a scenario where having their function fail to register because of arg names or type annotations would really help them debug a problem. but maybe i'm not thinking creatively enough.

perhaps extending that argument further would suggest we shouldn't do any sort of validation. i thought these validations made sense because they're really what the _framework_ cares about -- i.e. that it's going to be able to invoke the function successfully, and that it's going to get back from the function what it expects. besides that, we try to just treat the override function as a black box.

in any case, i'd definitely like to talk/think through this one more. i'll leave it unaddressed for the moment.

adamsachs:
noting here that unsafe CI checks succeeded after 2 failed attempts, without any code changes, so there's something fishy going on with the unsafe CI test.

in this case, both failures were on the [same assertion](https://github.com/ethyca/fidesops/blob/main/tests/ops/integration_tests/saas/test_hubspot_task.py#L155) in the hubspot test erasure task -- the assertion that confirms the results of the access request that is accessing the erasure seed data. The same assertion failure occurred on [this workflow run](https://github.com/ethyca/fidesops/runs/7561917134?check_suite_focus=true) yesterday.

i'll split out a separate issue for this

src/fidesops/service/saas_request/override_implementations/mailchimp_request_overrides.py
adamsachs: just noting here that we'll try to import the method such that you can just call `@register`

galvana (approved):
Thanks for making the changes and making the time to discuss some of the other points, this looks good to me!

PR #1040, opened 2022-08-05
Create finished audit log for privacy requests

### LINKED ISSUE #1022: Create `AuditLog` event when requests finish processing
When a `PrivacyRequest` finishes executing successfully, Fidesops should create an `AuditLog` event to mark the same time as `PrivacyRequest.finished_processing_at`

## Questions
- should we also create execution log events for other `PrivacyRequest.status` transitions? If so, maybe this is best implemented as a listener on the `PrivacyRequest` model rather than in the request's execution.

### Notes
* consider putting in auditlog instead with system user.

TheAndrewJackson:
# Purpose
Add `AuditLog` for finished requests so users can easily see when they're done.

# Changes
- Update fideslib and fix small issue with with a case change. ENCODING -> encoding
- New migration adding another enum option to support the updated `AuditLogAction`
- Create `AuditLog` when a subject request finishes

# Checklist

# Ticket

Fixes #1022


src/fidesops/service/privacy_request/request_runner_service.py
```diff
@@ -281,6 +282,15 @@ def run_privacy_request(
...
         privacy_request.finished_processing_at = datetime.utcnow()
+        AuditLog.create(
+            db=session,
+            data={
+                "user_id": "system",
+                "privacy_request_id": privacy_request.id,
+                "action": AuditLogAction.finished,
+                "message": "",
+            },
+        )
```
seanpreston, line 293: Something that would be nice here is to wrap this new `AuditLog` creation into a transaction with `privacy_request.save(...)` — if one of these operations fails without the other we'd end up with an incomplete state, so it's better to commit them both at once.

TheAndrewJackson: Good find. That is a potential issue when approved/denied are created as well. All of the audit logs should probably be updated at the same time when this is done.

Would you like this to be addressed in this PR or a follow up?

seanpreston: Let's do it as a [follow-up](https://github.com/ethyca/fidesops/issues/1057)

seanpreston (approved):

PR #1043, opened 2022-08-05
Removing saas_config.toml

### LINKED ISSUE #1042: Remove dependency on saas_config.toml

We originally used the `saas_config.toml` file to store secrets required for SaaS connectors. We've recently moved our secrets to Vault so keeping this file around is more of a liability. We already have this file in the .gitignore file but since we've force-committed a non-populated version of the file, any local changes are still eligible to be committed.

- Delete the `saas_config.toml` from our repo which will prevent any accidental commits
- Refactor our fixtures to not error out if this file is not present
- Continue to allow users to use the file for local secrets but without the risk of accidentally committing confidential values

galvana:
# Purpose
Removing dependency on `saas_config.toml`

# Changes
- Deleted the saas_config.toml from our repo which will prevent any accidental commits
- Refactored our fixtures to not error out if this file is not present
- Continue to allow users to use the file for local secrets but without the risk of accidentally committing confidential values

# Checklist

# Ticket

Fixes #1042


tests/ops/fixtures/saas_example_fixtures.py
```diff
@@ -223,3 +224,13 @@ def oauth2_connection_config(db: Session, oauth2_configuration) -> Generator:
...
+
+
+@pytest.fixture(scope="session")
+def saas_config() -> Dict[str, Any]:
+    saas_config: Dict[str, Any] = {}
+    try:
+        saas_config: Dict[str, Any] = load_toml(["saas_config.toml"])
+    except FileNotFoundError as e:
+        logger.warning("saas_config.toml could not be loaded: %s", NotPii(e))
+    return saas_config
```
sanders41, line 236: Is there a different way to do this fixture? With `saas_config.toml` in the `.gitignore` people either won't have this file at all, or will have different values. With that any time the fixture is used the test results will be different.

tests/ops/fixtures/saas/auth0_fixtures.py
```diff
@@ -39,10 +32,15 @@ def auth0_secrets():
...
 
 @pytest.fixture(scope="function")
-def auth0_identity_email():
+def auth0_identity_email(saas_config):
     return pydash.get(saas_config, "auth0.identity_email") or secrets["identity_email"]
 
 
+@pytest.fixture(scope="session")
+def auth0_erasure_identity_email():
+    return f"{cryptographic_util.generate_secure_random_string(13)}@email.com"
```
sanders41, line 41: This fixture and the `sendgrid_erasure_identity_email()` fixture in `tests/ops/fixtures/saas/sendgrid_fixtures.py` both create the same session fixture. This fixture could be moved to `conftest.py` with a generic name and shared.

tests/ops/fixtures/saas/sendgrid_fixtures.py
```diff
@@ -20,32 +19,31 @@
...
-def sendgrid_identity_email():
+def sendgrid_identity_email(saas_config):
     return (
         pydash.get(saas_config, "sendgrid.identity_email") or secrets["identity_email"]
     )
 
 
+@pytest.fixture(scope="session")
+def sendgrid_erasure_identity_email():
+    return f"{cryptographic_util.generate_secure_random_string(13)}@email.com"
```
sanders41, line 44: See previous comment about shared fixture.

tests/ops/fixtures/saas_example_fixtures.py
galvana: We store all of our secrets in Vault and are able to pull them down in the Github Actions or locally by using a personal Vault token. We limit write access to Vault so there is a requirement for devs working on SaaS connectors to be able to define secrets locally until we're able to persist them to Vault. I thought about using environment variables for this but I like being able to define the secrets in a single location and having the fixtures load them automatically. With that as context, I'm open to other suggestions if there is a more preferred/standard approach!

sanders41: My thought was environment variables also, and I can think of a way to make that work with a file. But looking at the way the secrets get loaded I'm not sure that helps because it doesn't look like we ever use environment variables for this? Outside of the test suite how do these values get loaded?

I do understand the problem you are trying to solve for here and wouldn't block it. I'm just tying to think if there is a way to reduce the chance it would pass tests in one place and not another causing confusion.

sanders41 (approved):

PR #560, opened 2022-05-25
refactor masking strategy factory for improved extensibility

adamsachs:
# Purpose
Make it easier for developers to implement their own `MaskingStrategy` implementations without requiring updates to core fidesops code

# Changes
- change `MaskingStrategyFactory` to register `MaskingStrategy` implementations with a decorator that is put on `MaskingStrategy` subclasses
- add those decorators to our core `MaskingStrategy` subclasses
- update docs accordingly

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file in appropriate sections
- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch) @ethyca/docs-authors 
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [x] Good unit test/integration test coverage (existing tests should cover us here)
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #547

PR #1163, opened 2022-08-26
unified strategy factory with decorator to register strategies

### LINKED ISSUE #562: Consider refactoring factories for increased extensibility

When attempting to extend the fidesops masking framework outside of the core fidesops repo, it became clear that the existing framework made this difficult to do cleanly - in order to register one's "custom" `MaskingStrategy` subclass in the central `masking_strategy_factory`, one had to make code updates to the `masking_strategy_factory` file/class. https://github.com/ethyca/fidesops/issues/547 (and corresponding PR https://github.com/ethyca/fidesops/pull/560) refactored the `masking_strategy_factory` to provide a more extensible framework.

This feature request is to consider implementing a similar strategy for other factory-like frameworks (is there a more precise term for this?) within `fidesops`, if they may benefit from improved extensibility. 

**Impact**
Will make it much easier to add connectors because currently any changes that have to be done have to be implemented on core and this ticket will allow us to work outside of core.

The decorator-based registration that was implemented for masking strategies in https://github.com/ethyca/fidesops/pull/560 seems to provide an extensible framework that's not burdensome on developers who wish to extend with more implementations (all that's needed is a simple and concise decorator on the extending subclass), while also providing for indexed lookups of subclasses (rather than iterating through a base class's `.__subclasses__()`, for example). The approach also seems relatively transparent and clear to a developer who wants to extend with their own subclass.

Note that with this approach, any extending subclasses will need to be loaded/imported, e.g. via an `__init__.py` -- in the existing frameworks, the import is necessarily done when the subclasses are explicitly referenced in the "registry" enums.

A quick overview of the current `fidesops` codebase points to these frameworks as _potential_ targets of this enhancement:
- authentication strategies (base class: `AuthenticationStrategy`)
- connectors (base class: `BaseConnector`) -- note they're not actually registered in a factory, but `src/fidesops/service/connectors/__init__.py` has constructs that essentially function as a factory
- pagination strategies (base class: `PaginationStrategy`)
- postprocessor strategies (base class: `PostProcessorStrategy`)

In some relatively brief research, I came across some other approaches for implementing the factory pattern with dynamic registries that can be extended without modifying the core factory code:
- Using metaclasses to register subclasses in a central registry (e.g. as suggested by [this SO answer](https://stackoverflow.com/a/5189271)). This may make things a bit easier for the extending developer (i don't think they'd need to do anything beyond just inheriting from the base class), but I think it comes at the expense of making the framework less transparent. But it seems like a viable alternative.
- Using the python builtin `.__subclasses__()` method on the base class to look up a requested subclass. This approach feels the most "built in," but I believe this will require linear look up times (attribute look up on elements in a list). It'd also likely require an extending developer to define a standard method or variable in their subclass that provides a name/id by which the subclass is identified and looked up, which feels clunkier than the decorator-based registration.

### Additional context

- [This article](https://medium.com/@geoffreykoh/implementing-the-factory-pattern-via-dynamic-registry-and-python-decorators-479fc1537bbe) provided some inspiration for the implementation, and it provides a nice walkthrough/example.
- Consider breaking down this work to implement one at a time to identify patterns that are common between and help identify if there are any dependencies/conflicts with other teams work.

Aha! Link: https://ethyca.aha.io/features/OPS-40

adamsachs:
# Purpose
Make our strategies more extensible by using a unified strategy factory that allows for decorator-based registration.

Before this change, many of our strategies (with the exception of `MaskingStrategy`s, which had been refactored in https://github.com/ethyca/fidesops/pull/560) were registered by means of a hardcoded enums in the core `fidesops` codebase. If a developer wanted to implement their own strategy, it required an update to the core `fidesops` codebase.

With this change, developers outside of core `fidesops` can implement their own strategy (whether that's an `AuthenticationStrategy`, `MaskingStrategy`, `PaginationStrategy`, or `PostProcessorStrategy`) and leverage it in the system by simply including the `@register` decorator on their strategy class, and ensuring it has the expected class variables and constructor as specified in the `Strategy` abstract base class.  As an example:

```
from fidesops.ops.schemas.saas.strategy_configuration import StrategyConfiguration
from fidesops.ops.service.pagination.pagination_strategy import PaginationStrategy
from fidesops.ops.service.strategy_factory import register

class SomeStrategyConfiguration(StrategyConfiguration):
            some_key: str = "default value"

@register
class SomePaginationStrategy(PaginationStrategy):
    name = "some strategy"
    configuration_model = SomeStrategyConfiguration

    def __init__(self, configuration: SomeStrategyConfiguration):
        self.some_config = configuration.some_key
        super().__init__(configuration)
```

Additionally, new abstract `Strategy` subtypes (e.g. a `PreProcessorStrategy` module, if the need arises) can be defined and hooked into this framework simply by subclassing the newly added `Strategy` abstract base class. A new factory does not need to be defined.

# Changes
- a unified strategy factory that provides a `register` class decorator that can be used as a hook to dynamically register `Strategy` implementations
    - remove existing `*StrategyFactory` classes as they are no longer needed
- a `Strategy` abstract base class that provides the general specification for what's needed in a `Strategy` implementation, which at this point is just:
     - class variables `name` and `configuration_model` which define the `Strategy` implementations name to be registered under, and pydantic model configuration class to use, respectively
     - a constructor that takes an argument whose type is the pydantic model configuration class specified by the `configuration_model` class variable
 - updates to existing `Strategy` implementations so that they are compatible with this new framework. there should be no user-facing effects of these changes, just rewiring on the backend.

# Checklist

# Ticket

Fixes #562


src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
```diff
@@ -29,16 +27,16 @@ def mask_value(request: MaskingAPIRequest) -> MaskingAPIResponse:
     """Masks the value(s) provided using the provided masking strategy"""
     try:
         values = request.values
-        masking_strategy = request.masking_strategy
-        strategy = MaskingStrategyFactory.get_strategy(
-            masking_strategy.strategy, masking_strategy.configuration
+        masking_strategy_spec = request.masking_strategy
+        masking_strategy: MaskingStrategy = strategy(  # type: ignore
```
adamsachs, line 31: i don't like having to suppress the type checker here, but i couldn't find a better way to get around this. by using a single, unified factory, we lose the ability to strongly type the return type of the `strategy` getter -- since the factory only knows about the abstract `Strategy` class, it must return a `Strategy` here and not a more specific subclass. i couldn't find a clean way for the calling ("client") code to specify what subtype of `Strategy` it was expecting back.

adamsachs:
@ethyca/docs-authors i updated the `Extensibility` section of the "masking strategies" guide page to incorporate the changes in this PR - would appreciate your review on those changes!

i don't _think_ these changes affect any other documentation - i don't believe we've yet documented how a developer can "implement their own" `AuthencationStrategy`, `PaginationStrategy`, or `PostProcessorStrategy`. This change does make it significantly easier for a developer to do so outside of the core fidesops repo, so I wonder if it is worth us creating a backlog ticket to provide some documented guidance on how a developer can do that (similar to what we have in the `Extensibility` section of the masking strategies page). I don't think it's a requirement for this PR, though, since at this point it's still likely going to be internal developers who are creating more strategies, and that knowledge can be shared appropriately internally.

conceptualshark (approved):
Good with the docs updates - agree that a new ticket for docs to cover some of the ways this can be used/extended is a good idea, if you haven't made it yet!

adamsachs:
> Good with the docs updates - agree that a new ticket for docs to cover some of the ways this can be used/extended is a good idea, if you haven't made it yet!

created https://github.com/ethyca/fidesops/issues/1169 to track this doc request. no real priority on it afaik

src/fidesops/ops/service/strategy_factory.py
```diff
@@ -0,0 +1,90 @@
...
+    if not hasattr(strategy_class, "configuration_model"):
+        raise NotImplementedError(
+            "A 'configuration_model' class variable is not defined for this Strategy subclass"
+        )
+
+
+strategy_factory = StrategyFactory()
+register = strategy_factory.register()
+strategy = strategy_factory.strategy
+strategies = strategy_factory.strategies
```
pattisdr, line 90: As you're probably aware, strategies are only available here if they have been imported into a file that was already loaded prior to calling this function.  I would expect "strategies" to include all classes marked by the "register" decorator. 

As an example, if you open up a fidesops shell, start python and import the strategies callable, the strategies are empty.  

Another example, if you look at what strategies are available in the `strategies()` call in `tests/ops/api/v1/endpoints/test_masking_endpoints.py::TestGetMaskingStrategies::test_read_strategies`, it shows 7 while it looks like you have about 15 decorated.  These seven come from strategies imported in application fixtures and in `src/fidesops/ops/service/masking/__init__.py`.

At minimum, I might import the strategies in the individual init files, similar to what was done in fidesops/ops/service/masking/__init__.py, (so add imports to fidesops/ops/service/processors/post_processor_strategy/__init__.py, fidesops/ops/service/pagination/__init__.py, fidesops/ops/service/authentication/__init__.py, etc.).  However, it seems brittle to me to require these imports have happened.

src/fidesops/ops/service/pagination/pagination_strategy_link.py
```diff
@@ -10,24 +10,25 @@
...
+    configuration_model = LinkPaginationConfiguration
+
     def __init__(self, configuration: LinkPaginationConfiguration):
         self.source = configuration.source
         self.rel = configuration.rel
         self.path = configuration.path
-
-    def get_strategy_name(self) -> str:
-        return STRATEGY_NAME
+        super().__init__(configuration)
```
pattisdr, line 31: Generally, it feels like we're trying to force all of these disparate strategies to all be registered in the same way, when they are distinct types of objects.  Their individual configurations are hardly related and I think it would cause confusion for someone trying to create their own strategy.

I might recommend keeping them separate, and allowing a user to creating their own "masking strategy", "pagination strategy", "post processor strategy", or "authentication strategy".

src/fidesops/ops/service/masking/strategy/masking_strategy_string_rewrite.py
```diff
@@ -10,23 +9,26 @@
 )
 from fidesops.ops.service.masking.strategy.format_preservation import FormatPreservation
 from fidesops.ops.service.masking.strategy.masking_strategy import MaskingStrategy
-from fidesops.ops.service.masking.strategy.masking_strategy_factory import (
-    MaskingStrategyFactory,
-)
+from fidesops.ops.service.strategy_factory import register
 
+# this constant is kept around because it is referenced in generated alembic migrations
 STRING_REWRITE_STRATEGY_NAME = "string_rewrite"
```
pattisdr, line 15: Instead of importing this into that migration then, I might just move this string directly into the migration.

src/fidesops/ops/service/strategy_factory.py
```diff
@@ -0,0 +1,90 @@
...
+                        "Strategy with name '%s' already exists. It previously referred to class '%s', but will now refer to '%s'",
+                        name,
+                        self.registry[name],
+                        strategy_class,
+                    )
+                )
+
+            self.registry[name] = strategy_class
+            self.valid_strategies = ", ".join(self.registry.keys())
+            return self.registry[name]
```
pattisdr, line 42: Instead of this register decorator,  you might explore Metaclasses, or the `__subclasses__` method that returns all child classes.  I see you mentioned these as alternatives in the original issue.   I might lean on something more built-in, the register decorator adds an extra step and makes me think on first glance that it's magically going to add the strategy to the registry with its usage alone, when the strategy still needs to be imported to be registered.

src/fidesops/ops/service/strategy_factory.py
pattisdr, line 90: This doesn't feel very Pythonic to me, but if we're exposing it, I'd at least show they are callable types, and rename to something like `get_strategy` and `get_strategies`.

pattisdr (changes requested):
Pausing review here, there are issues in bringing up the webserver itself with this change and generally I favor some of the alternatives you mentioned in the original issue.  I also don't know if all of these strategies should be lumped together.

adamsachs:
closing this PR, as we went with a different approach and opened up https://github.com/ethyca/fidesops/pull/1254 instead.

PR #1166, opened 2022-08-26
SaaS Connector Template Creation Fix: Integer fides_key

pattisdr:
# Purpose

Screen Shot 2022-08-26 at 11 27 22 AM

We want to have the same identifier used across a saas config and dataset config but integer keys were not accounted for. 
 Spotted by @chriscalhoun1974 - he created a saas connection config from a template in testing and gave it an "instance_fides_key" of "111".  This is saved as a  string type on the `DatasetConfig.fides_key` but is interpreted as an integer when parsing the yaml.  This throws a validation error later when you try to edit the dataset, and we see that the datasetconfig.fides_key field does not match the `fides_key` in the saas config.

# Changes
- Use quotes to force a string when replacing <"instance_fides_key"> in the saas config and dataset config yaml.

# Checklist

# Ticket

Unticketed

### DIFF
--- a/src/fidesops/ops/util/saas_util.py
+++ b/src/fidesops/ops/util/saas_util.py
@@ -39,7 +39,7 @@ def load_config_with_replacement(
 ) -> Dict:
     """Loads the saas config from the yaml file and replaces any string with the given value"""
     yaml_str: str = load_yaml_as_string(filename).replace(
-        string_to_replace, replacement
+        string_to_replace, f'"{replacement}"'
     )
     return yaml.safe_load(yaml_str).get("saas_config", [])
 
@@ -55,7 +55,7 @@ def load_dataset_with_replacement(
 ) -> Dict:
     """Loads the dataset from the yaml file and replaces any string with the given value"""
     yaml_str: str = load_yaml_as_string(filename).replace(
-        string_to_replace, replacement
+        string_to_replace, f'"{replacement}"'
     )
     return yaml.safe_load(yaml_str).get("dataset", [])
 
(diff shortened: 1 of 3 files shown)


src/fidesops/ops/util/saas_util.py
```diff
@@ -39,7 +39,7 @@ def load_config_with_replacement(
 ) -> Dict:
     """Loads the saas config from the yaml file and replaces any string with the given value"""
     yaml_str: str = load_yaml_as_string(filename).replace(
-        string_to_replace, replacement
+        string_to_replace, f'"{replacement}"'
```
sanders41, line 42: Is there any chance this causes a different problem? Would `'my "test" string'` replace to `'"my "test" string"'`?

pattisdr: good question. I'm open to better ways to do this, it would be just wrapping if it's an integer type? or a redundant check here that the replacement is a "fides_key" type?

Our current usage of this by the connection template would fail before we got here, because we only let fides_keys have alphanumberic characters, '.', "_" or '-', it wouldn't allow the quotes to go through.

sanders41: Would this work?

```py
yaml_str: str = load_yaml_as_string(filename).replace(
        string_to_replace, str(replacement)
```

pattisdr: No, it was already a string type here "111" so that is added to the yaml but when it is parsed, it comes out as an integer, so you need to add the quotes to force it to be a string.

sanders41 (approved):

PR #1253, opened 2022-09-02
Improved param_value aggregation to support one-to-many relationships

galvana:
# Purpose
Improved param value aggregation to support one-to-many/many-to-many relationships between references and connector params.

# Changes
- Added new fields, validators, and unit tests for `ConnectorParam`
  - `options`: a list of valid values for the connector param values
  - `multiselect`: a boolean which defaults to False indicating that the value is a list of values from the `options` field
- Added a new field to `ParamValue`
  - `unpack`: a boolean which defaults to False indicating that the values must be unpacked from a list/array before being used to generate the requests `[1,2,3] -> 1,2,3` (3 separate requests)
- Updated `SaaSSchema` validator to validate secret values are consistent with the new `option` and `multiselect` connector param fields
- Updated the core logic for generating SaaS requests from input values (references and connector_params)
- Moved SaaS query config logic and tests into their own files

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [x] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [x] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [x] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
- [x] Good unit test/integration test coverage
- [x] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #458

PR #1220, opened 2022-08-30
1106 saas config shopify access endpoints

HamzaWaseemOnBench:
# Purpose

- Purpose of this PR is to add Shopify connector access endpoints

# Changes

- Added config for Shopify Access Endpoints
- Added related dataset and configuration
- Added associated tests and fixtures to validate above functionality

# Checklist
- [ ] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [ ] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [ ] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [ ] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [ ] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [ ] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes # https://github.com/ethyca/fidesops/issues/1106

Issue #1169, opened 2022-08-26
Document how developers can implement their own "strategy"


adamsachs:
### Docs Update Description

Once https://github.com/ethyca/fidesops/pull/1163 is merged, developers outside of the core `fidesops` repo will be able to implement their own `AuthenticationStrategy`, `PaginationStrategy` or `PostProcessorStrategy`, similar to how they can already implement their own `MaskingStrategy`. It would be nice to provide some formal documentation on how this can be done.

### Additional context

The PR description of https://github.com/ethyca/fidesops/pull/1163 contains some notes about how strategies can be implemented - but more detail can be provided if needed.

Note that our documentation already includes an [`Extensibility` section in its masking strategy tutorial](https://ethyca.github.io/fidesops/guides/masking_strategies/#extensibility) - once the above PR is merged, the process should look very similar for the other strategy types mentioned above.

PR #1254, opened 2022-09-02
Refactor strategy instantiation for more extensitiliby

### LINKED ISSUE #562: Consider refactoring factories for increased extensibility

adamsachs:
# Purpose
Make our strategies more extensible by creating a `Strategy` abstract class using the builtin `__subclasses__()` method to find and instantiate Strategy subclases. 

Before this change, many of our strategies (with the exception of MaskingStrategys, which had been refactored in https://github.com/ethyca/fidesops/pull/560) were registered by means of a hardcoded enums in the core fidesops codebase. If a developer wanted to implement their own strategy, it required an update to the core fidesops codebase.

With this change, developers outside of core fidesops can implement their own strategy (whether that's an AuthenticationStrategy, MaskingStrategy, PaginationStrategy, or PostProcessorStrategy) and leverage it in the system by simply importing their subclass and ensuring that it defines a unique `name` class variable, along with a `configuration_model` variable pointing to the strategy's pydantic configuration class. As an example:
```
class SomeStrategyConfiguration(StrategyConfiguration):
    some_key: str = "default value"

class SomeStrategy(PostProcessorStrategy):
    name = "some postprocessor strategy"
    configuration_model = SomeStrategyConfiguration

    def __init__(self, configuration: SomeStrategyConfiguration):
        self.some_config = configuration.some_key

    def process(
        self, data: Any, identity_data: Dict[str, Any] = None
    ) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
        pass
```

# Changes
- an abstract base class `Strategy` that defines logic for strategy retrieval and instantiation, through a generic `get_strategy` method
    - update existing strategy types to inherit from this new base class. 
    - the base class defines standardized class variables `name` and `configuration_model` that are used to identify and instantiate strategy subtypes in a consistent manner
- remove existing strategy factories as they are no longer needed; update references in the codebase to use the new `get_strategy` method for the corresponding `Strategy` subtype

# Checklist

# Ticket

Fixes #562


adamsachs:
the shopify extrnal unsafe integration test is failing but i don't _think_ that's related to my changes?

looks like unsafe tests were never run on the [shopify PR](https://github.com/ethyca/fidesops/pull/1220) that was merged a few days ago, and looks like the unsafe tests have failed all runs since then, besides for @galvana's [draft PR](https://github.com/ethyca/fidesops/pull/1253/files#diff-e04a7b9735af1e45adcb8b439f863df4e746e66cf3ee12739067e020375e8cd3) which actually seems like it is amending the shopify issue.

@galvana is that accurate? if so, then i think we can ignore the failure here...

src/fidesops/ops/migrations/versions/55d61eb8ed12_add_default_policies.py
```diff
@@ -48,6 +45,7 @@
 FIDESOPS_AUTOGENERATED_STORAGE_KEY = "fidesops_autogenerated_storage_destination"
 AUTOGENERATED_ACCESS_KEY = "download"
 AUTOGENERATED_ERASURE_KEY = "delete"
+STRING_REWRITE_STRATEGY_NAME = "string_rewrite"
```
adamsachs, line 48: is it OK to edit this file by hand? i hesitated before doing so since it's a generated migration file...

src/fidesops/ops/service/strategy.py
```diff
@@ -0,0 +1,72 @@
...
+
+from pydantic import ValidationError
+
+from fidesops.ops.common_exceptions import NoSuchStrategyException
+from fidesops.ops.common_exceptions import ValidationError as FidesopsValidationError
+from fidesops.ops.schemas.saas.strategy_configuration import StrategyConfiguration
+
+logger = logging.getLogger(__name__)
+T = TypeVar("T", bound="Strategy")
+C = TypeVar("C", bound=StrategyConfiguration)
```
adamsachs, line 15: i'm a bit unsure on this type var. this was the simplest way i could find to allow subclasses to define their own subtype of `StrategyConfiguration` for their `configuration_model` class var, without getting mypy failures. but i may be overlooking something more straightforward and/or idiomatic.

src/fidesops/ops/service/strategy.py
```diff
@@ -0,0 +1,72 @@
...
+        return cls
+    for sub in cls.__subclasses__():
+        found = _find_strategy_subclass(sub, strategy_name)
+        if found is not None:
+            return found
+    return None
+
+
+def _find_all_strategy_subclasses(cls: Type[T]) -> List[Type[T]]:
+    subs = [cls] if hasattr(cls, "name") else []
```
adamsachs, line 31: i don't love that this creates a bunch of single element lists upon recursion, i feel like it's probably giving unnecessary overhead. an alternative i'm thinking of is just returning the singleton element if there's only one, and then checking for the return type of the method call on recursion. it'd be a little more verbose, but i'm thinking it's probably worth it?

curious of other opinions and/or ideas.

galvana:
@adamsachs, yes, we can ignore the Shopify issues as part of this ticket

src/fidesops/ops/migrations/versions/55d61eb8ed12_add_default_policies.py
galvana: This is fine since it's functionally equivalent, we're just cleaning up the constants.

sanders41:
I believe I found the shopify issue and fixed it as part of #1260 🤞

src/fidesops/ops/migrations/versions/55d61eb8ed12_add_default_policies.py
adamsachs: oh, nice - thanks for clarifying that @pattisdr!

src/fidesops/ops/service/strategy.py
```diff
@@ -0,0 +1,77 @@
...
+
+def _find_all_strategy_subclasses(
+    cls: Type[T], subs: List[Type[T]] = None
+) -> List[Type[T]]:
+    if not subs:
+        subs = []
+    if hasattr(cls, "name"):
+        subs.append(cls)
+    for sub in cls.__subclasses__():
+        _find_all_strategy_subclasses(sub, subs)
```
pattisdr, line 38: Your use of `extend` probably worked better from the previous commit, with this change, the return is prematurely ending the `for` loop.

src/fidesops/ops/service/strategy.py
```diff
@@ -0,0 +1,77 @@
...
+        Returns the strategy given the name and configuration.
+        Raises NoSuchStrategyException if the strategy does not exist
+        """
+
+        strategy_class = _find_strategy_subclass(cls, strategy_name)
+
+        if strategy_class is None:
+            valid_strategies = ", ".join(
+                [sub.name for sub in _find_all_strategy_subclasses(cls)]
+            )
```
pattisdr, line 64: Do you have an example of when `get_strategy` would not find the strategy class, and we'd have to resort to calling `_find_all_strategy_subclasses`?   Since `_find_all_strategy_subclasses` returns an empty list on your latest commit, I don't think we have tests that cover this use case.

src/fidesops/ops/service/strategy.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _find_strategy_subclass(
+    cls: Type[Strategy], strategy_name: str
+) -> Optional[Type[Strategy]]:
+    if hasattr(cls, "name") and cls.name == strategy_name:
+        return cls
+    for sub in cls.__subclasses__():
+        found = _find_strategy_subclass(sub, strategy_name)
+        if found is not None:
```
pattisdr, line 25: I might do `if found` instead since I don't think the None comparison is needed specifically?

pattisdr (changes requested):
Nice work, main thing is fixing up `_find_all_strategy_subclasses` - the latest commit broke this, and adding tests around it.  I do like this implementation generally.

src/fidesops/ops/service/strategy.py
adamsachs: good point - `test_retrieve_nonexistent_strategy` does hit that code path, but it wasn't testing for the string populated by `_find_all_strategy_subclasses`, so i'll update it to verify that string is being populated correctly.

tests/ops/api/v1/endpoints/test_masking_endpoints.py
```diff
@@ -7,33 +7,31 @@
...
 
 
 class TestGetMaskingStrategies:
     def test_read_strategies(self, api_client: TestClient):
         expected_response = []
-        for strategy in MaskingStrategyFactory.get_strategies():
+        for strategy in MaskingStrategy.get_strategies():
             expected_response.append(strategy.get_description())
 
         response = api_client.get(V1_URL_PREFIX + MASKING_STRATEGY)
```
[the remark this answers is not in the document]
adamsachs: yeah, i've been a bit uncertain about this test since i first saw it. but i'm also not really sure on the best way to resolve it - i feel like any approach to getting a "true" list of strategies is either going to rely on duplicating the logic of `get_strategies()`, or it's going to be prone to false negatives if/when more masking strategies are added to the testing runtime (whether that's core fidesops updates or, potentially, some extended runtime like `-plus`).

obviously we need more robust testing for `get_strategies()`, as you've correctly pointed out. but what do you think about keeping this as is, and just focusing on firming up the tests around `get_strategies()` itself?

src/fidesops/ops/service/strategy.py
adamsachs: thank you for catching this!

it actually seems the issue was the `if not subs:` conditional - that evaluates to `True` in the case where `subs` is an empty list, and i was assuming it would evaluate to `False` (whoops). my next commit will update the conditional to `if subs is None:`, and this seems to be working as expected locally.

if you still think the `extend` implementation is better, than i'm happy to change it back. i think with the current approach here we avoid a bunch of overhead in creating temporary lists on each subcall in the recursion, and i also find this a bit more readable.

adamsachs:
@ethyca/docs-authors would you be able to take a quick look at the docs changes here? this PR is instead of https://github.com/ethyca/fidesops/pull/1163, as we decided on a different implementation approach. let me know if you've got suggestions.

i've also tweaked the description of related docs ticket https://github.com/ethyca/fidesops/issues/1169 to reference this PR, rather than the now outdated #1163

src/fidesops/ops/service/strategy.py
pattisdr: Ah that makes sense, this implementation is fine!

tests/ops/api/v1/endpoints/test_masking_endpoints.py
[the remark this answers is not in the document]
pattisdr: The get_strategies() improvement was the main thing, but perhaps asserting that the response is non-empty at least?

src/fidesops/ops/service/strategy.py
pattisdr: ah thanks!

pattisdr (approved):
Good work @adamsachs 🏆 

A small improvement might be adding https://github.com/ethyca/fidesops/pull/1254#discussion_r964869998, but otherwise this looks good to me.  

I'll let your team take care of merging in case there's more left to do -

docs/fidesops/docs/guides/masking_strategies.md
```diff
@@ -241,6 +245,4 @@ any defaults that should be applied in their absence. All configuration classes
 
 ### Integrate the masking strategy factory
 
-In order to leverage an implemented masking strategy, the `MaskingStrategy` subclass must be registered with the `MaskingStrategyFactory`. To register a new `MaskingStrategy`, use the `register` decorator on the `MaskingStrategy` subclass definition, as shown in the above example.
-
-The value passed as the argument to the decorator must be the registered name of the `MaskingStrategy` subclass. This is the same value defined by [callers](#using-fidesops-as-a-masking-service) in the `"masking_strategy"."strategy"` field.
+In order to leverage an implemented masking strategy, the `MaskingStrategy` subclass must be imported into the application runtime. Also, the `MaskingStrategy` class must define two class variables: `name`, which is the unique, registered name that callers [callers](#using-fidesops-as-a-masking-service) will use in their `"masking_strategy"."strategy"` field to invoke the strategy; and `configuration_model`, which references the configuration class used to parameterize the strategy.
```
conceptualshark, line 248:
```suggestion
In order to leverage an implemented masking strategy, the `MaskingStrategy` subclass must be imported into the application runtime. Also, the `MaskingStrategy` class must define two class variables: `name`, which is the unique, registered name that [callers](#using-fidesops-as-a-masking-service) will use in their `"masking_strategy"."strategy"` field to invoke the strategy; and `configuration_model`, which references the configuration class used to parameterize the strategy.
```

conceptualshark (commented):
Just a little tweak - otherwise this looks good!

galvana (approved):
I like this approach! I also like how there's a distinction between the different strategy "types" instead of just one combined pool of strategies.

PR #1274, opened 2022-09-08
Update EmailRequestFulfillmentBodyParams Type for Easier Coercion

pattisdr:
# Purpose

Make EmailRequestFulfillmentBodyParams a Pydantic schema instead of just a python type

# Changes

Instead of passing around EmailRequestFulfillmentBodyParams which is currently type: `Dict[CollectionAddress, Optional[CheckpointActionRequired]]`, update the type to a list of CheckpointActionRequired. 

so: 

`List[CheckpointActionRequired]`


seanpreston:
Thanks for making this change!

PR #1281, opened 2022-09-08
Add `days_left` field

### LINKED ISSUE #1148: Add `days_left` to PrivacyRequest API
We want to display the number of days that are left to action a given PrivacyRequest in the API. Given this number changes with time, we shouldn't store it in the DB directly for now, but calculate it at API level based on the timedelta between now and when the request is due.

This field should be optional, since `Policy.execution_timeframe` is also optional.

TheAndrewJackson:
# Purpose
Add dynamic `days_left` field that is generated based off of the current `due_date` value of a privacy request.

# Changes
- Add new property to `PrivacyRequest` model
- Add new tests and update existing tests to work with new update

# Checklist

# Ticket

Fixes #1148


src/fidesops/ops/models/privacy_request.py
```diff
@@ -138,6 +138,14 @@ def generate_request_callback_jwe(webhook: PolicyPreWebhook) -> str:
     return generate_jwe(json.dumps(jwe.dict()), config.security.app_encryption_key)
 
 
+def get_days_left(self: PrivacyRequest) -> Union[int, None]:
+    if self.due_date is None:
+        return None
+
+    delta = self.due_date.date() - datetime.now(timezone.utc).date()
+    return delta.days
```
sanders41, line 146: I would rename `self` here. `self` makes it look like it is part of the class but it isn't so I initially thought it wasn't indented correctly.

You could also slightly simplify the utc date with `utcnow`.

```suggestion
def get_days_left(privacy_request: PrivacyRequest) -> Union[int, None]:
    if privacy_request.due_date is None:
        return None

    delta = privacy_request.due_date.date() - datetime.utcnow().date()
    return delta.days
```

src/fidesops/ops/models/privacy_request.py
```diff
@@ -212,6 +220,7 @@ class PrivacyRequest(Base):  # pylint: disable=R0904
     paused_at = Column(DateTime(timezone=True), nullable=True)
     identity_verified_at = Column(DateTime(timezone=True), nullable=True)
     due_date = Column(DateTime(timezone=True), nullable=True)
+    days_left = property(get_days_left)
```
sanders41, line 223: I am guessing it has something to do with sqlalchemy being unhappy, but is there a reason this was defined as a property this way instead of:

```py
@property
def days_left(self)...
```

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1322,6 +1334,41 @@ def test_get_failed_request_resume_info_from_email_send(
         }
         assert data["resume_endpoint"] == f"/privacy-request/{privacy_request.id}/retry"
 
+    @pytest.mark.parametrize(
```
sanders41, line 1337: Making use of the parametrization, I like it!

src/fidesops/ops/models/privacy_request.py
TheAndrewJackson: There isn't a specific reason. I'll try this out and see if it works. The `property` function/decorator is newer to me so I didn't realize I could do it that way.

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
TheAndrewJackson: Thanks! I've been looking for opportunities to use it

TheAndrewJackson:
@sanders41 I updated the `days_left` field to use the `property` decorator and moved the function body inside the model. Since the function is inside the model now I left the parameter as self. If that doesn't make sense for the `property` decorator I can push up a quick update changing it

sanders41:
The property update looks good. I'll approve and merge as soon as CI finishes.

sanders41:
@TheAndrewJackson one thing I just realized, this updates a database model. Do we need a migration?

TheAndrewJackson:
@sanders41 AFAIK we don't need a migration. We decided to calculate this property at the api level because it's a value that changes every day and there could be many changes if the org has a lot of privacy requests. 

If at a later date we decided to move this property into the DB and have an automated job that update all the due dates we would need a migration.

sanders41 (approved):

PR #1282, opened 2022-09-08
1156 email upon privacy request completion

### LINKED ISSUE #1156: Email notification acknowledging PrivacyRequest completed
Add the `FIDESOPS__NOTIFICATIONS__SEND_REQUEST_COMPLETION_NOTIFICATION` config var and default it to `True`.

If `FIDESOPS__NOTIFICATIONS__SEND_REQUEST_COMPLETION_NOTIFICATION` is `True` then send an email acknowledging that the data subject's privacy request has completed processing successfully.

If the request is an access request, it should include a link to the uploaded data package.

## Questions
- what should the email copy read?
  - we should use placeholders for this during dev
- can we always provide a secure link via S3 only accessible for the user who has it?
  - we always have the URL but it's TBD on whether it's signed — so there could be complexity here
  - can we add restrictions to number of times available for download? — would they be settable from Fidesops' code or only at the bucket level?

eastandwestwind:
# Purpose
Send email upon privacy request completion, including download link(s) if applicable.

# Changes
- Adds 2 new email templates, one for access completion, one for deletion completion
- Updates s3 upload to generate 1 time download link that respects TTL
- Adds new config variables, one for whether to send completion email, one for download link TTL
- Updates privacy request runner to send notification email upon completion of request
- Update documentation

# Checklist

# Ticket

Fixes #1156


src/fidesops/ops/service/email/email_dispatch_service.py
```diff
@@ -26,10 +24,7 @@ def dispatch_email(
     db: Session,
     action_type: EmailActionType,
     to_email: Optional[str],
-    email_body_params: Union[
-        SubjectIdentityVerificationBodyParams,
-        EmailRequestFulfillmentBodyParams,
-    ],
+    email_body_params: Any,
```
eastandwestwind, line 27: still having issues with mypy errors here

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +355,58 @@ async def run_privacy_request(
...
+    if is_access_request:
+        if not identity_data.get(ProvidedIdentityType.email.value):
+            # fatal bc user will not get results
+            raise IdentityNotFoundException("Identity email was not found, so access request email could not be sent.")
+        dispatch_email(
+            db=session,
+            action_type=EmailActionType.PRIVACY_REQUEST_COMPLETE_ACCESS,
+            to_email=identity_data.get(ProvidedIdentityType.email.value),
+            email_body_params=AccessRequestCompleteBodyParams(
+                download_links=access_result_urls
```
eastandwestwind, line 395: To remain consistent with how we perform local upload, we still send a download link if data package contains no data

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -123,11 +129,11 @@ def upload_access_results(
     access_result: Dict[str, List[Row]],
     dataset_graph: DatasetGraph,
     privacy_request: PrivacyRequest,
-) -> None:
+) -> List[str]:
     """Process the data uploads after the access portion of the privacy request has completed"""
+    download_urls: List[str] = []
```
eastandwestwind, line 134: It's possible for multiple rules in policy to each point to an s3 upload location (and in the future, may include other non-s3 upload locations), so we support returning multiple download links

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +355,58 @@ async def run_privacy_request(
...
+            db=session,
+            action_type=EmailActionType.PRIVACY_REQUEST_COMPLETE_ACCESS,
+            to_email=identity_data.get(ProvidedIdentityType.email.value),
+            email_body_params=AccessRequestCompleteBodyParams(
+                download_links=access_result_urls
+            ),
+        )
+    else:
+        if not identity_data.get(ProvidedIdentityType.email.value):
+            # not fatal bc data was deleted as requested
```
eastandwestwind, line 400: I don't consider this fatal, so do not consider this an errored privacy request, but open to handling this as we do access requests here https://github.com/ethyca/fidesops/pull/1282/files#diff-3d5f0b1a3f488cadb9d63aea80c3cae154b472a3a6f066d5963b76c841d1756cR388

tests/ops/fixtures/application_fixtures.py
```diff
@@ -810,10 +810,13 @@ def _create_privacy_request_for_policy(
         db=db,
         data=data,
     )
+    email_identity = "test@example.com"
+    identity_kwargs = {"email": email_identity}
+    pr.cache_identity(identity_kwargs)
```
eastandwestwind, line 815: need to cache identity so that we can retrieve it to send email

eastandwestwind:
Access request completion email:
Screen Shot 2022-09-08 at 11 44 08 AM

Deletion request completion email:

conceptualshark (approved):
👍 Docs updates look good! I like the email templates - thanks for handling these!

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +354,70 @@ async def run_privacy_request(
         )
         if not proceed:
             return
-
+        if config.notifications.send_request_completion_notification:
+            try:
+                initiate_privacy_request_completion_email(
+                    session, policy, access_result_urls, identity_data
+                )
```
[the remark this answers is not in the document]
pattisdr: We also need to decide what happens if the email doesn't send and we put the PrivacyRequest in an error state (in the EmailDispatchException case), what is the best course of action?  

Do they submit a new PrivacyRequest? (If there are erasure rules defined it is possible that this request can't be re-run)

Do they restart from this point? To be able to restart from failure, you need to cache a checkpoint:  `privacy_request.cache_failed_checkpoint_details(
step=CurrentStep.post_notification_send(new checkpoint), collection=None)` so fidesops knows where to resume the request.  Is there any way to be able to rebuild the S3 link in the case of an access request? For an erasure, we could just attempt to resend just the email.

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +354,70 @@ async def run_privacy_request(
...
+            except IdentityNotFoundException as e:
```
pattisdr, line 362: If there was no email supplied originally, even if we set it up so a fidesops admin can restart this privacy request (by caching a checkpoint), it doesn't change the fact that we don't have an email at all.  These will just stay in a failure state I assume? It seems like there's no way to correct.

tests/ops/service/privacy_request/request_runner_service_test.py
```diff
@@ -1773,3 +1836,114 @@ def test_email_connector_no_updates_needed(
...
+
+            pr = get_privacy_request_results(
+                db,
+                policy,
+                run_privacy_request_task,
+                data,
+            )
+            pr.delete(db=db)
+
+        assert mailgun_send.called is False
```
pattisdr, line 1949: After changes noted above are completed to handle this EmailDispatchException, I'd also verify the PrivacyRequest is put into error status.

docs/fidesops/docs/guides/privacy_requests.md
```diff
@@ -54,6 +54,17 @@ to continue privacy request execution.  Until the Privacy Request identity is ve
...
+
+By default, a request completion email will be sent to users, along with a link to download their data, if applicable. To change this behavior, set the `send_request_completion_notification`
+variable in your `fidesops.toml`. You must also set up an [EmailConfig](./email_communications.md) that lets fidesops send automated emails
+to your users.
+
+!!! Note
+For security purposes, the data package download link is a one-time link and expires in 24 hrs by default. To change TTL, update the `subject_request_download_link_ttl_seconds`
+variable in your `fidesops.toml`.
+
+
```
pattisdr, line 67: Nice docs here

fidesops.toml
```diff
@@ -44,3 +45,6 @@ analytics_opt_out = false
 
 [admin_ui]
 enabled = true
+
+[notifications]
+send_request_completion_notification = true
```
pattisdr, line 50: Do we need an easier way to disable this during local dev? We don't want to be sending random emails unintentionally.  It seems like in dev mode the default should be to not send.

src/fidesops/ops/service/email/email_dispatch_service.py
```diff
@@ -26,10 +24,7 @@ def dispatch_email(
     db: Session,
```
pattisdr, line 24: Nice work on this broader email service you implemented earlier; it's worked well for all of the use cases we've added lately.

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +354,70 @@ async def run_privacy_request(
...
         logging.info("Privacy request %s run completed.", privacy_request.id)
+        privacy_request.save(db=session)
+
+
+def initiate_privacy_request_completion_email(
+    session: Session,
+    policy: Policy,
+    access_result_urls: List[str],
+    identity_data: Dict[str, Any],
+) -> None:
```
pattisdr, line 392: Can you add a docstring to this function?

tests/ops/service/privacy_request/request_runner_service_test.py
```diff
@@ -1773,3 +1836,114 @@ def test_email_connector_no_updates_needed(
...
+
+        pr = get_privacy_request_results(
+            db,
+            policy,
+            run_privacy_request_task,
+            data,
+        )
+        pr.delete(db=db)
+
+        mailgun_send.assert_called_once()
```
pattisdr, line 1911: Other things that might be worth testing are when there's both an access and erasure policy defined, that multiple emails are sent. 

You might also assert the arguments that the access versus erasure mailgun send were callled with

pattisdr (changes requested):
good work @eastandwestwind, the main thing I'd like to see addressed here is what happens when the email send fails and the path we lay out to allow this to be retried.

fidesops.toml
eastandwestwind: It makes sense to instead default to false in our own `fidesops.toml`, so I'll update!

src/fidesops/ops/service/privacy_request/request_runner_service.py
eastandwestwind: Right, so this is the edge case where, for example,  the original identity input would be only a phone number (no email), and so the email notification cannot send. Right now, this shouldn't ever happen, since we always require an email in privacy center. But it's possible a user could build their own privacy center and call fidesops with only a phone number. 

My thinking is at least adding to docs that it's required for a privacy center to intake email identity for email notifications to function properly.

src/fidesops/ops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
eastandwestwind: To help answer this, let's talk about 2 likely issues with email send:

1. Something is misconfigured with the user's email server, e.g. proper domain sending validation / MX records haven't been set up.
2. User email domain rejects email for some reason, or isn't valid

I think it makes sense, in case 1, to allow for restarting from failure. We can probably regenerate a new signed URL (which renews the expiry) as long as we know the privacy request ID. I'll add this in.

For case 2, I don't think anything can be done about an invalid email domain, bc we can't email them to submit a new request.

Something I want to add in the future is implementing retries upon failure, regardless of reason.

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -343,20 +354,70 @@ async def run_privacy_request(
...
+    if policy.get_rules_for_action(action_type=ActionType.access):
```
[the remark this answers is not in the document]
eastandwestwind: got it, so to clarify, how will I check for any `manual_data` within the `request_runner_service.py`?

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -145,12 +152,14 @@ def upload_access_results(
             privacy_request.id,
         )
         try:
-            upload(
+            download_url: Optional[str] = upload(
                 db=session,
                 request_id=privacy_request.id,
                 data=filtered_results,
                 storage_key=rule.storage_destination.key,  # type: ignore
             )
```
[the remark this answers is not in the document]
eastandwestwind: good catch, I missed this. I'll see what I can do with returning something like `"your local fides_uploads folder"` for local upload

src/fidesops/ops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
pattisdr: nothing to do here since my work isn't merged, it might be me that reconciles, but there will be a `manual_data` dictionary created inside `run_privacy_requeset`, so you could just pass that into this same function alongside the access_result_urls.  The manual_data also only exists for access requests.

src/fidesops/ops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
eastandwestwind: Had a chat with @seanpreston and it's OK to bypass handling these cases for now. Also a note that once his Celery work is merged, we can have Celery handle retries with a simple annotation.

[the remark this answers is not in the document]
pattisdr: OK I'd still catch the error and put it the PrivacyRequest in a failed state, instead of leaving it as in-processing, even if we don't allow this to be retried.

src/fidesops/ops/service/privacy_request/request_runner_service.py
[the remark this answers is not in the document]
eastandwestwind: Makes sense, thanks for the heads up!

src/fidesops/ops/service/privacy_request/request_runner_service.py
eastandwestwind: I've chatted with Sean about this also, and he wishes to consider this an errored request, since regulatory practices require that the subject be notified upon completion of requests.

src/fidesops/ops/service/email/email_dispatch_service.py
eastandwestwind: this makes me so happy to hear!

eastandwestwind:
Thanks for the review @pattisdr ! Just working through some strange issues I'm seeing with pytest not recognizing my `privacy_request_complete_email_notification_enabled` fixture- https://github.com/ethyca/fidesops/runs/8278916771?check_suite_focus=true#step:6:1092. I tried moving the fixture local to the test file, but that didn't work. Have you run into this before?

pattisdr:
yeah @eastandwestwind it's tricky with the autouse fixtures.  With the similar `subject_identity_verification_not_required` autouse fixture, I created a fixture of the same name but just at the function level defined within the class I was testing to override the original.

src/fidesops/ops/service/privacy_request/request_runner_service.py
pattisdr: Thanks for the parameter descriptions, that's helpful too, but I was mainly requesting a one-sentence description of this function's purpose!

pattisdr (approved):
great work here @eastandwestwind 

Noting for future work, if the email send fails here, currently this request will be put in an "error" state and cannot be retried because we don't cache a point from which to retry yet.

PR #1284, opened 2022-09-08
Sort privacy requests by `due_date`

### LINKED ISSUE #1151: Ability to sort `PrivacyRequest`s based on `PrivacyRequest.due_date`
The `get_request_status` API view doesn't currently support sorting based on dynamic fields, opting instead to order all fields by `created_at.desc()`. This increment is to support sorting of PrivacyRequests by dynamic fields passed in via query params.

Specifically important for this epic is to be able to sort PrivacyRequests into descending order based on their `due_date` fields. This will ensure the most urgent are able to be displayed at the top of the table in the UI, and enable us to show the `PrivacyRequest`s in order of those with the fewest `days_left` (calculated via the API) at the top of the table.

## Questions
- How does SQLAlchemy's default working in the case where `PrivacyRequest.due_date` is null?
  - We could give the due_date a field very far into the future in cases where it's null

TheAndrewJackson:
# Purpose
Add ability to sort privacy requests by `due_date`

# Changes
- Update `get_request_status` to have new `sort_due_date` query param

# Checklist

# Ticket

Fixes #1151


src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -483,7 +485,19 @@ def _filter_privacy_request_queryset(
...
+def _sort_privacy_request_queryset(
+    query: Query, sort_due_date: Optional[ColumnSort]
+) -> Query:
+    if sort_due_date is None:
+        return query.order_by(PrivacyRequest.created_at.desc())
+
+    if sort_due_date is ColumnSort.ASC:
+        return query.order_by(nullslast(PrivacyRequest.due_date.asc()))
+
+    return query.order_by(nullslast(PrivacyRequest.due_date.desc()))
```
seanpreston, line 500: This function might be better off as:

```
def _sort_privacy_request_queryset(
    query: Query,
    sort_field: str = "created_at",
    sort_direction: str = "desc",
) -> Query:
    sort_object_attribute = getattr(PrivacyRequest, sort_field)
    sort_func = getattr(sort_object_attribute, sort_direction)
    return query.order_by(nullslast(sort_func()))
```

We could add some error handling or Enum support to mitigate the brittle nature of the string matching, and then we have a sort function we can extend to any field and direction.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -558,6 +572,7 @@ def get_request_status(
     verbose: Optional[bool] = False,
     include_identities: Optional[bool] = False,
     download_csv: Optional[bool] = False,
+    sort_due_date: Optional[ColumnSort] = None,
```
seanpreston, line 575: This could be switched for `sort_field` and `sort_direction`

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -583,6 +599,7 @@ def get_request_status(
         errored_gt,
         external_id,
     )
+    query = _sort_privacy_request_queryset(query, sort_due_date)
```
seanpreston, line 602: A line logging `logger.info("Sorting requests by field: %s and direction: %s" % (sort_field, sort_direction))` would be good

seanpreston (changes requested):
Nice work! Just a couple of small tweaks to consider

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
TheAndrewJackson: Great idea! I implemented this and added an additional validation that checks if the model has the request `sort_field`. The server returns a 500 error otherwise when trying to sort on a non existent field.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
TheAndrewJackson: I implemented this using the `ColumnSort` enum that I made. This makes it so fastapi/pydantic validate that only `asc` and `desc` are being used at options for `sort_direction`

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -489,15 +489,17 @@ def _filter_privacy_request_queryset(
 
 
 def _sort_privacy_request_queryset(
-    query: Query, sort_due_date: Optional[ColumnSort]
+    query: Query, sort_field: str, sort_direction: ColumnSort
```
seanpreston, line 492: I _think_ `sort_direction` might need to be `str` for this to work? Does `ColumnSort` expect a function?

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1369,6 +1369,59 @@ def test_get_privacy_requests_sets_days_left(
...
+                    "identity": {"email": "test@example.com"},
+                }
+            )
+            days_left_values.append(i + policy.execution_timeframe)
+
+        api_client.post(url, json=data)
+
+        auth_header = generate_auth_header(scopes=[PRIVACY_REQUEST_READ])
+        resp = api_client.get(
+            f"{url}?sort_direction=asc&sort_field=due_date",
```
seanpreston, line 1402: Does this conform to the spec of the Enum?

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -1369,6 +1369,59 @@ def test_get_privacy_requests_sets_days_left(
...
+            data.append(
+                {
+                    "requested_at": str(requested_at),
+                    "policy_key": policy.key,
+                    "identity": {"email": "test@example.com"},
+                }
+            )
+            days_left_values.append(i + policy.execution_timeframe)
+
+        api_client.post(url, json=data)
```
seanpreston, line 1398: It's worth noting that this test won't differentiate effectively whether you're sorting on `due_date` or `created_at`, since they're effectively the same order (created_at descending will return the same order as due_date descending).

Adding an element of randomness to the due date would fix that.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
TheAndrewJackson: It looks like it's working without using `.value`. That might be because I add the `str` type as a mixin
```python
class ColumnSort(str, Enum):
    DESC = "desc"
    ASC = "asc"
```

I'm okay with switching it to `sort_direction.value` to help it be more clear. It's a little confusing when python enums require using the `.value` in some cases and not others. I'll update it now

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
TheAndrewJackson: Yes, it does. The value that is passed in for `sort_direction` is validated by fastapi, or possibly pydantic, to ensure that it is one of the enumerations.

TheAndrewJackson: Here is what's return if an invalid option is used
```json
{
    "detail": [
        {
            "loc": [
                "query",
                "sort_direction"
            ],
            "msg": "value is not a valid enumeration member; permitted: 'desc', 'asc'",
            "type": "type_error.enum",
            "ctx": {
                "enum_values": [
                    "desc",
                    "asc"
                ]
            }
        }
    ]
}
```

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
TheAndrewJackson: True, good find. I'll update the test to not be the same as `created_at`

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
seanpreston: Ah great, that explains it, I didn't notice that

TheAndrewJackson: It looks like that may not work. The return type of the enum value property is `Any`. The type checker throws this error for `getattr`: `Expected type 'str', got '() -> Any`.

I was able to make the type checker happy by overloading the `value` property with one that has the correct type hints
```python
class ColumnSort(str, Enum):
    DESC = "desc"
    ASC = "asc"

    @property
    def value(self) -> str:
        return self._value_
```
This _should_ be okay but I'm not 100% sure if it's worth overloading the property. What do you think @seanpreston?


Also, I double checked and the `str` mixin is what makes it work without the `.value`. Here is the error that is displayed when the mixin is removed: `Expected type 'str', got 'ColumnSort' instead `. The mixin must automatically convert it to a `str`

seanpreston (approved):

PR #1285, opened 2022-09-09
Enable Manual Webhooks in Request Execution [#1228]

### LINKED ISSUE #1228: Backend - Add manual webhook execution

❗ Dependent on #1224 

Add API endpoints and support for manual webhooks in the execution layer.

#### Temporarily caching manual data received
- Add a method to the PrivacyRequest model that caches manual data received for a PolicyManualWebhook, including the `privacy_request_id`, the `policy_manual_webhook_id`, and a dictionary of fields/values received.
- Add an endpoint that lets you save data to the cache for a particular privacy request id and policy_manual_webhook with the fields values configured on the PolicyManualWebhook.
  - The request must be a in a "paused" state.  All this endpoint does is add data to the cache, it doesn't queue the privacy request for processing again.

#### Execution
- First thing in privacy execution, add a check that looks up the PolicyManualWebhooks in the database and compares them with the data received from the PolicyManualWebhooks in the cache for the given `privacy_request_id`.  If we don't have entries for all PolicyManualWebhooks, pause request execution.  
- Add a separate endpoint for resuming privacy request execution after all the data from the manual webhooks has been received:
  - Request must be in "paused" mode.
  - We at least must have entries in the cache for all policy_manual_webhooks, there doesn't need to be data, but we need confirmation that all have them have been approved in that the keys exist in the cache
  - Sets request to `in_processing`, and queues the privacy request again from the "access step".
-  Have `upload_access_results` separately look up if there's any input data from any `PolicyManualWebhooks` in the cache and adds those directly to the filtered results with each key as the connection_config key.  Unlike regular access results, these fields are just passed directly through, we don't filter on data category.

### Additional context

pattisdr:
👉  Note that new scopes have been added here.

# Purpose

Enable the new "Manual Webhook" Connector.  In short, if manual webhooks are configured, the privacy request will be in a state of `requires_input` until data has been supplied for each webhook.  Once the privacy request is resumed, the graph will run as usual and any manually uploaded data will be directly added to the final data package.  The data is not used as part of the graph and is not filtered by data category.  All uploaded data is passed to the user as-is.

# Changes
- Add four new API endpoints
   - Get all access manual webhooks `GET "/access_manual_webhook"`
   - Upload data for a manual webhook on a given privacy request: `PATCH /privacy-request/{privacy_request_id}/access_manual_webhook/{connection_key}`
     - There is no webhook key since there's only one AccessManualWebhook per ConnectionConfig of type `manual_webhook` so the connection_key is for the ConnectionConfig.
   - Resume privacy request in `requires_input` state once all webhook data is uploaded: `POST "/privacy-request/{privacy_request_id}/resume_from_requires_input"`
   - Get manually uploaded data for a given webhook `GET /privacy-request/{privacy_request_id}/access_manual_webhook/{connection_key}`
- Update privacy request execution to accommodate new manual webhooks.  Once all data has been uploaded, proceed with request execution, and then take the manual data and add it directly to the final upload package.

# Checklist

# Ticket

Fixes #1228


src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -233,6 +266,12 @@ async def run_privacy_request(
         logging.info("Dispatching privacy request %s", privacy_request.id)
         privacy_request.start_processing(session)
 
+        manual_data, proceed = get_access_manual_webhook_inputs(
+            session, privacy_request
+        )
+        if not proceed:
+            return
```
pattisdr, line 273: This is the primary change of this PR - this puts a PrivacyRequest in `requires_input` status if we don't have every manual webhook confirmed.  If we do have confirmation from all our manual webhooks, we retrieve any manually uploaded data, and this is passed onto the `upload_access_results` function and data is combined with the automatically retrieved data at the end.

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -60,6 +62,30 @@
...
+            manual_inputs[manual_webhook.connection_config.key] = [
+                privacy_request.get_manual_webhook_input(manual_webhook)
+            ]
+    except (MissingManualWebhookData, ValidationError) as exc:
+        logger.info(exc)
+        privacy_request.status = PrivacyRequestStatus.requires_input
+        privacy_request.save(db)
+        return manual_inputs, False
+
+    return manual_inputs, True
```
pattisdr, line 86: Normally the access results have keys in the format "dataset:collection", so here I am using the `manual_webhook` connectionconfig key instead, since there is no "dataset" or "collection" defined for the webhook results.   I think an overlap with existing keys is unlikely since they are mashups of dataset and collection names.

As an aside, since ConnectionConfigs and AccessManualWebhooks have a 1:1 relationship, I am not giving a separate identifier to the AccessManualWebhook, and using its ConnectionConfig key _everywhere_. So the question is, do we want them to be separate? We could store a separate AccessManualWebhook key and use that here, and on the UI it could be surfaced `connection_identifier`, like we use for saas configs. I didn't think we needed this extra identifier, but I can see an opposing argument.

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -60,6 +62,30 @@
...
 
+def get_access_manual_webhook_inputs(
+    db: Session, privacy_request: PrivacyRequest
+) -> Tuple[Dict[str, List[Dict[str, Optional[Any]]]], bool]:
+    """Retrieves manually uploaded data for all AccessManualWebhooks and formats in a way
+    to match automatically retrieved data. Also returns if execution should proceed.
+
+    This data will be uploaded to the user as-is, without filtering.
+    """
+    manual_inputs: Dict[str, List[Dict[str, Optional[Any]]]] = {}
```
pattisdr, line 73: Manual webhooks aren't currently connected to policies, but I am wondering if I should only run this if there are "access" rules defined on the policy?

src/fidesops/ops/service/privacy_request/request_runner_service.py
pattisdr: Discussed with Sean, seems okay for now to have this 1:1 mapping and use the connection_config key as the identifier for the entire webhook.

src/fidesops/ops/service/privacy_request/request_runner_service.py
pattisdr: I think this makes sense to add, will be a one-line addition.  This keeps the privacy request from being placed in a `requires_input` stage if the policy has strictly erasure rules only.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1152,3 +1166,146 @@ def _deny_request(
...
+        )
+
+    access_manual_webhooks: List[AccessManualWebhook] = AccessManualWebhook.get_enabled(
+        db
+    )
+    try:
+        for manual_webhook in access_manual_webhooks:
+            privacy_request.get_manual_webhook_input(manual_webhook)
+    except (MissingManualWebhookData, PydanticValidationError) as exc:
+        raise HTTPException(
```
eastandwestwind, line 1295: does this mean we don't allow for partial uploads? For example- one upload request for email, and another for address?

src/fidesops/ops/api/v1/scope_registry.py
```diff
@@ -20,9 +20,11 @@
 PRIVACY_REQUEST_READ = "privacy-request:read"
 PRIVACY_REQUEST_DELETE = "privacy-request:delete"
 PRIVACY_REQUEST_CALLBACK_RESUME = (
-    "privacy-request:resume"  # User has permission to restart a paused privacy request
+    "privacy-request:resume"  # User has permission to resume a privacy request
 )
 PRIVACY_REQUEST_REVIEW = "privacy-request:review"
+PRIVACY_REQUEST_UPLOAD_DATA = "privacy-request:upload_data"
+PRIVACY_REQUEST_VIEW_DATA = "privacy-request:view_data"
```
eastandwestwind, line 27: do we also need to add new scopes in fideslib at this point?

src/fidesops/ops/models/manual_webhook.py
```diff
@@ -23,3 +27,45 @@ class AccessManualWebhook(Base):
...
+
+        Returned as a default if no data has been uploaded for a privacy request.
+        """
+        return {
+            key: None
+            for key in (self.fields_schema.schema().get("properties") or {}).keys()
+        }
+
+    @classmethod
+    def get_enabled(cls, db: Session) -> List["AccessManualWebhook"]:
```
eastandwestwind, line 62: I haven't used this type notation before- what's the difference between wrapping the type in quotes vs just using `List[AccessManualWebhook]` ?

src/fidesops/ops/models/manual_webhook.py
```diff
@@ -23,3 +27,45 @@ class AccessManualWebhook(Base):
     )
 
     fields = Column(MutableList.as_mutable(JSONB), nullable=False)
+
+    @property
+    def fields_schema(self) -> BaseSchema:
+        """Build a dynamic Pydantic schema from fields defined on this webhook"""
```
eastandwestwind, line 33: this is really cool to see how you've build this dynamic schema 🙌

src/fidesops/ops/schemas/manual_webhook_schemas.py
```diff
@@ -26,10 +26,16 @@ class Config:
         orm_mode = True
 
 
+if TYPE_CHECKING:
```
eastandwestwind, line 29: didn't know about this built-in. Any particular reason to use this vs type ignore?

src/fidesops/ops/schemas/manual_webhook_schemas.py
```diff
@@ -41,6 +47,33 @@ def check_for_duplicates(
         """
         Verify that pii_fields and dsr_package_labels are unique.
 
+        Set the dsr_package_label to a snake_cased lower case version of pii field if it doesn't exist.
+        """
+        unique_pii_fields: Set[str] = {field.pii_field for field in value}
+        if len(value) != len(unique_pii_fields):
+            raise ValueError("pii_fields must be unique")
```
eastandwestwind, line 54: Are we translating these `ValueError`s into http exceptions in `patch_access_manual_webhook` and `create_access_manual_webhook` endpoints? I think `HTTP_400_BAD_REQUEST` makes sense for these

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -60,6 +62,34 @@
 logger = get_task_logger(__name__)
 
 
+def get_access_manual_webhook_inputs(
+    db: Session, privacy_request: PrivacyRequest, policy: Policy
+) -> Tuple[Dict[str, List[Dict[str, Optional[Any]]]], bool]:
```
eastandwestwind, line 67: Can we get this added as a pydantic type? Specifically, I'm unsure what the `bool` refers to.

tests/ops/service/privacy_request/request_runner_service_test.py
```diff
@@ -1773,3 +1773,91 @@ def test_email_connector_no_updates_needed(
...
+        )
+        db.refresh(pr)
+        assert (
+            pr.status == PrivacyRequestStatus.complete
+        )  # Privacy request not put in "requires_input" state
+        assert not mock_upload.called  # erasure only request, no data uploaded
+        assert mock_erasure.called
+
+    @mock.patch("fidesops.ops.service.privacy_request.request_runner_service.upload")
+    def test_pass_on_manually_added_input(
```
eastandwestwind, line 1842: can we additionally test the case where only some of the data needed is input?

eastandwestwind:
Nice work on this @pattisdr ! Learning some new things around pydantic models and types. Just a couple Qs to address.

Tagging @conceptualshark also to review docs.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
pattisdr: Discussed in slack, but we just need confirmation that data was queried for all manual webhooks.  It's okay if nothing was found; all fields can be None, but there needs to be an entry for each manual webhook in the cache.

src/fidesops/ops/api/v1/scope_registry.py
pattisdr: @sanders41 are we keeping fideslib scopes synced with fidesops scopes with the fides unification work, or is it okay that this is the most current list?

src/fidesops/ops/models/manual_webhook.py
pattisdr: This is only because this class method is within the AccessManualWebhook class itself. AccessManualWebhook isn't bound yet when we're evaluating the function return .  Wrapping it in quotes lets us delay when the type gets evaluated.  I think in later versions of Python this delayed evaluation will become the default.

src/fidesops/ops/models/manual_webhook.py
pattisdr: ah thanks!

src/fidesops/ops/schemas/manual_webhook_schemas.py
pattisdr: good question, I was just running into mypy issues with Pydantic's `conlist`, and this still gives mypy something to work with without ignoring it altogether. 

https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking
the `if TYPE_CHECKING` branch is only seen by mypy and the `else` branch is not seen by mypy and processed at runtime.

src/fidesops/ops/schemas/manual_webhook_schemas.py
pattisdr: Yes, http exception conversion is already automatically handled by FastAPI/Pydantic - any ValueErrors we raise in these validators end up getting converted to 422's https://fastapi.tiangolo.com/tutorial/handling-errors/#override-request-validation-exceptions.  

While I could add a custom exception handler, I think this is more consistent with our other endpoints, in that most of the time if something fails at the pydantic schema level, we're raising this default 422.  

There's also a unit test included with this work that confirms an http exception is thrown!

src/fidesops/ops/service/privacy_request/request_runner_service.py
pattisdr: sure, it's in the docstring, but I see that it's less clear.  the bool is whether or not request execution should proceed, but I can convert this into a pydantic schema.

src/fidesops/ops/models/manual_webhook.py
eastandwestwind: Oh wow, didn't know about this, thanks for the explanation!

src/fidesops/ops/schemas/manual_webhook_schemas.py
eastandwestwind: oh cool, looks like it can be useful for those circular import issues we sometimes run into- https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles . Although ideally we'd structure the code such that this isn't an issue...

src/fidesops/ops/schemas/manual_webhook_schemas.py
eastandwestwind: I'd forgotten about that automatic conversion, this makes sense!

src/fidesops/ops/service/privacy_request/request_runner_service.py
eastandwestwind: If it's not too much work, that'd be great to have this as a pydantic schema

tests/ops/service/privacy_request/request_runner_service_test.py
pattisdr: yes I can better verify the behavior in these end-to-end tests, good call

pattisdr:
Thanks for your comments @eastandwestwind, back to you!

src/fidesops/ops/service/privacy_request/request_runner_service.py
```diff
@@ -69,32 +70,40 @@
 logger = get_task_logger(__name__)
 
 
+class ManualWebhookResults(BaseSchema):
+    """Represents manual webhook data retrieved from the cache and whether privacy request execution should continue"""
+
+    manual_data: Dict[str, List[Dict[str, Optional[Any]]]]
```
eastandwestwind, line 76: thanks!

eastandwestwind (approved):
Appreciate all the detail on these changes, @pattisdr !

src/fidesops/ops/api/v1/scope_registry.py
sanders41: I have been trying to keep them in sync, but it won't break if they get out of sync for now.

pattisdr:
Thanks for taking a look @conceptualshark!

PR #1289, opened 2022-09-09
1183 saas connectors shopify erasure

### LINKED ISSUE #1183: Shopify
[Spec issue](https://github.com/ethyca/fidesops/issues/1030)
[Spec Doc](https://ethyca.atlassian.net/wiki/spaces/EN/pages/2432630785/Shopify+Connector+Specification)

[Access issue](https://github.com/ethyca/fidesops/issues/1106)

Only needs erasure

HamzaWaseemOnBench:
# Purpose

- Purpose of this PR is to add Shopify connector erasure endpoints

# Changes

- Added config for Shopify Erasure Endpoints
- Updated related dataset and added configuration
- Added associated tests and fixtures to validate above functionality

# Checklist

# Ticket

Fixes #1183


tests/ops/fixtures/saas/shopify_fixtures.py
```diff
@@ -100,3 +104,153 @@ def shopify_dataset_config(
...
     dataset.delete(db=db)
+
+
+@pytest.fixture(scope="function")
+def shopify_erasure_data(
+    shopify_connection_config, shopify_erasure_identity_email, shopify_secrets
+) -> Generator:
+    """
+    Creates a dynamic test data record for erasure tests.
+    Yields customer ID as this may be useful to have in test scenarios
```
adamsachs, line 115: nit: i'd probably update this comment since it's more than just customer ID that's yielded

tests/ops/fixtures/saas/shopify_fixtures.py
```diff
@@ -100,3 +104,153 @@ def shopify_dataset_config(
...
+    assert orders_response.ok
+    order = orders_response.json()
+    order_id = order["order"]["id"]
+
+    # Get Blog
+    blogs_response = requests.get(
+        url=f"{base_url}/admin/api/2022-07/blogs.json", headers=headers
+    )
+    assert blogs_response.ok
+    blog = blogs_response.json()["blogs"][1]
```
adamsachs, line 188: any particular reason we get a blog rather than creating one for the test fixture? i haven't really looked at the API and i don't have much context, so it may be a dumb suggestion/infeasible. but ideally we'd not have to rely on any data being present for the tests to pass, so if we could create the blog in the fixture (and then clean it up afterward), i think it'd be preferred. let me know what you think.

tests/ops/integration_tests/saas/test_shopify_task.py
```diff
@@ -260,3 +262,247 @@ async def test_shopify_access_request_task(
...
+
+    assert x == {
+        f"{dataset_name}:customers": 1,
+        f"{dataset_name}:blogs": 0,
+        f"{dataset_name}:customer_orders": 1,
+        f"{dataset_name}:customer_addresses": 1,
+        f"{dataset_name}:blog_articles": 0,
+        f"{dataset_name}:blog_article_comments": 1,
+        f"{dataset_name}:customer_order_transactions": 0,
+    }
```
adamsachs, line 506: i assume that the updates take too long on the source system (i.e. on the shopify platform) to actually query to make sure our data gets updated - but just want to check.

tests/ops/fixtures/saas/shopify_fixtures.py
```diff
@@ -100,3 +104,153 @@ def shopify_dataset_config(
...
+    order_delete_response = requests.delete(
+        url=f"{base_url}/admin/api/2022-07/orders/{order_id}.json",
+        headers=headers,
+    )
+    assert order_delete_response.ok
+    article_delete_response = requests.delete(
+        url=f"{base_url}/admin/api/2022-07/articles/{article_id}.json",
+        headers=headers,
+    )
+    assert article_delete_response.ok
```
adamsachs, line 236: maybe i'm mising something, but is there a reason we don't delete/clean up the `customer` that we've created for the test here?

adamsachs (commented):
generally looks good @HamzaWaseemOnBench , just a few questions that i wanted to check on before merging. happy to chat about it offline if that's easier!

also - please update the `CHANGELOG.md` (after merging or rebasing off of `main`), as we'll need that before we can merge the PR

tests/ops/fixtures/saas/shopify_fixtures.py
HamzaWaseemOnBench: Basically in Shopify we can't filter specific blogs by our customer email so creating one blog in fixture and using it to create article was actually resulting in creating articles in all existing blogs. I had a session with Adrian to verify this when I identified this issue so we agreed to create one blog for testing there already and then we should just get that blog to create new article for erasure endpoint. I can explain this offline in case this isn't clear.

tests/ops/fixtures/saas/shopify_fixtures.py
HamzaWaseemOnBench: Yes basically we don't have a delete customer endpoint in Shopify API, that is why we can't delete newly created customer.

tests/ops/integration_tests/saas/test_shopify_task.py
HamzaWaseemOnBench: So here we were first going for only getting existing data in erasure fixture instead of creating new(like we are doing for blogs endpoint). Since after first test data was masked already, quering it again to check if it is masked didn't make sense at that time that is why we didn't add actual query check. Since now as we have changed our approach after discussion and we are actually creating some new data(and deleting it after fixture) it'll be good to add actual query checks, I'll add it now.

tests/ops/fixtures/saas/shopify_fixtures.py
HamzaWaseemOnBench: Nice catch, I'll update it too in new commit.

tests/ops/fixtures/saas/shopify_fixtures.py
adamsachs: makes sense - thanks for explaining!

tests/ops/fixtures/saas/shopify_fixtures.py
adamsachs: hmm, OK - so do we end up just creating a new customer every time we run a test? could that get unwieldy after a while? just wondering if shopify provides any insight into whether having too many customers could become a problem. i guess it's something they should be able to handle if they don't offer any API to delete a customer.

anyway, not a blocker, especially since i don't think there's a whole lot we can do about this. but i am a _little_ bit concerned we may start running into problems after months of running tests and creating customers each time.

HamzaWaseemOnBench: I had discussion with Adrian related to it too and we were rooting for approach of not creating customer in every call first too and get already created customer only but we had a problem that data was already masked just after our first call and there was practically no use of our assertions for it so we decided to go for creating customer instead. Since we created a new Shopify store with no data and there is no limit of creating customers as far as we researched so we opted for this solution.

adamsachs: got it - thanks for walking through that @HamzaWaseemOnBench !

adamsachs (approved):
this all looks good - thanks for the iterations @HamzaWaseemOnBench!

Issue #1161, opened 2022-08-25
API to record consent preferences


seanpreston:
Add an API endpoint to:

<br/>

<br/>

```<br/>POST /api/v1/consent

<br/>

<br/>

{<br/>"identity": " [test@example.com](mailto:test@example.com)", # should this be hashed?<br/>"identity_type": "email",<br/> "regulation": "gdpr",<br/> "data_use": "this can validated against Fideslang until we unify Fides", # we won't support custom data uses<br/>"opt_in": "true"<br/>}

<br/>

<br/>

```

<br/>

<br/>

1. Create a `ProvidedIdentity` record for the identity value passed.
2. Create a corresponding `Consent` record.

<br/>

seanpreston:
Good spot — while it's impossible to verify phone numbers with Fidesops currently that doesn't mean people can't verify the numbers themselves and add them. I believe we've not hit this with Privacy Requests yet as the only folks using them are using email only (which is configurable in the privacy centre itself). In this scenario it's reasonable to expect an auth token in order to perform the action.

My main concern with that conclusion is how big a lift it is to build in another auth method depending on payload at the same URL, I don't think we have a convention for this yet.

PR #1301, opened 2022-09-12
Add consent table

### LINKED ISSUE #1160: Consent database updates
### Description

This PR describes the changes required to the Fidesops application DB in order to store consent preferences for data subjects.

### Acceptance Criteria

1. Add the following table to the DB:

```
Consent
- identity: ProvidedIdentity
- regulation: Enum(gdpr, ccpa, ...)  # This should be optional
- data_use: string  # once we unify Fides this will map to custom data uses created in Fidesctl
- data_use_description: string  # something we can use to describe a human readable version of what this use case _is_
- opt_in: bool
```

**Please note that all models should inherit from the FidesBase, which will also provide utility fields such as `uuid`, `created_at`, etc**

2. Make `ProvidedIdentity.privacy_request_id` a nullable field.
3. Update `ProvidedIdentity.hash_value()` to use no salt when hashing identities for consent purposes.
4. Update the `get_request_status` API view to exclude any `ProvidedIdentity`s where `privacy_request_id` is `null`.

sanders41:
# Purpose

❗ Contains migrations.

Adds a  table to store consent preferences for data subjects.

# Changes
- Added a consent table.

# Checklist

# Ticket

Fixes #1160


src/fidesops/ops/models/privacy_request.py
```diff
@@ -730,6 +732,18 @@ def hash_value(
         return hashed_value
 
 
+class Consent(Base):
+    """The DB ORM model for Consent."""
+
+    provided_identity_id = Column(String, ForeignKey(ProvidedIdentity.id), unique=True)
+    regulation = Column(EnumColumn(Regulations))
```
NevilleS, line 739: This doesn't make too much sense to me for a couple reasons- one is that consent is a distinct consult from regulation (you don't collect consent for GDPR, you collect for a specific use of a customer's data), and second is that we can't realistically maintain an enum of regulations (this regulations enum is already way too limited)

We should remove the regulation field entirely from this so it didn't confuse 👍

src/fidesops/ops/models/privacy_request.py
```diff
@@ -714,6 +714,7 @@ class ProvidedIdentity(Base):  # pylint: disable=R0904
         ),
         nullable=True,
     )  # Type bytea in the db
+    consent = relationship("Consent", back_populates="provided_identity", uselist=False)
```
sanders41, line 717: I have `uselist` as `False` because I am assuming consent would only need to be given one time. Is that a correct assumption?

pattisdr:
Starting review..

src/fidesops/ops/models/privacy_request.py
```diff
@@ -730,6 +731,17 @@ def hash_value(
         return hashed_value
 
 
+class Consent(Base):
```
pattisdr, line 734: ~Do we need the regulation field that's listed in https://github.com/ethyca/fidesops/issues/1160?~ Nope, Neville requested removal above

src/fidesops/ops/models/privacy_request.py
```diff
@@ -730,6 +731,17 @@ def hash_value(
...
+    """The DB ORM model for Consent."""
+
+    provided_identity_id = Column(String, ForeignKey(ProvidedIdentity.id), unique=True)
+    data_use = Column(String)
+    data_use_description = Column(String)
+    opt_in = Column(Boolean, nullable=False)
+
+    provided_identity = relationship(ProvidedIdentity, back_populates="consent")
```
pattisdr, line 742: Since the unique constraint is defined on this table, I might define both sides of the relationship here with a backref, instead of using `back_populates` on both models, something like: 
```python3
provided_identity = relationship("ProvidedIdentity", backref=backref("consent", uselist=False))
```

and then the `consent` definition could be removed from ProvidedIdentity.   This is just a matter of preference though, in tending to define relationships on one side.

pattisdr: Good testing additions might be model-level tests that assert that both sides of the relationship are defined properly, asserting both `provided_identity.consent` and `consent.provided_identity`. We have a tests > models folder.

src/fidesops/ops/migrations/versions/a0e6feb5bdc8_add_consent.py
```diff
@@ -0,0 +1,66 @@
...
+        sa.Column("provided_identity_id", sa.String()),
+        sa.Column("data_use", sa.String(), nullable=True),
+        sa.Column("data_use_description", sa.String(), nullable=True),
+        sa.Column("opt_in", sa.Boolean(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["provided_identity_id"],
+            ["providedidentity.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("provided_identity_id"),
```
pattisdr, line 44: New tables are a good opportunity to update the database diagram https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md

src/fidesops/ops/models/privacy_request.py
```diff
@@ -685,7 +686,6 @@ class ProvidedIdentity(Base):  # pylint: disable=R0904
     privacy_request_id = Column(
         String,
         ForeignKey(PrivacyRequest.id_field_path),
-        nullable=False,
```
pattisdr, line 688: I see this is specified in the ticket, but where are privacy requests becoming non-nullable?

src/fidesops/ops/models/privacy_request.py
sanders41: I originally had `backref` and changed it because I saw [this](https://docs.sqlalchemy.org/en/14/orm/backref.html) in the sqlalchemy docs where is say `backref` is legacy now and `back_populates` should be preferred. Thoughts?

src/fidesops/ops/models/privacy_request.py
sanders41: I was wondering this also. The [other part](https://github.com/ethyca/fidesops/issues/1161) of the ticket shows receiving the data as the following so I think we are going to create an identity and consent with no privacy request here.

```json
{
  "identity": "test@example.com",  # should this be hashed?
  "identity_type": "email",
  "regulation": "gdpr",
  "data_use": "this can validated against Fideslang until we unify Fides",  # we won't support custom data uses
  "opt_in": "true"
}
```

The other thing I was questioning here is if we really need a new table for this or if we should just add the `opt_in` to the `ProvidedIdentity`. Since it's a 1 to 1 relationship it seems like they could just go together.

src/fidesops/ops/models/privacy_request.py
pattisdr: ah thanks I wasn't current with this latest recommendation.

src/fidesops/ops/models/privacy_request.py
pattisdr: ah I see. Maybe this could go on the ProvidedIdentity if we're sure it's going to stay a 1:1, I'd check with someone else who's more informed about this feature-

sanders41:
@pattisdr I updated the models to allow multiple consents and added a test, but still need to update the database diagram.

src/fidesops/ops/models/privacy_request.py
pattisdr: discussed in slack, we will be creating ProvidedIdentities that are not related to PrivacyRequests.  and there can be multiple data uses for a single identity.

sanders41:
Database diagram is updated now also so I think it is ready for another review.

tests/ops/models/test_privacy_request.py
```diff
@@ -669,3 +671,31 @@ def test_can_run_if_no_saved_checkpoint(self):
...
+    }
+    Consent.create(db, data=consent_data_1)
+
+    consent_data_2 = {
+        "provided_identity_id": provided_identity.id,
+        "data_use": "user.browsing_history",
+        "opt_in": False,
+    }
+    Consent.create(db, data=consent_data_2)
+    db.commit()
```
pattisdr, line 697: Not needed here specifically because our custom create does a commit by default

tests/ops/models/test_privacy_request.py
```diff
@@ -669,3 +671,31 @@ def test_can_run_if_no_saved_checkpoint(self):
...
         )
+
+
+def test_consent(db):
+    privided_identity_data = {
+        "privacy_request_id": None,
+        "field_name": "email",
+        "encrypted_value": {"value": "test@email.com"},
+    }
+    provided_identity = ProvidedIdentity.create(db, data=privided_identity_data)
```
pattisdr, line 682: You could clean up these resources at the end of your test by deleting the `provided_identity` and also verify then that both consent resources are created - a good opportunity to protect your cascade delete functionality with a unit test!

src/fidesops/ops/migrations/versions/a0e6feb5bdc8_add_consent.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
+        "providedidentity",
+        "privacy_request_id",
+        existing_type=sa.VARCHAR(),
+        nullable=False,
+    )
```
pattisdr, line 63: The downgrade might be problematic if there were provided identities created in the interim without privacy_request_ids.  Although we generally haven't paid much attention to the consequences of downgrades.

pattisdr (commented):
good work I just have a couple comments about the unit test and the downgrade but feel free to push back

tests/ops/models/test_privacy_request.py
```diff
@@ -669,3 +671,31 @@ def test_can_run_if_no_saved_checkpoint(self):
             )
             is True
         )
+
+
+def test_consent(db):
+    privided_identity_data = {
```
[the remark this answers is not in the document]
sanders41: Is it a typo or just a fancy variable name 🤔

tests/ops/models/test_privacy_request.py
sanders41: In general [this fixture](https://github.com/ethyca/fidesops/blob/5da10cc9f650c6890027be3c6f5c6c2ba6189c26/tests/ops/conftest.py#L107) will do the clean up, but I didn't consider the cascade delete test angle. I'll add it.

tests/ops/models/test_privacy_request.py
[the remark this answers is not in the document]
pattisdr: you can leave it if you wish :)

src/fidesops/ops/migrations/versions/a0e6feb5bdc8_add_consent.py
sanders41: I see your point here. I'm not sure there is anything we can do about it though? I guess we could leave it optional on a downgrade.

tests/ops/models/test_privacy_request.py
pattisdr: that's right, I forgot you added this cleanup fixture.  I was thinking we still had issues somewhere with collision issues but maybe those are more related to the cache and not the application db.

src/fidesops/ops/migrations/versions/a0e6feb5bdc8_add_consent.py
pattisdr: maybe it's fine that it fails. they won't be able to downgrade in this case and they'll have to adjust the existing data on their own instead of us guessing for them how to handle.

pattisdr (approved):
nice work here

sanders41:
Thanks @pattisdr!

PR #1307, opened 2022-09-13
Utility to update saas configs and datasets on system startup

### LINKED ISSUE #1098: Update Saas Configs and Datasets on System Startup

We update the saas config and saas dataset files over time, which can cause them to diverge with saas configs and saas datasets that have been saved in a user's database.

On system startup, if the version in the registry is newer than the SaaS config version in the connection config table, update the saas config and the dataset.  We also need to standardize the version number that already exists in the SaaS configs, they are all currently set to version 0.0.1

Instead of creating data migrations every time we update a saas config or dataset config template file, add logic to just make the changes automatically on startup.  

Steps:
-  get all saas connection configs
- get version number from SaaS config (figure out versioning)
- if registry version is newer
  - update saas config
  - update dataset

Add ability to turn off auto update 

### Additional context

Separated from #814 to release that functionality more quickly

adamsachs:
# Purpose
Utility run on server bootstrap to update SaaS config instances in the DB based on updates to SaaS config or dataset template updates in the template registry

# Changes
- utility to perform update "script" on server bootstrap
- invoke utility on server bootstrap
- refactor instantiation endpoint to use some shared code
- tests for above functionality

# Checklist

# Ticket

Fixes #1098


adamsachs:
a large amount of the added code here is from the test cases, apologies for that. i couldn't think of a way to make the test cases more concise while keeping them robust - they required a fair amount of rather cumbersome mocking.

i'm open to any suggestions on how to make those tests more concise and/or readable!

src/fidesops/main.py
```diff
@@ -237,12 +239,15 @@ def start_webserver() -> None:
...
 
     logger.info("Validating SaaS connector templates...")
-    load_registry(registry_file)
+    registry = load_registry(registry_file)
 
     if config.database.enabled:
         logger.info("Running any pending DB migrations...")
         try:
             init_db(config.database.sqlalchemy_database_uri)
+            SessionLocal = get_db_session(config)
```
adamsachs, line 248: wasn't totally sure this was the best way to go about things

src/fidesops/ops/api/v1/endpoints/saas_config_endpoints.py
```diff
@@ -294,47 +292,28 @@ def instantiate_connection_from_template(
...
-    connection_config.secrets = validate_secrets(
-        template_values.secrets, connection_config
-    ).dict()
-    connection_config.save(db=db)  # Not persisted to db until secrets are validated
-
-    try:
-        dataset_config: DatasetConfig = create_dataset_config_from_template(
-            db, connection_config, connector_template, template_values
-        )
+    except HTTPException as exc:
```
adamsachs, line 304: propagate up an HTTP exception that's thrown from the underlying function call, since that's ready to go back to the client, no need to wrap it further

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -133,3 +140,129 @@ def load_registry(config_file: str) -> ConnectorRegistry:
...
+    for connector_type in registry.connector_types():
+        logger.debug(
+            "Determining if any updates are needed for connectors of type %s based on templates...",
+            connector_type,
+        )
+        template: ConnectorTemplate = registry.get_connector_template(  # type: ignore
+            connector_type
+        )
+        saas_config_template = SaaSConfig.parse_obj(load_config(template.config))
+        template_version: Union[LegacyVersion, Version] = parse_version(
```
pattisdr, line 162: Unfamiliar with these LegacyVersion versus Versions here, what is this?

tests/ops/service/connectors/test_connector_registry_service.py
```diff
@@ -20,3 +53,397 @@ def test_get_connector_template(self):
...
+        db=db,
+        conditions=DatasetConfig.fides_key == tertiary_mailchimp_dataset.fides_key,
+    ).first()
+    validation_function(
+        tertiary_mailchimp_dataset,
+        mailchimp_template_config,
+        mailchimp_template_dataset,
+        tertiary_mailchimp_config.key,
+        tertiary_mailchimp_dataset.fides_key,
+    )
```
pattisdr, line 255: Your tests are clever but it takes awhile to trace through and figure out the comparisons you're making and believe that you're testing all the important things and not just mocking too much functionality away.  

This is just my opinion, I know others disagree here, but I prefer tests to be extremely clear about what they are testing. My tests tend to be less DRY than my application code but I view them as a secondary way to document the code: new users to the codebase can look at a test and quickly see what a function or API endpoint requires and what the expected results are without having to additionally understand unique quirks about the test itself.  I also view tests as the most important way to protect the critical features of your own code, and when unit tests are too clever, I think it's easy for a dev to just delete a unit test rather than take the time to understand it.  I also want to make sure we are testing as much end-to-end as possible too.

Not requesting a rewrite here, you've put a lot of work into this, just providing my viewpoint on tests really

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -133,3 +140,129 @@ def load_registry(config_file: str) -> ConnectorRegistry:
...
+                    )
+                except Exception:
+                    logger.error(
+                        "Encountered error attempting to update SaaS config instance %s",
+                        saas_config_instance.fides_key,
+                        exc_info=True,
+                    )
+
+
+def update_saas_instance(
```
pattisdr, line 195: I _would_ like to see an end-to-end test of this method.

pattisdr (changes requested):
nice refactoring work here @adamsachs to re-use existing code.

my primary concern here is the (likely, in my opinion) scenario when a version upgrade fails.

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -133,3 +140,129 @@ def load_registry(config_file: str) -> ConnectorRegistry:
...
+    with new instance configuration data based on the given ConnectorTemplate
+    """
+    template_vals = SaasConnectionTemplateValues(
+        name=connection_config.name,
+        key=connection_config.key,
+        description=connection_config.description,
+        secrets=connection_config.secrets,
+        instance_key=saas_config_instance.fides_key,
+    )
+    connection_config.delete(db)
```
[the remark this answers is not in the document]
adamsachs: thanks for calling this out @pattisdr, i hadn't really grasped how common an update failure may be - let me see how we can do the update in a less risky way.

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -1,12 +1,17 @@
...
+from typing import Dict, Iterable, List, Optional, Union
 
 from fideslib.core.config import load_toml
+from fideslib.exceptions import KeyOrNameAlreadyExists
+from packaging.version import LegacyVersion, Version
+from packaging.version import parse as parse_version
 from pydantic import BaseModel, validator
 from sqlalchemy.orm import Session
 
+from fidesops.ops.api.v1.endpoints.connection_endpoints import validate_secrets
```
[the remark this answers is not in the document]
adamsachs: agreed, i was feeling a bit uneasy about the HTTP exceptions coming out of here, and i like your suggestion - similar to the approach i'd taken in other parts of this code

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
adamsachs: this is just the return type coming from `packaging.version.parse` - i needed to include the union type here to satisfy mypy (or at least that was what i thought was the simplest way to satisfy)

adamsachs:
@pattisdr would you be able to take a look at my latest changes (https://github.com/ethyca/fidesops/pull/1307/files/eabe839e22b14378a46357b2eca5ac55e27ceb62..6ec9f492f6ed66b307e2804181bf49d5f66b8f3e) to see if they address your core concerns?

specifically, i refactored the update logic to no longer delete configs from the db, but instead update in place. while making these changes, i realized we should be able to do a much more "targeted" replace -- i.e., just replace the `SaaSConfig` attached to the `ConnectionConfig`, since the `ConnectionConfig` itself is independent of the template. 

similarly, i don't _think_ we need to actually validate secrets on this update path - in fact, i don't think we _want_ to, per your concern about the common error case of adding a secret to the template. we should aim to keep any existing configured secrets in tact, and we should add any default values provided by the new template if applicable (which this PR does). validating the secrets in this use case is nonsensical IMO, since the user isn't actually inputting any new secrets. of course, [this comment](https://github.com/ethyca/fidesops/issues/1098#issuecomment-1239763318) still applies, and we should make a new ticket to alert/warn the user about new secrets added to a template.

effectively, i think the new changes here both make it less likely for us to error on template updates, and also make things safer _if_ we happen to error. i think the code also ends up a bit cleaner.

in the meantime, i'll look into cleaning up some of the core tests, which i still haven't addressed. i just wanted to get this new approach up for your review sooner rather than later.

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -133,3 +138,83 @@ def load_registry(config_file: str) -> ConnectorRegistry:
...
+        secrets=connection_config.secrets,
+        instance_key=saas_config_instance.fides_key,
+    )
+
+    config_from_template: Dict = load_config_with_replacement(
+        template.config, "<instance_fides_key>", template_vals.instance_key
+    )
+    connection_config = ConnectionConfig.get_connection_config_or_error(
+        db, template_vals.key  # type: ignore
+    )
```
pattisdr, line 217: Why do you need to do an additional lookup here, when the resource is passed in above?

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -123,7 +128,7 @@ def create_dataset_config_from_template(
         "fides_key": template_values.instance_key,
         "dataset": dataset_from_template,
     }
-    dataset_config = DatasetConfig.create(db, data=data)
+    dataset_config = DatasetConfig.create_or_update(db, data=data)
```
pattisdr, line 131: This seems fine given that the endpoint that creates the resources from a template initially has an early check to make sure that the DatasetConfig fides_key doesn't already exist.

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
```diff
@@ -133,3 +138,83 @@ def load_registry(config_file: str) -> ConnectorRegistry:
...
+        instance_key=saas_config_instance.fides_key,
+    )
+
+    config_from_template: Dict = load_config_with_replacement(
+        template.config, "<instance_fides_key>", template_vals.instance_key
+    )
+    connection_config = ConnectionConfig.get_connection_config_or_error(
+        db, template_vals.key  # type: ignore
+    )
+    connection_config.update_saas_config(db, SaaSConfig(**config_from_template))
```
pattisdr, line 218: As you mentioned, if our secrets are missing or invalid now we need to make it very clear somewhere that they need to update this.

pattisdr (commented):
nice @adamsachs I like this approach much better, doing more targeted updates to the connectionconfig.saas_config and updating the existing datasetconfig instead of a delete/create.

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
adamsachs: right, and i wavered between just subbing this in or having the function take another param to indicate whether a create or an update was desired.

i can see how there's a chance this functionality could be missed by someone else calling this function. i'll update the docstring of the function to make it clear that it creates or updates.

let me know if you think we should approach it differently

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
adamsachs: whoops, i think i just missed this in refactoring. i'll clean it up

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
adamsachs: yup - but in discussions with @galvana, we'd agreed to tackle this in a separate issue/PR (per https://github.com/ethyca/fidesops/issues/1098#issuecomment-1239763318). it does make me a bit uneasy to include this as is, but in the interest of iterating, are you OK with splitting that piece out?

pattisdr: makes sense to me!

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
pattisdr: The function name could be changed too, I think that would also help!

src/fidesops/ops/service/connectors/saas/connector_registry_service.py
pattisdr: Clarified in person, Adam's existing test harness is bigger picture than this and it is being covered in a more end-to-end way than i realized.

pattisdr (approved):
Nice work here @adamsachs.

Discussed separately, there's some followups needed to surface to the user if they need to update their secrets, or some portion of the automatic update failed so the user knows they need to adjust on their end, without it just being buried in the logs. Adam's creating a followup ticket for this.

I'll leave it to your team to merge.

PR #748, opened 2022-06-28
706 Adds SaaS connection type to SaaS yaml config

eastandwestwind:
# Purpose
 Adds SaaS connection type to SaaS yaml config

# Changes
- Adds SaaS connection type to SaaS yaml config
- Updates test for invalid saas type
- Updates docs and postman collection

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [x] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [x] Add description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [x] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [x] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [x] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [ ] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #706

PR #1327, opened 2022-09-15
Timescale POC [#1311]

### LINKED ISSUE #1311: [Spike] TimescaleDB

pattisdr:
# Purpose

Demonstrate running an access request against a Timescale DB using the existing PostgreSQL connector. 

Unknowns:
- What additional authentication methods do we need to support?
- What additional data types do we need to support?

# Changes
-

# Checklist

# Ticket

Fixes #1311


docker/sample_data/timescale_example.sql
```diff
@@ -0,0 +1,191 @@
...
+('employee-1@example.com', '2022-01-01 09:00:00'),
+('employee-1@example.com', '2022-01-02 09:00:00'),
+('employee-1@example.com', '2022-01-03 09:00:00'),
+('employee-2@example.com', '2022-01-04 09:00:00'),
+('employee-1@example.com', '2022-01-05 09:00:00'),
+('employee-2@example.com', '2022-01-06 09:00:00'),
+('employee-2@example.com', '2022-01-07 09:00:00'),
+('employee-2@example.com', '2022-01-08 09:00:00');
+
+SELECT create_hypertable('onsite_personnel', 'time', migrate_data => true);
```
pattisdr, line 115: [Creates a hypertable here](https://docs.timescale.com/timescaledb/latest/how-to-guides/hypertables/create/#create-a-hypertable)  by first creating a postgres table and then converting it into a hypertable.  You have to specify the column that holds the "time" values. In this contrived example, I have unique times that should be the primary key with a "responsible" employee who was on call at that time.

I've added an integration test that demonstrates a basic access and erasure request on this hypertable.  Some comments are that we require primary keys to erase. Noting that timescale requires the unique/primary key dimension to contain the `time` dimension.  You can have composite primary keys including the time dimension in Timescale and fidesops has support for this.

Also worth noting that "datetimes" are not a fully supported date type in Fidesops yet.  A lot of our existing examples that deal with dates tend to treat dates as strings.  Masking date fields can be problematic outside the null masking strategy.

src/fidesops/ops/schemas/connection_configuration/connection_secrets_timescale.py
```diff
@@ -0,0 +1,12 @@
+from fidesops.ops.schemas.base_class import NoValidationSchema
+from fidesops.ops.schemas.connection_configuration.connection_secrets_postgres import (
+    PostgreSQLSchema,
+)
+
+
+class TimescaleSchema(PostgreSQLSchema):
```
sanders41, line 7: `PostgresSQLSchema` because timescale is built on that and you are just extending it?

src/fidesops/ops/service/connectors/timescale_connector.py
```diff
@@ -0,0 +1,5 @@
+from fidesops.ops.service.connectors.sql_connector import PostgreSQLConnector
+
+
+class TimescaleConnector(PostgreSQLConnector):
```
sanders41, line 4: Same question about `PostgreSQLConnector`

tests/ops/integration_tests/test_sql_task.py
```diff
@@ -1089,3 +1094,315 @@ async def test_retry_erasure(
...
+    timescale_integration_db,
+) -> None:
+    rule = erasure_policy.rules[0]
+    target = rule.targets[0]
+    target.data_category = "user"
+    target.save(db)
+
+    database_name = "my_timescale_db_1"
+    privacy_request = PrivacyRequest(
+        id=f"test_timescale_access_request_task_{random.randint(0, 1000000)}"
```
sanders41, line 1209: Is the random number important? It would be simpler and more reproducible without it, but if it is needed it's ok.

Is it to keep it from clashing with the other tests? If so there is the chance the numbers match causing random failures (small chance but still a chance).

tests/ops/integration_tests/test_sql_task.py
```diff
@@ -1089,3 +1094,315 @@ async def test_retry_erasure(
...
+
+@pytest.mark.integration_timescale
+@pytest.mark.integration
+@pytest.mark.asyncio
+async def test_timescale_query_and_mask_hypertable(
+    db, policy, erasure_policy, timescale_connection_config, timescale_integration_db
+) -> None:
+    database_name = "my_timescale_db_1"
+    privacy_request = PrivacyRequest(
+        id=f"test_timescale_access_request_task_{random.randint(0, 1000000)}"
```
sanders41, line 1336: Same random number question as previously.

tests/ops/integration_tests/test_sql_task.py
```diff
@@ -1089,3 +1094,315 @@ async def test_retry_erasure(
...
+@pytest.mark.asyncio
+async def test_timescale_access_request_task(
+    db,
+    policy,
+    timescale_connection_config,
+    timescale_integration_db,
+) -> None:
+    database_name = "my_timescale_db_1"
+    privacy_request = PrivacyRequest(
+        id=f"test_timescale_access_request_task_{random.randint(0, 1000000)}"
```
sanders41, line 1110: Same random number question as previously.

docker/docker-compose.integration-timescale.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  timescale_example:
+    image: timescale/timescaledb:latest-pg12
```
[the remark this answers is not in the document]
pattisdr: Good question, I personally viewed this as a starting point that we could continue to refine based on customer needs.  I used `12` because this is based on our Postgres connector which also uses `12`. There very well may be 14 functionality we need to add in the future.  As an aside, there's a lot of planned functionality that the postgres connector itself doesn't yet support. 

I did just attempt the 14 quickly and am running into `"SCRAM authentication requires libpq version 10 or above" ` issues. I haven't tried to troubleshoot this further.

src/fidesops/ops/migrations/versions/cf88efa1ad89_add_timescale_db.py
```diff
@@ -0,0 +1,45 @@
...
+# revision identifiers, used by Alembic.
+revision = "cf88efa1ad89"
+down_revision = "021d288d0ce3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add to ConnectionType enum
+    op.execute("alter type connectiontype rename to connectiontype_old")
```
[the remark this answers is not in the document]
pattisdr: There's a historical reason for this, let me track down the old comment.

OK here! https://github.com/ethyca/fidesops/pull/748#issuecomment-1176406578

First, I'm updating an enum here. We did used to update the enum directly as you suggested, with `alter type`, but Catherine added a migration awhile back that relied on the enum existing, but `alter type add value` doesn't commit this until the end of the transaction, which I think is the last migration.  This strategy let the enum be in the database at the point Catherine needed it.  I don't think we technically need to be doing this downstream of her migration, but I've just been keeping it consistent.

tests/ops/fixtures/timescale_fixtures.py
```diff
@@ -0,0 +1,76 @@
...
+def timescale_integration_session_cls(timescale_connection_config):
+    example_timescale_uri = TimescaleConnector(timescale_connection_config).build_uri()
+    engine = get_db_engine(database_uri=example_timescale_uri)
+    SessionLocal = get_db_session(
+        config=config,
+        engine=engine,
+        autocommit=True,
+        autoflush=True,
+    )
+    yield SessionLocal
```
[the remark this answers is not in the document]
pattisdr: good point, instead of closing it here, which is just the sessionmaker I could close it in the function below which actually creates the session.

tests/ops/integration_tests/test_integration_timescale_example.py
```diff
@@ -0,0 +1,32 @@
...
+import pytest
+from sqlalchemy import func, select, table
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.integration
+def test_timescale_example_data(timescale_integration_db):
+    """Confirm that the example database is populated with simulated data"""
+    expected_counts = {
```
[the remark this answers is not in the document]
pattisdr: yes that's tricky I'm not sure how to pull it out of the SQL file right off, but at the same time, if there are rows being added, in my opinion it's good that this test fails because that means potentially more rows might need to be returned in other tests that run access requests against this database.  I think it helps flag this.

src/fidesops/ops/schemas/connection_configuration/connection_secrets_timescale.py
pattisdr: exactly yes, I'll add a code comment to make this clearer. I could have used the PostgreSQLSchema and PostgreSQLConnector themselves below directly but I assumed at some point the two might diverge?

tests/ops/integration_tests/test_sql_task.py
pattisdr: We might not need the random number specifically, but the main thing we want is for this id not to clash with other privacy request ids in the redis cache because that will affect request execution which relies heavily on the cache.

The privacy request fixtures that we use elsewhere are recreated for each test so their id is different. Here, we're not using a fixture, but creating a resource that isn't persisted to the db (but its data is persisted to the cache).

This is a pattern that we follow elsewhere in the tests on this page - I increased from 1000 in other tests to 1000000. Open to more ideas around this because I do think occasional collisions are part of our testing issues.

sanders41: Is it required that it is a number? Would a uuid work?

pattisdr: Yes I could change it to a uuid thanks

sanders41 (approved):

pattisdr:
Thanks very much for your comments @sanders41!

PR #1227, opened 2022-08-31
Rollbar [access and erasure]

Zeeshan-Ethyca:
# Purpose

- This PR aims to add a Rollbar connector Projects, Occurrences,  People (_affected users_) access endpoints, and People erasure endpoint.

# Changes
- Added config for Rollbar Projects Access Endpoints
- Added config for Rollbar Project details Access Endpoints
- Added config for Rollbar Occurences Access Endpoints
- Added config for Rollbar People (affected users) erasure Endpoints
- Added related dataset and configuration
- Added associated tests and fixtures to validate the above functionality

# Checklist
- [x] Update [`CHANGELOG.md`](https://github.com/ethyca/fidesops/blob/main/CHANGELOG.md) file
  - [ ] Merge in main so the most recent `CHANGELOG.md` file is being appended to
  - [ ] Add a description within the `Unreleased` section in an appropriate category. Add a new category from the list at the top of the file if the needed one isn't already there.
  - [x] Add a link to this PR at the end of the description with the PR number as the text. example: [#1](https://github.com/ethyca/fidesops/pull/1)
- [x] Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).
- If docs updated (select one):
  - [ ] documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)
  - [ ] documentation issue created (tag docs-team to complete issue separately)
- [x] Good unit test/integration test coverage
- [ ] This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging
- [x] The `Run Unsafe PR Checks` label has been applied, and checks have passed, if this PR touches any external services

# Ticket

Fixes #798

PR #1331, opened 2022-09-16
334 api key authentication strategy

### LINKED ISSUE #334: API Key authentication strategy

Some API endpoints require an API key to be included in the HTTP request, either in the header or as a query param. A reusable API key auth strategy will allow us to include this api key in each request without having to specify it per endpoint.

A new authentication strategy which will allow the user to include an API key in the header or query params.

**Header**
```
authentication:
  strategy: api_key
  configuration:
    headers:
    - name: Token
      value: <api_token>
 ```

**Query Pararm**
```
authentication:
  strategy: api_key
  configuration:
    query_params:
    - name: token
      value: <api_token>
```

The value in the placeholder `<api_token>` should reference the name of a param in `connector_params` (implied, it won't need to be added as a param value).

### Additional context

This strategy uses a different format to specify required values compared to `basic_authentication` and `bearer_authentication` but this is done to be more inline with the established pattern for params and param values in the `request` field within the SaaS config. The goal will be to use this new format for the other two auth strategies and across the rest of the SaaS config.

Bearer authentication can also be implemented using this new API key strategy since we could just do something like this
```
authentication:
  strategy: api_key
  configuration:
    header:
    - name: Bearer
      value: <api_token>
```
But I believe it is important to have basic and bearer auth as separately named strategies since those are the most commonly used and it might not be clear by the name of the strategy that "api_key" would be a way to implement bearer auth. This new strategy is meant to support non-standard authentication methods that might require the key to be prefixed with custom values or to be placed in a specific location.

Aha! Link: https://ethyca.aha.io/features/OPS-92

### PR DESCRIPTION
**Pull request by adamsachs**
# Purpose

Add an `api_key` `AuthenticationStrategy` for SaaS connectors that allows an API key (i.e. any arbitrary `connector_param`) to be passed in as a request header or query param for all of the requests generated by the given SaaS connector. This is to simplify configuration for APIs that use this type of authentication; users will no longer have to add a `header` or `query_param` configuration for each request in their config.

See #334 for usage details - the feature was implemented close to how was initially described in the issue, and we've since updated the issue description to reflect what's been implemented on this PR

# Changes
- add `api_key` auth strategy and associated tests
- update `shopify` and `datadog` SaaS configs to use the new strategy

# Checklist

# Ticket

Fixes #334

### CODE REVIEW

File: src/fidesops/ops/schemas/saas/strategy_configuration.py [hunk 1]
```diff
@@ -87,6 +91,33 @@ class CursorPaginationConfiguration(StrategyConfiguration):
...
+class ApiKeyAuthenticationParameter(QueryParam):
+    method: Optional[List[HTTPMethod]]
+
+
+class ApiKeyAuthenticationConfiguration(StrategyConfiguration):
+    """
+    API key parameter to be added in as a header or query param
+    """
+
+    header: Optional[List[ApiKeyAuthenticationParameter]]
```
galvana, line 103: What are your thoughts on changing header and query_param to plural? This way it can be consistent with the way headers and query_params are defined in the requests and it better reflects the fact that you can include a list of both.

src/fidesops/ops/schemas/saas/strategy_configuration.py
```diff
@@ -87,6 +91,33 @@ class CursorPaginationConfiguration(StrategyConfiguration):
     field: str
 
 
+class ApiKeyAuthenticationParameter(QueryParam):
+    method: Optional[List[HTTPMethod]]
```
galvana, line 95: This took me a while but I see that we use the method to indicate if the headers/query_params should be added to a request. Did we run into this use case with the connectors already?

adamsachs: we did run into this with [Rollbar](https://github.com/ethyca/fidesops/pull/1227), but per discussion with @galvana , we decided to de-scope this feature from the PR/issue, and allow Rollbar to continue to be configured "manually" via query param. Adding this functionality to the auth strategy seemed beyond the scope of what auth strategies are meant to do.

src/fidesops/ops/schemas/saas/strategy_configuration.py
```diff
@@ -87,6 +87,27 @@ class CursorPaginationConfiguration(StrategyConfiguration):
     field: str
 
 
+class ApiKeyAuthenticationConfiguration(StrategyConfiguration):
+    """
+    API key parameter to be added in as a header or query param
+    """
+
+    headers: Optional[List[QueryParam]]
```
galvana, line 95: Minor, but we have a `Header` type in `saas_config.py`

adamsachs: oh nice. let me go ahead and change that, feels quite a bit cleaner.

galvana (approved):
Thanks for making the requested changes! This looks good to me 👍

adamsachs:
i think we can get this merged because i think the single shopify failure that came up is one that's known to be intermittent and unrelated to these code changes, right @galvana?

PR #1364, opened 2022-09-21
Updating offset pagination strategy to parse numeric strings as integers

galvana:
# Purpose
ConnectionConfig secret values might be stored as strings, but when these values are used in the offset pagination strategy, they need to be converted back to integers. This update casts numeric strings to integers or throws a validation error if the string cannot be cast to an integer.

# Changes
- Added str to int conversion in the offset pagination strategy and added the appropriate tests

# Checklist


adamsachs (approved):
lgtm

eastandwestwind (approved):
Looks good, thanks for the thorough tests as well @galvana !

PR #1376, opened 2022-09-22
Enable retries on saas connectors for failures at the http request level

### LINKED ISSUE #1290: Enable Retries on Http calls for Saas Connectors

Currently our saas connectors fail when hitting a throttle from http servers without any sort of retrying.

Saas connectors use the `requests` library to make http request. This library supports retrying on throttles but we should also consider retries at the Node level.

* It is possible to handle these exceptions at the Node level
* We're also planning on working in a client side rate limiter but these can either work together or be exclusive features

### Additional context

[Example](https://stackoverflow.com/questions/23267409/how-to-implement-retry-mechanism-into-python-requests-library)
```
import logging
import requests

from requests.adapters import HTTPAdapter, Retry

logging.basicConfig(level=logging.DEBUG)

s = requests.Session()
retries = Retry(total=5, backoff_factor=1, status_forcelist=[ 502, 503, 504 ])
s.mount('http://', HTTPAdapter(max_retries=retries))

s.get("http://httpstat.us/503")
```

earmenda:
# Purpose
Http calls within our saas connectors currently don't support retries. Even though retries are supported at the task node level, we don't want to retry all the work which has already been done within the node. 

Rather than relying on `requests` library for retry logic we implemented ourselves so that it is simple to track retries in our upcoming rate limiting work. 

# Changes
- Added a decorator `retry_send` to `send` method within `AuthenticatedClient`
- Decorator sleeps after any failures according to input `backoff_factor` and `retry_count`
- Configured decorator to retry connection exceptions and a list of status codes(429, 502, 503, 504)
- Retry logic looks for `Retry-After` header and sleeps based on that. Logic inspired by https://github.com/urllib3/urllib3/blob/26383f7acfcfd631942bbb579f284c0a6ec84f19/src/urllib3/util/retry.py#L301
- Added a way to test `AuthenticatedClient` by mocking requests session and testing different retry scenarios
- Removed sleep from `test_zendesk_task.py` which was previously needed due to throttles

# Checklist
- [ ] ~Applicable documentation updated (guides, quickstart, postman collections, tutorial, fidesdemo, [database diagram](https://github.com/ethyca/fidesops/blob/main/docs/fidesops/docs/development/update_erd_diagram.md).~
- ~If docs updated (select one):~
  - [ ] ~documentation complete, or draft/outline provided (tag docs-team to complete/review on this branch)~
  - [ ] ~documentation issue created (tag docs-team to complete issue separately)~
- [ ] ~This PR contains a DB migration. If checked, the reviewer should confirm with the author that the [down_revision correctly references the previous migration](https://ethyca.github.io/fidesops/development/contributing_details/#alembic-migrations) before merging~

# Ticket

Fixes #1290

### DIFF
--- a/src/fidesops/ops/service/connectors/saas/authenticated_client.py
+++ b/src/fidesops/ops/service/connectors/saas/authenticated_client.py
@@ -81,8 +81,7 @@ class AuthenticatedClient:
         # otherwise just return the prepared request
         return req
 
-    def retry_send(
-        self,
+    def retry_send(  # type: ignore
         retry_count: int,
         backoff_factor: float,
         retry_status_codes: List[int] = [429, 502, 503, 504],
@@ -117,7 +116,7 @@ class AuthenticatedClient:
 
                         if status_code not in retry_status_codes:
                             break
-                        
+
                         # override sleep time if retry after header is found
                         retry_after_time = get_retry_after(response)
                         sleep_time = (
@@ -141,7 +140,7 @@ class AuthenticatedClient:
 
         return decorator
 
-    @retry_send(backoff_factor=1,retry_count=4)
+    @retry_send(backoff_factor=1, retry_count=4)
     def send(
         self, request_params: SaaSRequestParams, ignore_errors: Optional[bool] = False
     ) -> Response:
--- a/src/fidesops/ops/service/connectors/saas/authenticated_client.py
+++ b/src/fidesops/ops/service/connectors/saas/authenticated_client.py
@@ -91,7 +91,7 @@ class AuthenticatedClient:
 
         Exponential backoff factor uses the following formula:
         backoff_factor * (2 ** (retry_attempt))
-        For an backoff_factor of 1 it will sleep for 2,4,8,16 seconds
+        For an backoff_factor of 1 it will sleep for 2,4,8 seconds
 
         General exceptions are always retried. RequestFailureResponseException exceptions are only retried
         if the status code is in retry_status_codes.
@@ -103,8 +103,8 @@ class AuthenticatedClient:
                 self = args[0]
                 last_exception: Optional[Union[BaseException, Exception]] = None
 
-                for attempt in range(retry_count):
-                    sleep_time = backoff_factor * (2 ** (attempt - 1))
+                for attempt in range(retry_count + 1):
+                    sleep_time = backoff_factor * (2 ** (attempt + 1))
                     try:
                         return func(*args, **kwargs)
                     except RequestFailureResponseException as exc:  # pylint: disable=W0703
@@ -140,7 +140,7 @@ class AuthenticatedClient:
 
         return decorator
 
-    @retry_send(retry_count=4, backoff_factor=1.0)  # pylint: disable=E1124
+    @retry_send(retry_count=3, backoff_factor=1.0)  # pylint: disable=E1124
     def send(
         self, request_params: SaaSRequestParams, ignore_errors: Optional[bool] = False
     ) -> Response:
--- a/src/fidesops/ops/service/connectors/saas/authenticated_client.py
+++ b/src/fidesops/ops/service/connectors/saas/authenticated_client.py
@@ -93,7 +93,7 @@ class AuthenticatedClient:
         backoff_factor * (2 ** (retry_attempt))
         For an backoff_factor of 1 it will sleep for 2,4,8 seconds
 
-        General exceptions are always retried. RequestFailureResponseException exceptions are only retried
+        General exceptions are not retried. RequestFailureResponseException exceptions are only retried
         if the status code is in retry_status_codes.
         """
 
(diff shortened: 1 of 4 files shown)


earmenda:
@galvana this is still a draft but I tested this and this lets us uncomment the sleep from the zendesk tests

In my testing zendesk returns a 429 which includes this header:
```'retry-after': '38'```
Verified we respect this:
```Retrying http request in 38 seconds```

And tests pass!

**Comment by earmenda** [comment]
This should be ready for review

File: tests/ops/service/connectors/saas/test_authenticated_client.py [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pytest.fixture
+def test_authenticated_client(test_connection_config) -> AuthenticatedClient:
+    return AuthenticatedClient("https://test_uri", test_connection_config)
+
+
+@pytest.mark.unit_saas
+@mock.patch.object(Session, "send")
+class TestAuthenticatedClient:
```
adamsachs, line 54: the mocking here is nicely done - these tests are really easy to follow.

src/fidesops/ops/service/connectors/saas/authenticated_client.py
```diff
@@ -75,27 +81,77 @@ def get_authenticated_request(
...
+                        )
+                        sleep(sleep_time)
+
+                raise last_exception  # type: ignore
+
+            return result
+
+        return decorator
+
+    @retry_send(retry_count=3, backoff_factor=1.0)  # pylint: disable=E1124
```
adamsachs, line 143: maybe i'm missing something - but i was expecting these values to be config driven, rather than hardcoded here.

i assume these are just some defaults and we do plan to expose some config to drive the values here, as different endpoints and use cases have different requirements and allowances for retries. in that case, i wonder if it'd be a bit cleaner to put the default arg values in the `retry_send` method definition itself?

src/fidesops/ops/service/connectors/saas/authenticated_client.py
```diff
@@ -75,27 +81,77 @@ def get_authenticated_request(
...
+        retry_status_codes: List[int] = [429, 502, 503, 504],
+    ) -> Callable:
+        """
+        Retry decorator for http requests, backing off exponentially or listening to server retry-after header
+
+        Exponential backoff factor uses the following formula:
+        backoff_factor * (2 ** (retry_attempt))
+        For an backoff_factor of 1 it will sleep for 2,4,8 seconds
+
+        General exceptions are always retried. RequestFailureResponseException exceptions are only retried
```
adamsachs, line 96: this struck me as a bit counter-intuitive. but i suppose we're just sort of "playing it safe" here, and assuming we should retry if we don't know more about the exception? i'd just like to understand the reasoning a bit better on why we always retry on "general" exceptions (which, to me, seem more likely to be caused by bad config or some user error) but are more discerning of the `RequestFailureResponseException`s

adamsachs (commented):
this looks great @earmenda, i think it's a really clean implementation and you've got good test coverage. i like that we're supporting both "client-driven" and "server-driven" (i.e. response-header-based) logic.

i have a few minor comments and questions, but nothing too substantial. my general understanding of this is that it's setting up the infrastructure for the client-driven configuration of the retry logic, but on a quick search, i didn't find the issue for that subsequent work of exposing this functionality in the saas config anywhere -- so i just want to confirm that understanding.

happy to chat about it more offline if that's better.

tests/ops/integration_tests/saas/test_zendesk_task.py
```diff
@@ -202,7 +202,6 @@ async def test_zendesk_erasure_request_task(
     graph = DatasetGraph(merged_graph)
 
     # Since we sometimes get response: b'Number of allowed API requests per minute exceeded' so adding this line to avoid it
```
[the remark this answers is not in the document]
earmenda: thanks!

tests/ops/service/connectors/saas/test_authenticated_client.py
earmenda: ty ty, surprised it ended up working so nicely tbh

src/fidesops/ops/service/connectors/saas/authenticated_client.py
earmenda: We discussed offline. For now we'll go forward with these values to only be configurable within our code. This only controls retries for certain error responses, we might need to expand this in the future so that a user could customize this.

src/fidesops/ops/service/connectors/saas/authenticated_client.py
earmenda: Also discussed this offline. We decided to change this so that general exceptions are not retries. I also added a comment around this.

earmenda:
@adamsachs thanks for the review. i updated the PR with what we discussed

adamsachs (approved):
looks great!

just for transparency, @earmenda and i talked offline and clarified that my previous understanding of the changes were slightly off base. the client-driven rate limiting will be added on top of this code, but we'll keep the retry logic as it is here.

PR #1394, opened 2022-09-27
Backend: Updating or Deleting Access Webhooks [#1388][#1389]

### LINKED ISSUE #1388: Manual Webhooks Re-Queue Requires Input

It is possible that a privacy request is stuck in "requires_input" state because at the time it was submitted, there were manual webhooks configured, but now those webhooks have been deleted or disabled.

1.  When deleting a webhook, check if this affects privacy requests still in a `requires_input` state.  Queue these webhooks for processing automatically.
2.  Set up a recurring job that looks for any privacy requests missed by the above and re-queues them, as a backup.  

### Additional context

### LINKED ISSUE #1389: Backend: Indicate Webhook Fields have Changed

If webhook fields are edited, they may differ from data already saved for a `requires_input` PrivacyRequest and require re-review before the privacy request is queued again for processing.

Currently the backend throws a 429 when retrieving mismatched data, but we need to make this more actionable. 

- When retrieving manual input data for a privacy request (`view_uploaded_manual_webhook_data`), if the fields saved to the privacy request differ from the fields on the webhook (Either a field has been deleted or a field has been removed), we need to surface in the API that this webhook needs to be re-reviewed (checked=False).  Currently we just show checked=False if there's no data in the cache. Perhaps we change the variable name here.
- Noting, don't update the data in the cache (to preserve that this webhook still needs to be re-reviewed, even if the page is refreshed).  This data will need to be overwritten.
- However, in the API return just the data that overlaps between existing data and the current webhook fields.
- Leave other endpoints the same: `resume_privacy_request_from_requires_input` will still 429 if there's a mismatch to protect this from a backend perspective.

### Additional context

pattisdr:
# Purpose

Address some edge cases that surfaced in testing https://github.com/ethyca/fidesops/pull/1377

1. Deleting or disabling a webhook can leave Privacy Requests stuck in a `requires_input` state with no way to move them forward 
2. If data has already been saved for a webhook, but then the webhook definition itself changes, we can run into issues trying to retrieve the data for the webhook because of the mismatch.  The user needs to be alerted that data needs to be resubmitted for the altered webhook.  We currently throw a 429 if the webhook definition had a field removed, but we should also handle if the webhook definition had a field added. Both should require a re-review. 

# Changes
- If a manual webhook is deleted or disabled, check if there are any remaining active manual webhooks configured. If not, meaning we just disabled/deleted the last one, queue any Privacy Requests stuck in "requires_input" for processing.
- In the `view_uploaded_manual_webhook_data` endpoint, load cached webhook data for a privacy request in strict mode. If it fails (no data saved, extra field saved, field missing), return `checked=False`, so the user knows they need to re-upload data for this webhook before it can be submitted.
  - Previously, we would return a 429 if there were extra fields saved for the webhook. Now we return a 200 but surface that the data needs to be re-uploaded.
  - Return the data in non-strict mode, so we just show the overlap between the data saved and the fields defined.
  - Any webhooks that have `checked=True` will prevent the privacy request from being submitted

# Checklist

# Ticket

Fixes #1388
Fixes #1389

### DIFF
--- a/src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
+++ b/src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
@@ -58,6 +58,9 @@ from fidesops.ops.schemas.connection_configuration.connection_secrets import (
 )
 from fidesops.ops.schemas.shared_schemas import FidesOpsKey
 from fidesops.ops.service.connectors import get_connector
+from fidesops.ops.service.privacy_request.request_service import (
+    queue_requires_input_requests,
+)
 from fidesops.ops.util.api_router import APIRouter
 from fidesops.ops.util.logger import Pii
 from fidesops.ops.util.oauth_util import verify_oauth_client
@@ -222,6 +225,9 @@ def patch_connections(
                 )
             )
 
+    # Check if possibly disabling a manual webhook here causes us to need to queue affected privacy requests
+    queue_requires_input_requests(db)
+
     return BulkPutConnectionConfiguration(
         succeeded=created_or_updated,
         failed=failed,
@@ -238,8 +244,14 @@ def delete_connection(
 ) -> None:
     """Removes the connection configuration with matching key."""
     connection_config = get_connection_config_or_error(db, connection_key)
+    connection_type = connection_config.connection_type
     logger.info("Deleting connection config with key '%s'.", connection_key)
     connection_config.delete(db)
+
+    # Access Manual Webhooks are cascade deleted if their ConnectionConfig is deleted,
+    # so we queue any privacy requests that are no longer blocked by webhooks
+    if connection_type == ConnectionType.manual_webhook:
+        queue_requires_input_requests(db)
 
 
 def validate_secrets(
--- a/src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
+++ b/src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
@@ -39,6 +39,8 @@ from fidesops.ops.common_exceptions import (
     ConnectionException,
 )
 from fidesops.ops.models.connectionconfig import ConnectionConfig, ConnectionType
+from fidesops.ops.models.manual_webhook import AccessManualWebhook
+from fidesops.ops.models.privacy_request import PrivacyRequest, PrivacyRequestStatus
 from fidesops.ops.schemas.api import BulkUpdateFailed
 from fidesops.ops.schemas.connection_configuration import (
     connection_secrets_schemas,
@@ -58,8 +60,8 @@ from fidesops.ops.schemas.connection_configuration.connection_secrets import (
 )
 from fidesops.ops.schemas.shared_schemas import FidesOpsKey
 from fidesops.ops.service.connectors import get_connector
-from fidesops.ops.service.privacy_request.request_service import (
-    queue_requires_input_requests,
+from fidesops.ops.service.privacy_request.request_runner_service import (
+    queue_privacy_request,
 )
 from fidesops.ops.util.api_router import APIRouter
 from fidesops.ops.util.logger import Pii
@@ -226,7 +228,7 @@ def patch_connections(
             )
 
     # Check if possibly disabling a manual webhook here causes us to need to queue affected privacy requests
-    queue_requires_input_requests(db)
+    requeue_requires_input_requests(db)
 
     return BulkPutConnectionConfiguration(
         succeeded=created_or_updated,
@@ -251,7 +253,7 @@ def delete_connection(
     # Access Manual Webhooks are cascade deleted if their ConnectionConfig is deleted,
     # so we queue any privacy requests that are no longer blocked by webhooks
     if connection_type == ConnectionType.manual_webhook:
-        queue_requires_input_requests(db)
+        requeue_requires_input_requests(db)
 
 
 def validate_secrets(
@@ -368,3 +370,25 @@ async def test_connection_config_secrets(
     connection_config = get_connection_config_or_error(db, connection_key)
     msg = f"Test completed for ConnectionConfig with key: {connection_key}."
     return connection_status(connection_config, msg, db)
+
+
+def requeue_requires_input_requests(db: Session) -> None:
+    """
+    Queue privacy requests with request status "requires_input" if they are no longer blocked by
+    access manual webhooks.
+    """
+    if not AccessManualWebhook.get_enabled(db):
+        for pr in PrivacyRequest.filter(
+            db=db,
+            conditions=(PrivacyRequest.status == PrivacyRequestStatus.requires_input),
+        ):
+            logger.info(
+                "Queuing privacy request '%s with '%s' status now that manual inputs are no longer required.",
+                pr.id,
+                pr.status.value,
+            )
+            pr.status = PrivacyRequestStatus.in_processing
+            pr.save(db=db)
+            queue_privacy_request(
+                privacy_request_id=pr.id,
+            )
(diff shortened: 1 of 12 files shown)


src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1377,20 +1384,20 @@ def view_uploaded_manual_webhook_data(
...
-        checked = False
-    except PydanticValidationError:
-        raise HTTPException(
-            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=f"Saved fields differ from fields specified on webhook '{access_manual_webhook.connection_config.key}'. "
-            f"Re-upload manual data using '{PRIVACY_REQUEST_ACCESS_MANUAL_WEBHOOK_INPUT.format(connection_key=connection_config.key, privacy_request_id=privacy_request.id)}'.",
+        data = privacy_request.get_manual_webhook_input_non_strict(
+            manual_webhook=access_manual_webhook
         )
+        checked = False
```
pattisdr, line 1400: Previously, "checked=False" was only returned if there was no data in the cache, so Chris's UI would show that that webhook still needed to be reviewed and the submit DSR button would be grayed out.

In addition to this use case, we now show `checked=False` if the webhook definition has changed, which prompts the user to have to re-review. 

The variable name `checked` might need to be updated in the future, but I left it like this for now, because it's the field Chris is using in the FE.

src/fidesops/ops/api/v1/endpoints/privacy_request_endpoints.py
```diff
@@ -1377,20 +1384,20 @@ def view_uploaded_manual_webhook_data(
...
         logger.info(exc)
-        data = access_manual_webhook.empty_fields_dict
-        checked = False
-    except PydanticValidationError:
-        raise HTTPException(
-            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=f"Saved fields differ from fields specified on webhook '{access_manual_webhook.connection_config.key}'. "
-            f"Re-upload manual data using '{PRIVACY_REQUEST_ACCESS_MANUAL_WEBHOOK_INPUT.format(connection_key=connection_config.key, privacy_request_id=privacy_request.id)}'.",
+        data = privacy_request.get_manual_webhook_input_non_strict(
+            manual_webhook=access_manual_webhook
```
pattisdr, line 1398: This is the endpoint that returns data that was already uploaded for a webhook if applicable.

I load the webhook data in non-strict mode here if there's an issue, which really just returns the overlap between the saved fields and the new webhook field definitions.

Note that I don't update the data that is actually cached, I just return this "safe mode version".  This keeps this endpoint idempotent in case it is hit multiple times or by different people.  Regardless, until the webhook fields are updated by a user to match the latest webhook definition, the privacy request itself will not be able to be submitted, because it will also try to access the webhook data in "strict mode".

src/fidesops/ops/models/privacy_request.py
```diff
@@ -505,27 +506,50 @@ def cache_manual_webhook_input(
             parsed_data.dict(),
         )
 
-    def get_manual_webhook_input(
+    def get_manual_webhook_input_strict(
```
pattisdr, line 509: Updated this existing method to "strict mode" which used to just look for extra fields and/or no data cached, but now it also looks for missing fields.

src/fidesops/ops/models/privacy_request.py
```diff
@@ -505,27 +506,50 @@ def cache_manual_webhook_input(
...
+            if set(data.keys()) != set(manual_webhook.fields_schema.__fields__.keys()):
+                raise ManualWebhookFieldsUnset(
+                    f"Fields unset for privacy_request_id '{self.id}' for connection config '{manual_webhook.connection_config.key}'"
+                )
+            return data
         raise NoCachedManualWebhookEntry(
             f"No data cached for privacy_request_id '{self.id}' for connection config '{manual_webhook.connection_config.key}'"
         )
 
+    def get_manual_webhook_input_non_strict(
```
pattisdr, line 536: Added a "non strict" mode to load webhook data where we really just preserve the overlap.

src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -238,9 +246,15 @@ def delete_connection(
...
     """Removes the connection configuration with matching key."""
     connection_config = get_connection_config_or_error(db, connection_key)
+    connection_type = connection_config.connection_type
     logger.info("Deleting connection config with key '%s'.", connection_key)
     connection_config.delete(db)
 
+    # Access Manual Webhooks are cascade deleted if their ConnectionConfig is deleted,
+    # so we queue any privacy requests that are no longer blocked by webhooks
+    if connection_type == ConnectionType.manual_webhook:
+        requeue_requires_input_requests(db)
```
pattisdr, line 256: As a first step, I am just re-queuing applicable privacy requests if a connection is disabled or deleted. I omitted adding a recurring task to also check if any of these fell through the cracks in the interest of time. I think this kind of thing could be added as a piece of a broader work where we look for privacy requests of multiple statuses that are stuck.

tests/ops/api/v1/endpoints/test_connection_config_endpoints.py
```diff
@@ -859,6 +913,17 @@ def test_delete_manual_webhook_connection_config(
             .first()
             is None
         )
+        assert (
+            mock_queue.called
+        ), "Deleting this last webhook caused 'requires_input' privacy requests to be queued"
```
seanpreston, line 918: Just to confirm this is deleting the last webhook, so Fidesops is sure there's no more input required from the user before dispatch?

tests/ops/api/v1/endpoints/test_privacy_request_endpoints.py
```diff
@@ -3295,11 +3295,40 @@ def test_cached_data_differs_from_webhook_fields(
...
+            {"pii_field": "id_no", "dsr_package_label": "id_number"}
         )
+        access_manual_webhook.save(db)
+        response = api_client.get(url, headers=auth_header)
+
+        assert response.status_code == 200
+        assert response.json() == {
+            "checked": False,
+            "fields": {
+                "id_number": None,
```
seanpreston, line 3327: Using the user preferred naming for this field in the response is a nice touch

src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
```diff
@@ -356,3 +370,25 @@ async def test_connection_config_secrets(
     connection_config = get_connection_config_or_error(db, connection_key)
     msg = f"Test completed for ConnectionConfig with key: {connection_key}."
     return connection_status(connection_config, msg, db)
+
+
+def requeue_requires_input_requests(db: Session) -> None:
+    """
+    Queue privacy requests with request status "requires_input" if they are no longer blocked by
+    access manual webhooks.
```
seanpreston, line 378: It'd be better to say explicitly this is if all access manual webhooks are disabled or deleted.

seanpreston (approved):

tests/ops/api/v1/endpoints/test_connection_config_endpoints.py
pattisdr: Yes, the last webhook is deleted, so the privacy request just needs to be re-queued. Data is no longer required for the webhook. If any data was previously uploaded, it would be ignored.

src/fidesops/ops/api/v1/endpoints/connection_endpoints.py
pattisdr: Thanks Sean, I've tried to improve this docstring

PR #1395, opened 2022-09-27
[#1393] Update Fidesops config with sane defaults where necessary

### LINKED ISSUE #1393: Fidesops requires a minimal config TOML file to start and several config variables are missing safe defaults
### Bug Description

In our 1.8.0 release, we removed the default config TOML from our published Docker image. In doing so, we fixed one issue where there were unexpected defaults being set for a variety of config variables, but we also broke any deployments that relied on configuring the service using ENV variables exclusively.

#### Steps to Reproduce

1. Run `docker run ethyca/fidesops:1.8.0` (without any env variables, any volume mounts, etc.)
1. Observe the following error:
```
~/git/fidesops% docker run ethyca/fidesops:1.8.0
Startup configuration: reloading = False, dev_mode = False
Startup configuration: pii logging = False
fidesops.toml or fidesctl.toml or fides.toml could not be loaded:
ValidationError: 8 validation errors for FidesopsConfig
database
  field required (type=value_error.missing)
redis
  field required (type=value_error.missing)
security
  field required (type=value_error.missing)
execution
  field required (type=value_error.missing)
root_user
  field required (type=value_error.missing)
admin_ui
  field required (type=value_error.missing)
notifications
  field required (type=value_error.missing)
port
  field required (type=value_error.missing)
```

### Expected behavior

If you read the error carefully above, it's actually attempting to load a `fidesops.toml` and failing, so it becomes unable to load any of the sub-configurations for `database`, `redis`, `security`, etc. Because of this, it never gets to the stage where it can attempt to read ENV variables to set those values.

In other words: it should be possible to simply `docker run ethyca/fidesops` and for the resulting error should be that a small handful of required config variables need to be set, including:
* `FIDESOPS__SECURITY__APP_ENCRYPTION_KEY`
* `FIDESOPS__SECURITY__OAUTH_ROOT_CLIENT_ID`
* `FIDESOPS__SECURITY__OAUTH_ROOT_CLIENT_SECRET`
* `FIDESOPS__DATABASE__SERVER`
* `FIDESOPS__DATABASE__USER`
* `FIDESOPS__DATABASE__PASSWORD`
* `FIDESOPS__DATABASE__DB`
* `FIDESOPS__DATABASE__PORT`
* `FIDESOPS__REDIS__HOST`
* `FIDESOPS__REDIS__PORT`
* `FIDESOPS__REDIS__PASSWORD`

These are the minimum ENV variables documented in the deployment guide here: https://ethyca.github.io/fidesops/deployment/#install-fidesops-via-docker

I believe all other config variables should be either optional (i.e. if not set, the features that rely on them can't be used at runtime) or have safe defaults. There may be a few exceptions which we can discuss and document, but ultimately the goal here should be to make it possible to run this again via docker.

### Workaround
Today, I found that the safest workaround is two parts.

First, to workaround the required TOML loading, provide this minimal config TOML to the container (via a mount, e.g. the `--mount` or `-v` option to `docker run`):
```
port = 8080

[database]
enabled = true

[redis]
db_index = 0

[security]

[execution]
task_retry_count = 0
task_retry_delay = 1
task_retry_backoff = 1

[root_user]

[admin_ui]

[notifications]
```

Second, to workaround a few other config variables that don't have safe defaults, provide some ENV vars:
* `FIDESOPS__SECURITY__DRP_JWT_SECRET=<some secret>` (this should be optional)

Lastly, to workaround the requirement that a TOML file exists and is writeable for analytics purposes, provide a few more ENV vars:
* `FIDESOPS__ROOT_USER__ANALYTICS_OPT_OUT=true`
* `FIDESOPS__ROOT_USER__ANALYTICS_ID=<some guid>`

I don't like this third workaround, but note that it's related to a different issue we are tracking here: https://github.com/ethyca/fidesops/issues/1369. I only document it here for temporary purposes, but solving that "read-only" TOML problem should be addressed in https://github.com/ethyca/fidesops/issues/1369, not here.

### Environment

* Version: 1.8.0
* OS: Mac OSX
* Python Version: 3.9
* Docker Version: 20.10.17

seanpreston:
# Purpose

This PR addresses some shortcomings in the Fidesops config highlighted by the recent removal of the `fidesops.toml` from the Docker build. I've not tried to set defaults for any config vars within the `[database]`, `[redis]`, or `[security]` subsections, since they're either handled separately, or setting a default is an anti-pattern.

This [Fideslib PR](https://github.com/ethyca/fideslib/pull/73/) also aims to clarify which part of config. loading any error has occurred in.

# Changes
- Only populate `analytics_id` if `analytics_opt_out` is either `None` or `False` — this still preserves the behaviour of an explicit opt-out.
- Sets `port` to `8080`
- Sets `task_retry_` vars to `0` — this means by default nodes in the Fidesops identity graph traversal will not attempt to retry if they fail.

# Checklist

# Ticket
Fixes #1393

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ celery[pytest]==5.2.7
 click==8.1.3
 cryptography~=3.4.8
 dask==2022.8.0
-emails
 fastapi-caching[redis]
 fastapi-pagination[sqlalchemy]~= 0.10.0
 fastapi[all]==0.82.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ fastapi-caching[redis]
 fastapi-pagination[sqlalchemy]~= 0.10.0
 fastapi[all]==0.82.0
 fideslang==1.2.0
-fideslib==3.1.3
+fideslib==3.1.4
 fideslog==1.2.3
 hvac==0.11.2
 Jinja2==3.1.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ fastapi-caching[redis]
 fastapi-pagination[sqlalchemy]~= 0.10.0
 fastapi[all]==0.82.0
 fideslang==1.2.0
-fideslib==3.1.2
+fideslib==3.1.3
 fideslog==1.2.3
 hvac==0.11.2
 Jinja2==3.1.2
--- a/src/fidesops/ops/core/config.py
+++ b/src/fidesops/ops/core/config.py
@@ -174,10 +174,10 @@ class FidesopsConfig(FidesSettings):
     database: FidesopsDatabaseSettings
     redis: RedisSettings
     security: FidesopsSecuritySettings
-    execution: ExecutionSettings
-    root_user: RootUserSettings
-    admin_ui: AdminUiSettings
-    notifications: FidesopsNotificationSettings
+    execution: Optional[ExecutionSettings]
+    root_user: Optional[RootUserSettings]
+    admin_ui: Optional[AdminUiSettings]
+    notifications: Optional[FidesopsNotificationSettings]
 
     port: int = 8080  # Run the webserver on port 8080 by default
     is_test_mode: bool = os.getenv("TESTING", "").lower() == "true"
--- a/src/fidesops/ops/core/config.py
+++ b/src/fidesops/ops/core/config.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 class FidesopsDatabaseSettings(DatabaseSettings):
     """Configuration settings for Postgres."""
 
-    ENABLED: bool = True
+    enabled: bool = True
 
     class Config:
         env_prefix = "FIDESOPS__DATABASE__"
@@ -174,10 +174,12 @@ class FidesopsConfig(FidesSettings):
     database: FidesopsDatabaseSettings
     redis: RedisSettings
     security: FidesopsSecuritySettings
-    execution: Optional[ExecutionSettings]
-    root_user: Optional[RootUserSettings]
-    admin_ui: Optional[AdminUiSettings]
-    notifications: Optional[FidesopsNotificationSettings]
+    execution: Optional[ExecutionSettings] = ExecutionSettings()
+    root_user: Optional[RootUserSettings] = RootUserSettings()
+    admin_ui: Optional[AdminUiSettings] = AdminUiSettings()
+    notifications: Optional[
+        FidesopsNotificationSettings
+    ] = FidesopsNotificationSettings()
 
     port: int = 8080  # Run the webserver on port 8080 by default
     is_test_mode: bool = os.getenv("TESTING", "").lower() == "true"
--- a/src/fidesops/ops/core/config.py
+++ b/src/fidesops/ops/core/config.py
@@ -35,9 +35,10 @@ class ExecutionSettings(FidesSettings):
     """Configuration settings for execution."""
 
     privacy_request_delay_timeout: int = 3600
-    task_retry_count: int
-    task_retry_delay: int  # In seconds
-    task_retry_backoff: int
+    # By default Fidesops will not retry graph nodes
+    task_retry_count: int = 0
+    task_retry_delay: int = 0  # In seconds
+    task_retry_backoff: int = 0
     subject_identity_verification_required: bool = False
     require_manual_request_approval: bool = False
     masking_strict: bool = True
@@ -118,15 +119,21 @@ class FidesopsSecuritySettings(SecuritySettings):
 class RootUserSettings(FidesSettings):
     """Configuration settings for Analytics variables."""
 
-    analytics_opt_out: Optional[bool]
-    analytics_id: Optional[str]
+    analytics_opt_out: Optional[bool] = True
+    analytics_id: Optional[str] = None
 
     @validator("analytics_id", pre=True)
-    def populate_analytics_id(cls, v: Optional[str]) -> str:
+    def populate_analytics_id(
+        cls,
+        v: Optional[str],
+        values: Dict[str, str],
+    ) -> str:
         """
         Populates the appropriate value for analytics id based on config
         """
-        return v or cls.generate_and_store_client_id()
+        if not v and not values.get("analytics_opt_out"):
+            v = cls.generate_and_store_client_id()
+        return v
 
     @staticmethod
     def generate_and_store_client_id() -> str:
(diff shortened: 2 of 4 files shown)


src/fidesops/ops/core/config.py
```diff
@@ -35,9 +35,10 @@ class ExecutionSettings(FidesSettings):
     """Configuration settings for execution."""
 
     privacy_request_delay_timeout: int = 3600
-    task_retry_count: int
-    task_retry_delay: int  # In seconds
-    task_retry_backoff: int
+    # By default Fidesops will not retry graph nodes
```
NevilleS, line 38: Why not? Our previous defaults for this (set in the TOML) did have nonzero retries, and I think that's helpful behaviour!

src/fidesops/ops/core/config.py
```diff
@@ -118,15 +119,21 @@ class Config:
 class RootUserSettings(FidesSettings):
     """Configuration settings for Analytics variables."""
 
-    analytics_opt_out: Optional[bool]
-    analytics_id: Optional[str]
+    analytics_opt_out: Optional[bool] = True
+    analytics_id: Optional[str] = None
```
NevilleS, line 123: Thanks - this helps for sure

src/fidesops/ops/core/config.py
```diff
@@ -167,12 +174,14 @@ class FidesopsConfig(FidesSettings):
...
-
-    port: int
+    execution: Optional[ExecutionSettings] = ExecutionSettings()
+    root_user: Optional[RootUserSettings] = RootUserSettings()
+    admin_ui: Optional[AdminUiSettings] = AdminUiSettings()
+    notifications: Optional[
+        FidesopsNotificationSettings
+    ] = FidesopsNotificationSettings()
+
+    port: int = 8080  # Run the webserver on port 8080 by default
```
NevilleS, line 184: This variable can't be set via an ENV option right now, since there isn't a env var prefix set for the top-level config object

src/fidesops/ops/core/config.py
```diff
@@ -167,12 +174,14 @@ class FidesopsConfig(FidesSettings):
     database: FidesopsDatabaseSettings
     redis: RedisSettings
     security: FidesopsSecuritySettings
-    execution: ExecutionSettings
-    root_user: RootUserSettings
-    admin_ui: AdminUiSettings
-    notifications: FidesopsNotificationSettings
-
-    port: int
+    execution: Optional[ExecutionSettings] = ExecutionSettings()
```
NevilleS, line 177: There's a (required!) config variable missing from this schema: fidesops.security.drp_jwt_secret. Can you add it to the schema and ensure it's optional?

NevilleS (changes requested):
Thanks - few notes from my end

src/fidesops/ops/core/config.py
seanpreston: It's in the [Fideslib config](https://github.com/ethyca/fideslib/blob/main/fideslib/core/config.py#L97) already, so we shouldn't specify it twice

src/fidesops/ops/core/config.py
seanpreston: Good spot!

src/fidesops/ops/core/config.py
NevilleS: OK, I'll trust you to roll this into future improvements so that our default retry logic is reasonable. I'd argue that we should disable it in CI if we want (we can override this ourselves!) but understand if the retries are surprising right now. We can improve that in our error reporting work

src/fidesops/ops/core/config.py
NevilleS: Ah, I see! I'd like it to have a default though, can't we make that optional?

NevilleS (dismissed):
One more suggestion: can you bump the fideslib requirement in this PR and ensure the related PR lands? So that this all comes together into the ops codebase and closing the related issue.

requirements.txt
```diff
@@ -5,7 +5,6 @@ celery[pytest]==5.2.7
 click==8.1.3
 cryptography~=3.4.8
 dask==2022.8.0
-emails
```
seanpreston, line 8: This is unused as far as I can tell so I'm removing it here as part of this housekeeping.

src/fidesops/ops/core/config.py
```diff
@@ -167,12 +174,19 @@ class FidesopsConfig(FidesSettings):
...
+    root_user: Optional[RootUserSettings] = RootUserSettings()
+    admin_ui: Optional[AdminUiSettings] = AdminUiSettings()
+    notifications: Optional[
+        FidesopsNotificationSettings
+    ] = FidesopsNotificationSettings()
+
+    port: int = int(
```
seanpreston, line 184: I'd be open to other ways of doing this than casting to `int(...)` though it seemed practical for now.

My main gripe is that this could be confusing for users as the env vars are expected to be strings, where port must then be an integer. It's technically possible to run Fidesops with `FIDESOPS__PORT="iamastring"` and the error message would be along the lines of `ValueError: invalid literal for int() with base 10: 'iamastring'`.

src/fidesops/ops/core/config.py
seanpreston: Good spot, have made [this update to Fideslib](https://github.com/ethyca/fideslib/pull/75)

src/fidesops/ops/core/config.py
```diff
@@ -118,15 +119,21 @@ class Config:
...
-    analytics_id: Optional[str]
+    analytics_opt_out: Optional[bool] = True
+    analytics_id: Optional[str] = None
 
     @validator("analytics_id", pre=True)
-    def populate_analytics_id(cls, v: Optional[str]) -> str:
+    def populate_analytics_id(
+        cls,
+        v: Optional[str],
+        values: Dict[str, str],
```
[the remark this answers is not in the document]
seanpreston: Yep! It's the incoming type

seanpreston:
> After these updates jwt_key: str = config.security.drp_jwt_secret in src/fidesops/ops/api/v1/endpoints/drp_endpoints.py has the wrong type now

I updated that already, you might just need to refresh the diff. Will merge away.

sanders41 (approved):

src/fidesops/ops/core/config.py
```diff
@@ -35,9 +35,10 @@ class ExecutionSettings(FidesSettings):
     """Configuration settings for execution."""
 
     privacy_request_delay_timeout: int = 3600
-    task_retry_count: int
-    task_retry_delay: int  # In seconds
-    task_retry_backoff: int
+    # By default Fidesops will not retry graph nodes
+    task_retry_count: int = 0
+    task_retry_delay: int = 0  # In seconds
```
ThomasLaPiana, line 40: a nit here, but `# In seconds` doesn't help end-users. If we want to clarify it we should update the variable name to contain the unit, i.e. `task_retry_delay_seconds`

sanders41: You don't think all users are going to read the source code first 😄 .

Making this change would be a breaking change, `task_retry_delay` was in previous releases it just gets a default value now. So the question now is it worth it to make a breaking change or could we add something to the docs and that be enough?

PR #1399, opened 2022-09-28
firebase auth integration

### LINKED ISSUE #1332: Firebase Authentication

adamsachs:
# Purpose
Firebase Auth (https://firebase.google.com/products/auth) connector - access and update. Leverages the [Firebase Admin SDK](https://firebase.google.com/docs/auth/admin/manage-users) for integration.

# Changes
- Firebase connector config, dataset, test artifacts, etc. -- basic SaaS connector artifacts
- Request overrides which define the integration functionality using the Firebase Admin SDK
- Introduces the Firebase Admin SDK as a project dependency

# Checklist

# Ticket

Fixes #1332


data/saas/saas_connector_registry.toml
```diff
@@ -16,6 +16,12 @@ dataset = "data/saas/dataset/datadog_dataset.yml"
 icon = "data/saas/icon/default.svg"
 human_readable = "Datadog"
 
+[firebase_auth]
+config = "data/saas/config/firebase_auth_config.yml"
+dataset = "data/saas/dataset/firebase_auth_dataset.yml"
+icon = "data/saas/icon/default.svg"
```
adamsachs, line 22: not sure if we want to put an image in here or just use the default...

data/saas/config/firebase_auth_config.yml
```diff
@@ -0,0 +1,54 @@
...
+        query_param:
+        - name: key
+          value: <api_key>
+
+  test_request:
+    method: POST
+    path: /v1/accounts:createAuthUri
+
+
+  endpoints:
```
adamsachs, line 42: decided to configure `read` and `update` endpoints for now; we could easily implement `delete` (and i have in fact implemented the override function), but chose to keep it unconfigured (i.e. disabled) for now to avoid any confusion. open to discussing and/or adjusting.

requirements.txt
```diff
@@ -12,6 +12,7 @@ fastapi[all]==0.82.0
 fideslang==1.2.0
 fideslib==3.1.2
 fideslog==1.2.3
+firebase-admin==5.3.0
```
adamsachs, line 15: not sure if anything else is needed to bring in this dependency, but things seem to work with just this update locally

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+            disabled=disabled,
+            email_verified=email_verified,
+            app=app,
+        )
+        rows_updated += 1
+    return rows_updated
+
+
+@register("firebase_auth_user_delete", [SaaSRequestType.DELETE])
+def firebase_auth_user_delete(
```
adamsachs, line 101: as noted above, this isn't actually being leveraged currently. but i figured i'd keep the function implementation here just for reference, in case we do want to go down this route at some point. happy to include this in some other manner, but didn't want this work to get lost.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+        row = {"email": user.email, "uid": user.uid}
+        if user.display_name is not None:
+            row["display_name"] = user.display_name
+        if user.phone_number is not None:
+            row["phone_number"] = user.phone_number
+        if user.disabled is not None:
+            row["disabled"] = user.disabled
+        if user.email_verified is not None:
+            row["email_verified"] = user.email_verified
+        if user.provider_data:
```
adamsachs, line 48: i was having a bit of trouble testing this functionality because i don't think we're able to programmatically create provider data on a given user, so we can't seed the test data for this. i'll look to see if i can manually verify the functionality moving forward.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+    for row_param_values in param_values_per_row:
+        email = row_param_values.get("email")
+        user: UserRecord = auth.get_user_by_email(email, app=app)
+        masked_fields = row_param_values["masked_object_fields"]
+        display_name = masked_fields.get("display_name", user.display_name)
+        phone_number = masked_fields.get("phone_number", user.phone_number)
+        disabled = masked_fields.get("disabled", user.disabled)
+        email_verified = masked_fields.get("email_verified", user.email_verified)
+
+        auth.update_user(
```
adamsachs, line 87: `provider_data` cannot be updated programmatically via the API, so we don't try to update it here.

adamsachs:
this may need a few more iterations, but i wanted to get an initial cut in front of some of your eyes. @seanpreston i put you on this just to make sure everything looks OK with bringing in the new `firebase_admin` dependency. 

@galvana / @earmenda  i've added test credentials to 1pw, could you help me add them to vault so that we can get the integration tests running on github?

adamsachs:
the firebase test task is passing now, i'm getting an error with sendgrid that is almost certainly not related to these changes

adamsachs:
now getting an error with braze in the saas tests, which i also doubt is related to the PR

data/saas/config/firebase_auth_config.yml
earmenda: commented on this below

earmenda (commented):
changes look good! just a few comments

data/saas/saas_connector_registry.toml
earmenda: don't think this is important atm

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: yup, that's how i felt about it too. maybe it's worth me throwing in a comment here just to clarify it moving forward.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: interesting, thanks for digging that up! i'm with you on keeping the tests self-contained (and trying to run a CLI command within a pytest harness doesn't sound very appealing), but evaluating the provider data functionality may end up being important...so i'm kinda split here. i'll try to do some more digging on my end, thanks for the pointer.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: that's not a bad point. perhaps it may depend on the use case - maybe we need some input from our first use case, and generally we should look to enable both pieces of functionality (as i've done by defining `update` and `delete` override functions)

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+            pds = []
+            for pd in user.provider_data:
+                pds.append(
+                    {
+                        "display_name": pd.display_name,
+                        "provider_id": pd.provider_id,
+                        "email": pd.email,
+                    }
+                )
+            row["provider_data"] = pds
```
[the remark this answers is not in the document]
adamsachs: good catch! the problems that come with not having test coverage here...🤦

data/saas/dataset/firebase_auth_dataset.yml
```diff
@@ -0,0 +1,49 @@
+dataset:
```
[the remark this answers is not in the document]
adamsachs: yeah, i'm with you on the `photo_url` - funny enough, i'd thought of including that in the dataset but then couldn't come up with a suitable fideslang data category to assign it, so then i just excluded it. but that's not a very good reason, so i think i'll include it again. any thoughts on what would be a good data category in the default taxonomy to assign it? (no worries if you don't have something off the top of your head)

the custom claims piece looks interesting - thanks for digging into that - but i'm just not sure how we'd handle that generically, since it's effectively schema-less. i _could_ see it containing relevant PII in certain use cases, so i don't feel great about just ignoring it

if we did decide to support this, what i'm thinking is we could allow users to configure their dataset according to what their `custom_claims` metadata looks like. then i could probably have something generic in the function to dynamically support whatever fields the dataset defines in the sub-object.

that's not a trivial amount of work, though, and we don't have a great paradigm in place for allowing users to overwrite the default datasets we're providing them, which would be necessary for this to work. so i'm a bit unsure here.

data/saas/config/firebase_auth_config.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: client_x509_cert_url
+
+
+  client_config:
+    protocol: https
+    host: <domain>
+    authentication:
+      strategy: api_key
+      configuration:
+        query_param:
```
galvana, line 33: Should this be `query_params`?

data/saas/dataset/firebase_auth_dataset.yml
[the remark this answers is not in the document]
galvana: I usually go with a more generic data category if a specific one isn't available, in this case I would lean towards just annotating `photo_url` as `user` data.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+def initialize_firebase(secrets: Dict[str, Any]) -> App:
+    """
+    Initializes a Firebase "app" instance based on the given
+    secrets `dict`. The value of the `project_id` key of the
+    secrets dict is used for the `name` of the Firebase app.
+    """
+    creds = get_firebase_credentials(secrets)
+    project_id = creds.project_id
+    try:
+        app = firebase_admin.get_app(project_id)
```
galvana, line 131: I remember you saying that we would get an error if we tried to initialize an "app" if one already existed. Is that why we're doing the try block? Not sure if I understand the issue but would Python's `global` keyword help here? We use it in `cache.py` to always return a single instance of the Redis cache.

data/saas/config/firebase_auth_config.yml
adamsachs: good catch! i didn't mean to commit this, it was from an earlier iteration.

you know this actually isn't used at all - none of the `client_config` is. i don't think i can remove that wholesale, but i can remove the `authentication` since that's optional. i'll put in dummy values for the `protocol` and `host` to make it clearer that it's not being used.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: actually, this is sort of the opposite of what i'd described. after mentioning the piece about initializing an app if it already exists, i found this method in the SDK that allows us to effectively retrieve an app by name/id, if it already exists; if it doesn't exist, it throws an error. and so we catch that here and initialize if it doesn't exist yet.

the `global` keyword may be a bit trickier here, because we don't want to assume there's only a single firebase `app` per fidesops runtime - that would prevent a single fidesops instance from integrating with multiple firebase projects, which is a limitation we should avoid if possible imo. perhaps we could maintain a global map of firebase `app`s tied to particular saas config instances, or something of the sort -- but that's effectively what we're getting for free here with the admin SDK!

i don't love that we need to try/catch, but other than that, i actually find this a pretty clean implementation. let me know what you think.

data/saas/dataset/firebase_auth_dataset.yml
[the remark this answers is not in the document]
adamsachs: @earmenda and i discussed offline that we won't support `custom_claims` in our first iteration, but if it turns out to be an important client need, we can look to implement it.

we'd probably look to support it by either writing some code into our function to dynamically traverse/enumerate the subfields within the `custom_claims` field as declared in the dataset. we could also likely just serialize and dump out the `custom_claims` subobject as a JSON blob in an access request. we can look to explore those if/when the need arises.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: noting that i found an SDK [function](https://firebase.google.com/docs/reference/admin/python/firebase_admin.auth?authuser=2#import_users)/[class](https://firebase.google.com/docs/reference/admin/python/firebase_admin.auth?authuser=2#importuserrecord) that allows us to import users with provider data defined. i'm updating test fixtures to create our sample users with provider data and include those pieces of data in the tests.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
adamsachs: we're going to try to include some test coverage for both update and delete functionality, and keep both override functions in the codebase. that way it will essentially be a config switch (i.e. updating the saas config) whether to use update or delete in any given use case. we can adjust the default if needed, once we get some feedback from the real-world use cases.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
```diff
@@ -0,0 +1,166 @@
...
+            "email",
+            "uid",
+            "display_name",
+            "phone_number",
+            "photo_url",
+            "disabled",
+            "email_verified",
+        ]
+        for key in keys:
+            row[key] = getattr(user, key, None)
```
adamsachs, line 50: this is going to add `None` values in for fields that are not populated in the response record dictionary/object. is that OK? i _think_ the behavior in "normal" saas requests is just to not include the field in the response record, in which case i think i'll need to handle this slightly differently...

@galvana @earmenda

adamsachs:
@earmenda i think i've addressed all that we talked about. still outstanding questions on whether we'll need to handle the `client_claims` data and whether we want to default to `update` or `delete`, but per our discussion, we're not going to hold up merging on those and we can come back to them if/when needed.

i do still have one minor question (https://github.com/ethyca/fidesops/pull/1399/files#r985340851) that i'd like to clear up before merging. @galvana maybe you can confirm the general behavior there? or @earmenda if you know off the top of your head.

src/fidesops/ops/service/saas_request/override_implementations/firebase_auth_request_overrides.py
earmenda: i'd say use `hasattr` to avoid this

earmenda (approved):
lgtm!

galvana (approved):
A lot of interesting use cases with this connector and an overall good example of request overrides, approved!

PR #1421, opened 2022-10-03
DX Improvements from `v1.8.1` release

seanpreston:
# Purpose

This PR improves the DX in the following ways:
- better reporting of errors when `DatasetConfig` `get_or_create_dataset` is attempted, to log the `IntegrityError` that's thrown by SQLAlchemy when a `DatasetConfig` already exists, and it belongs to a different `ConnectionConfig` to the one specified in the data.
- don't attach the analytics middleware if the user has opted out via the config — this saves a lot of unnecessary logging output during development

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,7 @@ src_paths = ["src", "tests"]
 disable=[
     "bad-option-value",
     "broad-except",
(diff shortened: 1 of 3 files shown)


NevilleS (approved):
LGTM! Merge away

PR #1428, opened 2022-10-14
Specify Multiple Masking Strategies

pattisdr:
# Purpose

Allow multiple masking strategies to be specified when using fides as a masking engine.  

# Changes
- We could already specify a single masking strategy.  Add in a backwards compatible way so you can specify a single strategy or a list of strategies to the same field.  If multiple strategies are supplied, they are executed in order.
- Allow specifying the config for the masking strategy to be much more flexible to accommodate the needs of new masking strategies

# Checklist

# Ticket

Fixes https://github.com/ethyca/fidesops-plus/issues/74

### DIFF
--- a/src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
+++ b/src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
@@ -31,9 +31,9 @@ def mask_value(request: MaskingAPIRequest) -> MaskingAPIResponse:
 
     """
     try:
-        values = request.values or []
-        masked_values: Optional[List[Any]] = request.values.copy()
-        masking_strategies: List[PolicyMaskingSpec] = request.masking_strategies or []
+        values: List[Any] = request.values
+        masked_values: List[Any] = request.values.copy()
+        masking_strategies: List[PolicyMaskingSpec] = request.masking_strategies
 
         num_strat: int = len(masking_strategies)
 
@@ -52,7 +52,7 @@ def mask_value(request: MaskingAPIRequest) -> MaskingAPIResponse:
                 len(values),
                 strategy.strategy,
             )
-            masked_values = masking_strategy.mask(
+            masked_values = masking_strategy.mask(  # type: ignore
                 masked_values, None
             )  # passing in masked values from previous strategy
 
--- a/src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
+++ b/src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import Any, List, Optional
 
 from fastapi import HTTPException
 from starlette.status import HTTP_400_BAD_REQUEST, HTTP_404_NOT_FOUND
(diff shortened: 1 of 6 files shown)


pattisdr:
@ethyca/docs-authors Adding this PR in fidesops instead of fides after talking to Sean - sounds like there may be another release -

Small docs changes around using fides as a masking engine

src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
```diff
@@ -23,19 +24,38 @@
...
-            masking_strategy.strategy, masking_strategy.configuration
-        )
-        logger.info(
-            "Starting masking of %s value(s) with strategy %s",
-            len(values),
-            masking_strategy.strategy,
-        )
-        masked_values = strategy.mask(values, None)
+        values = request.values or []
+        masked_values: Optional[List[Any]] = request.values.copy()
```
seanpreston, line 35: Should this be optional? It looks like it might always be an empty list if nothing else?

src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
```diff
@@ -23,19 +24,38 @@
...
+
+        num_strat: int = len(masking_strategies)
+
+        if num_strat > 1:
+            logger.info(
+                "%s masking strategies requested; running in order.",
+                num_strat,
+            )
+
+        for strategy in masking_strategies:
```
seanpreston, line 46: Should we also handle the case where `len(masking_strategies) == 0`? Right now it looks like it'll return the same empty list, should we be erroring instead? It's effectively a no-op.

conceptualshark:
👍 Looks good on the docs front, thank you!

src/fidesops/ops/api/v1/endpoints/masking_endpoints.py
pattisdr: yes I agree with you, I was getting  "Incompatible types in assignment" mypy errors because `masking.mask` is allowed to take in None.  I think we should instead ignore the type hint error then when we call `masking.mask`

pattisdr:
thanks for reviewing @conceptualshark 😄

pattisdr:
Made some type adjustments @seanpreston back to you!

src/fidesops/ops/schemas/masking/masking_api.py
```diff
@@ -9,7 +9,28 @@ class MaskingAPIRequest(BaseModel):
     """The API Request for masking operations"""
 
     values: List[str]
-    masking_strategy: PolicyMaskingSpec
+    masking_strategy: Union[PolicyMaskingSpec, List[PolicyMaskingSpec]]
+    masking_strategies: List[PolicyMaskingSpec] = []
+
+    @root_validator(pre=False)
+    def build_masking_strategies(cls, values: Dict) -> Dict:
```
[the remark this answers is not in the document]
seanpreston: > This change is to allow multiple -plus strategies to be applied

Thanks for clarifying, this makes more sense.

seanpreston (approved):
Thanks @pattisdr

